PR #60, opened 2019-05-13
Added CheckMarkerInstTest

cormensratio:


vulder (approved):
lgtm
this currently breaks the tests until we merge the InstVerifier branch into vara-dev, therefore, merge needs to be delayed.

PR #83, opened 2019-06-18
Fix case-study creation

julianbreiteneicher:


varats/driver.py
```diff
@@ -480,10 +480,14 @@ def add_common_args(sub_parser):
...
-        cmap = generate_commit_map(git_path, args['end'],
-                                   args['start'] if 'start' in args else None)
-
         args['project_name'] = git_path.stem.replace("-HEAD", "")
+
+        args['get_cmap'] = create_lazy_commit_map_loader(args['project_name'],
+                                                         args.get('cmap', None),
+                                                         args['end'],
+                                                         args['start'] if 'start' in args else None)
+        cmap = args['get_cmap']()
```
vulder, line 489: if it is not optional to create a cmap we can just directly create one without calling the lazy generator, especially when we pass the create cmap directly

vulder (changes requested):

varats/driver.py
julianbreiteneicher: My thinking was that when I need to create the cmap anyways, I might as well just do it like this so that the cmap is cached afterwards

vulder: hmm, ok I don't think we need it here but it also does not hurt. So fine with me.

vulder (approved):

PR #87, opened 2019-06-19
Plot gradient based extender only adds new revisions.

vulder:
The plot gradient based sampler now only adds revisions that are not
present in any stage of the case-study.
Furthermore, this fixes some small typing errors.


julianbreiteneicher (approved):
looks good!

PR #88, opened 2019-06-19
 Fix typing problems when running mypy

vulder:


julianbreiteneicher (approved):
Changes look good to me!

PR #92, opened 2019-06-20
Implement PaperConfigOverviewPlot

julianbreiteneicher:
resolves se-passau/VaRA#398


varats/driver.py
```diff
@@ -250,14 +250,20 @@ def main_gen_graph():
...
 
-    args['get_cmap'] = create_lazy_commit_map_loader(args['project'],
-                                                     args.get('cmap', None))
-
-    # Setup default result folder
-    if 'result_folder' not in args:
-        args['result_folder'] = str(CFG['result_dir']) + "/" + args['project']
-        print("Result folder defaults to: {res_folder}".format(
-            res_folder=args['result_folder']))
+    if args['plot_type'] is not PlotTypes.paper_config_overview_plot:
```
vulder: why do you need this? The Idea behind the lazy init is that it does not matter if the plot needs a cmap or not. Do you get any special requirements if you include this?

varats/driver.py
```diff
@@ -250,14 +250,20 @@ def main_gen_graph():
...
+        # Setup default result folder
+        if 'result_folder' not in args:
+            args['result_folder'] = str(CFG['result_dir']) + "/" + args['project']
+            print("Result folder defaults to: {res_folder}".format(
+                res_folder=args['result_folder']))
+    else:
+        if 'result_folder' not in args:
+            args['result_folder'] = str(CFG['result_dir'])
+            print("Result folder defaults to: {res_folder}".format(
+                res_folder=args['result_folder']))
```
vulder: We should merge these to result folder setups, and only choose if we want to append the `args['project']`

varats/plots/paper_config_overview.py
```diff
@@ -0,0 +1,153 @@
...
+from varats.utils.project_util import get_local_project_git_path
+
+
+def _gen_overview_plot():
+    """
+    Generate the data for the PaperConfigOverviewPlot.
+    """
+    PC.load_paper_config(
+        str(CFG["paper_config"]["folder"]) + "/" +
+        str(CFG["paper_config"]["current_config"]))
```
vulder: Should we move this, config based, paper config path into the load_paper_config function as a default argument? It starts to look redundant to pass the default config path.

varats/plots/paper_config_overview.py
```diff
@@ -0,0 +1,153 @@
...
+
+    for case_study in sorted(
+            current_config.get_all_case_studies(),
+            key=lambda cs: (cs.project_name, cs.version)):
+        processed_revisions = list(
+            dict.fromkeys(case_study.processed_revisions(CommitReport)))
+
+        git_path = get_local_project_git_path(case_study.project_name)
+        repo_path = pygit2.discover_repository(str(git_path))
+        repo = pygit2.Repository(repo_path)
```
vulder, line 41: Maybe we should start a small git utils module in utils that makes repo handling easier, so we do not have to call discover in all places. As far as I understand it discover is mandatory before loading a repository.

varats/plots/paper_config_overview.py
```diff
@@ -0,0 +1,153 @@
...
+
+        plt.savefig(
+            result_dir / ("{graph_name}.{filetype}".format(
+                graph_name=self.name, filetype=filetype)),
+            dpi=1200,
+            bbox_inches="tight",
+            format=filetype)
+
+    def calc_missing_revisions(self, boundary_gradient) -> tp.Set:
+        return set()
```
vulder: Maybe throwing NotImplemented here would be better, this way the failure is silent.

varats/plots/paper_config_overview.py
```diff
@@ -0,0 +1,153 @@
...
+        super(PaperConfigOverviewPlot, self).__init__("paper_config_overview_plot")
+        self.__saved_extra_args = kwargs
+
+    def plot(self, view_mode):
+        style.use(self.style)
+        _plot_overview_graph(
+            _gen_overview_plot())
+
+    def show(self):
+        self.plot(False)
```
vulder:
```suggestion
        self.plot(True)
```
currently plot does not use this information but, show is called in view mode so we should at least pass the correct value

vulder (changes requested):
Graph looks good, found just a few nits

varats/driver.py
julianbreiteneicher: I will improve that code. The reason was that `args['project']` doesn't exist, so `create_lazy_commit_map_loader` doesn't work.
Should I simply do `if args['project'] ...`? But then i guess it will also not be executed when you don't supply the -p parameter for an interaction plot?

varats/driver.py
julianbreiteneicher: Good idea. Will do

varats/plots/paper_config_overview.py
julianbreiteneicher: You mean so you can simply call `PC.load_paper_config()` instead? I think thats a good idea

varats/plots/paper_config_overview.py
julianbreiteneicher: Will do

varats/plots/paper_config_overview.py
julianbreiteneicher: Makes sense

varats/plots/paper_config_overview.py
julianbreiteneicher: Not sure if it's really worth it. You have to use the pygit2 API anyways. So the difference is that you either call `pygit2.discover_repository()` or our `foo.load_repository()` or whatever...

vulder:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/VaRA/VaRA-Tool-Suite/pullRequest?prid=3748314&bid=13166186)

varats/driver.py
```diff
@@ -251,12 +251,18 @@ def main_gen_graph() -> None:
         for k, v in vars(parser.parse_args()).items() if v is not None
     }
 
-    args['get_cmap'] = create_lazy_commit_map_loader(args['project'],
-                                                     args.get('cmap', None))
+    if args['plot_type'] is PlotTypes.interaction_plot and 'project' not in args:
+        print("ERROR: interaction_plot requires argument \'project\'")
```
vulder: This should not be needed. By design this checking is done with the @check_required_args(['project'])

did you get an error when creating an interaction plot?

vulder (changes requested):

varats/driver.py
julianbreiteneicher: Correct. I've removed the line.

vulder (approved):

PR #100, opened 2019-06-28
Add support for named stages 

julianbreiteneicher:
resolves se-passau/VaRA#433

They can already be used with the 'per_year_add'
extension stragegy by using the '--revs_year_sep'
parameter.

To update the case-study files to the new format, the following command can be used:
```
find . -name "*.case_study" -exec sed -i '/^- !CSStage/a \ \ _CSStage__name: null' {} \;
```

### DIFF
--- a/varats/paper/case_study.py
+++ b/varats/paper/case_study.py
@@ -236,11 +236,14 @@ class CaseStudy(yaml.YAMLObject):
             for _ in range(abs(offset)):
                 self.__stages.pop(remove_index)
 
-    def insert_empty_stage(self, pos: int) -> None:
+    def insert_empty_stage(self, pos: int) -> CSStage:
         """
         Insert a new stage at the given index, shifting the list elements to the right.
+        The newly created stage is returned.
         """
-        self.__stages.insert(pos, CSStage())
+        new_stage = CSStage()
+        self.__stages.insert(pos, new_stage)
+        return new_stage
 
     def include_revision(self,
                          revision: str,
--- a/varats/paper/case_study.py
+++ b/varats/paper/case_study.py
@@ -222,8 +222,10 @@ class CaseStudy(yaml.YAMLObject):
         Shift a stage in the case-studie's stage list by an offset.
         Beware that shifts to the left (offset<0) will destroy stages.
         """
-        assert 0 <= from_index < len(self.__stages)
-        assert from_index + offset >= 0, "Shifting out of bounds"
+        if not (0 <= from_index < len(self.__stages)):
+            raise AssertionError("from_index out of bounds")
+        if (from_index + offset) < 0:
+            raise AssertionError("Shifting out of bounds")
 
         if offset > 0:
             for _ in range(offset):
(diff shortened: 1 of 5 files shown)


julianbreiteneicher:
@vulder: New revisions are now inserted at the correct position of the stages
list.

vulder:


```diff

Issues
======
- Added 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/VaRA/VaRA-Tool-Suite/pullRequest?prid=3785060&bid=13264240)

vulder:


```diff

Issues
======
- Added 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/VaRA/VaRA-Tool-Suite/pullRequest?prid=3785060&bid=13264240)

varats/paper/paper_config_manager.py
```diff
@@ -139,8 +139,13 @@ def color_rev_state(rev_state: str) -> str:
         return rev_state
 
     if sep_stages:
+        stages = case_study.stages
         for stage_num in range(0, case_study.num_stages):
-            status += "  Stage {idx}\n".format(idx=stage_num)
+            status += "  Stage {idx}".format(idx=stage_num)
+            stage_name = stages[stage_num].name
```
vulder, line 145:
```
            stage_name = stages[stage_num].name
```
this line could be move into the if

varats/paper/case_study.py
```diff
@@ -180,6 +217,31 @@ def has_revision_in_stage(self, revision: str, num_stage: int) -> bool:
...
+        if offset > 0:
+            for _ in range(offset):
+                self.__stages.insert(from_index, CSStage())
+
+        if offset < 0:
+            remove_index = from_index + offset
+            for _ in range(abs(offset)):
+                self.__stages.pop(remove_index)
+
+    def insert_empty_stage(self, pos: int) -> None:
```
vulder:
```suggestion
    def insert_empty_stage(self, pos: int) -> CSStage:
```

vulder: maybe we should directly return the created stage, so it becomes easier for the user to interact with it.
E.g., this is always tedious in std::vector code where you `push_back()` and then afterwards need to call vec.back() to interact with the newly created element

vulder (changes requested):

varats/paper/paper_config_manager.py
julianbreiteneicher: I don't understand what you mean. stage_name is used two times here

varats/paper/case_study.py
julianbreiteneicher: I think this would contradict the `stages` getter of the `CaseStudy` class for example,
because the getter creates a copy of the stages list and returns that, so that the internal stage of the `CaseStudy` cannot be modified directly from the outside.

vulder: I don't think it does. `.stages` only creates a new list so the caller cannot manipulate the internal state of the stages list but the stages are not copied, they are references.

varats/paper/paper_config_manager.py
vulder: ah yeah, it's also used in the condition

varats/paper/case_study.py
julianbreiteneicher: OK, in this case, it should be ok.
I'll change that

julianbreiteneicher: Done

vulder (approved):

PR #101, opened 2019-07-02
Warn if UnaryFilter is used w/o Source-/TargetOperator

julianbreiteneicher:
Example:
2019-07-02-235643_600x657_scrot

### DIFF
--- a/uicomponents/FilterProperties.ui
+++ b/uicomponents/FilterProperties.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>285</width>
-    <height>196</height>
+    <width>289</width>
+    <height>255</height>
    </rect>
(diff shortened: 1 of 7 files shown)


vulder:
maybe: `Currently this filter applies to source and target, therefore, only interactions where both satisfy the filter are included.`

vulder (commented):
maybe: `Currently this filter applies to source and target, therefore, only interactions where both satisfy the filter are included.`

vulder (approved):

julianbreiteneicher:
> maybe: `Currently this filter applies to source and target, therefore, only interactions where both satisfy the filter are included.`

Sure, I could easily change that...?

julianbreiteneicher:
Done.
2019-07-04-231318_633x694_scrot

PR #112, opened 2019-07-29
Enable custom file types for successful reports

labrenz:
Allows different reports to specify their own file type extension, while vara-cs still recognizes them as successful runs and paints their statuses accordingly.


tests/data/test_report.py
```diff
@@ -31,7 +31,7 @@ def test_is_result_file(self):
             MetaReport.is_result_file(self.success_filename.replace("_", "")))
         self.assertFalse(
             MetaReport.is_result_file(self.success_filename.replace("-", "")))
-        self.assertFalse(
+        self.assertTrue(
```
vulder: you can remove this, we don't need to check if a filename keeps being correct after we did no change

vulder (changes requested):
only found a small nit

we also need a small bash command to migrate the data like we did before probably something similar like this should do:
```
cd vara/results/
rename -v '.yaml' '_success.yaml' **/*
```
what other renames do we also need?
PS: `-n` let's rename run without changing files, for local testing

tests/data/test_report.py
labrenz: All right, i will provide a small bash script to update the old data.
I figured the tests was pointless, yet was to scared to remove it, but i will do so, the updates should follow in a commit tomorrow.

labrenz: Ok i added the commits, thanks for the tip with rename.
It would have been much more effort with sed, xargs and all that pre-installed bash stuff.

However the path to the vara results is currently hard encoded, and i am not quite happy with that yet.

labrenz: The call for changing the names is: ``rename -n 's/.yaml/_sucess.yaml/' **/*``
You have to be in the results directory first, in my case you can get there by: ``cd ~/vara/results``

I will also remove the script, i uploaded.

labrenz: I just removed the script since it was faulty and also should not be in the repo.
However, to rename all old files, run these commands:
 ```
cd vara/results/
rename 's/.yaml/_sucess.yaml/' **/*
rename 's/.failed/_failed.yaml/' **/*
rename 's/.cerror/_cerror.yaml/' **/*
```
**Make sure you run the command renaming the existing yamls first**, otherwise it will also transform already renamed, which then would have multiple extensions.

**Comment by vulder** [comment]
@labrenz can you please post the update script, you want users to apply for fixing, here so I can add it to the commit when merging?

**Comment by labrenz** [comment]
As mentioned in the conversation earlier:
> I just removed the script since it was faulty and also should not be in the repo.
> However, to rename all old files, run these commands:
> 
```
cd vara/results/
rename 's/.yaml/_sucess.yaml/' **/*
rename 's/.failed/_failed.yaml/' **/*
rename 's/.cerror/_cerror.yaml/' **/*
``` 
**Make sure you run the command renaming the existing yamls first**, otherwise it will also transform the already renamed files, which then would have multiple extensions.

**Comment by vulder** [comment]
@labrenz ah, sry I somehow missed that

**Comment by labrenz** [comment]
Np

**Comment by vulder** [comment]


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/VaRA/VaRA-Tool-Suite/pullRequest?prid=3927888&bid=13642747)

vulder:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/VaRA/VaRA-Tool-Suite/pullRequest?prid=3927888&bid=13642747)

vulder (approved):

PR #121, opened 2019-09-06
Bump Benchbuild version to 4.0.1

julianbreiteneicher:
related se-passau/VaRA#384
related se-passau/VaRA-Tool-Suite#111


vulder (approved):
ah, nice then we can close the other PRs

PR #125, opened 2019-09-18
Set error file extension to .txt

labrenz:
Each experiment can define own file types for its compile errors, failed runs and successful runs.
The commit_reports still used .yaml for all of them, but now only uses .yaml for successful runs and .txt otherwise.


labrenz:
Fixed typing and stayed with the custom file ending one has to provide for call of a report generation.
This way other error reports can be used. If no extension is provided the default file extension provided in the report is used as file extension, regardless of the status of the run that created the report.

vulder (approved):

PR #139, opened 2019-10-22
Adds missing dependency to libyaml-dev in README.md

boehmseb:
To use pyyaml's CLoader, the package `libyaml-dev` is needed on debian based systems according to https://github.com/yaml/pyyaml/issues/108.
This PR adds this package to the list of install dependencies for the tool suite.


vulder (approved):
lgtm

PR #144, opened 2019-10-28
Disables output on same line feature by default

vulder:
Using buildsetup with automatically resetting printed lines makes it sometimes hard to detect bugs and to debug them, therefore, we disable same line printing by default.

### DIFF
--- a/varats/driver.py
+++ b/varats/driver.py
@@ -101,15 +101,24 @@ def main_graph_view() -> None:
     driver.main()
 
 
-def update_term(text: str) -> None:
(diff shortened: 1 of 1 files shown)


boehmseb:
Will there be a way to change the default?

vulder:
Currently I'm not planning to do it but we could add it later, therefore, I wanted to keep the functionality.
Up to now people only have complaint about the feature, so, I guess, nobody will miss it :laughing:

boehmseb (approved):

PR #180, opened 2019-11-28
Blocks failing revisions in gravity

boehmseb:
All blocks are there due to a bug in gravity or its build system. No blocks due to missing dependencies/environment, so I will not open an accompanying issue.


vulder (approved):
looks fine

PR #185, opened 2020-01-14
Minor bugfix

MarcJose:
Fixes warning when comparing with zero and a print format error.

### DIFF
--- a/varats/vara_manager.py
+++ b/varats/vara_manager.py
@@ -318,7 +318,7 @@ def setup_vara(init: bool,
 
     if init:
         if os.path.exists(llvm_folder):
-            print("LLVM was already checked out in '{}'.".format(llvm_folder))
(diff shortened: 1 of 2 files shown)


varats/vara_manager.py
```diff
@@ -318,7 +318,7 @@ def setup_vara(init: bool,
 
     if init:
         if os.path.exists(llvm_folder):
-            print("LLVM was already checked out in '%s'.", llvm_folder)
+            print("LLVM was already checked out in '{}'.".format(llvm_folder))
```
vulder:
```suggestion
            print(f"LLVM was already checked out in '{llvm_folder}'.")
```

vulder: a nice, when you start fixing things
that should also work :)
using python3.6s new f strings

vulder (changes requested):

vulder (approved):
lgtm

PR #189, opened 2020-01-29
Update dep to new matplotlib

vulder:
This update removes the broken seaborn plot.

resolves se-passau/VaRA#480


boehmseb (approved):
lgtm

PR #206, opened 2020-02-25
Polishes version header documentation

vulder:
related se-passau/VaRA#571

### DIFF
--- a/varats/data/version_header.py
+++ b/varats/data/version_header.py
@@ -62,7 +62,7 @@ class VersionHeader():
     def from_version_number(cls, doc_type: str,
                             version: int) -> 'VersionHeader':
         """
-        Creates a new VersionHeader object from a doc_type string and a
+        Creates a new VersionHeader object from a ``doc_type`` string and a
         version number.
 
         Args:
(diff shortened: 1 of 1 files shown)


varats/data/version_header.py
```diff
@@ -70,13 +78,21 @@ def doc_type(self) -> str:
         return self.__doc_type
 
     def is_type(self, type_name: str) -> bool:
-        """Checks if the type of the following yaml file is type_name."""
+        """
+        Checks if the type of the following yaml file is type_name.
```
boehmseb: `type_name` in double-backticks?

varats/data/version_header.py
```diff
@@ -60,6 +64,10 @@ def from_version_number(cls, doc_type: str,
         """
         Creates a new VersionHeader object from a doc_type string and a
```
boehmseb: `doc_type` in double-backticks?

boehmseb (approved):

varats/data/version_header.py
vulder:
```suggestion
        Creates a new VersionHeader object from a ``doc_type`` string and a
```

varats/data/version_header.py
vulder:
```suggestion
        Checks if the type of the following yaml file is ``type_name``.
```

boehmseb (approved):

PR #262, opened 2020-04-24
Adds compression to cached data frames

boehmseb:
Resolves se-passau/VaRA#631


vulder (approved):
lgtm
but you have to fix the test

PR #267, opened 2020-04-29
Adds a file viewing utility

boehmseb:
Resolves se-passau/VaRA#489

### DIFF
--- a/varats/data/revisions.py
+++ b/varats/data/revisions.py
@@ -88,12 +88,12 @@ def __get_supplementary_result_files_dict(
     return result_files
 
 
-def __get_files_with_status(
-    project_name: str,
-    result_file_type: MetaReport,
-    file_statuses: tp.List[FileStatusExtension],
-    file_name_filter: tp.Optional[tp.Callable[[str], bool]] = None
-) -> tp.List[Path]:
+def __get_files_with_status(project_name: str,
+                            result_file_type: MetaReport,
+                            file_statuses: tp.List[FileStatusExtension],
+                            file_name_filter: tp.Callable[
+                                [str], bool] = lambda x: False,
+                            only_newest: bool = True) -> tp.List[Path]:
     """
         Find all file paths to revision files with given file statuses.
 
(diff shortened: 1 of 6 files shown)


vulder:


```diff

Issues
======
- Added 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/gh/se-passau/VaRA-Tool-Suite/pullRequest?prid=5440070&bid=17839875)

varats/data/revisions.py
```diff
@@ -88,40 +88,69 @@ def __get_supplementary_result_files_dict(
...
+                          true if the file_name should not be checked
+        only_newest: whether to include all result files, or only the newest
+
+    Returns:
+        a list of file paths to correctly processed revision files
+    """
+    return __get_files_with_status(project_name, result_file_type, [
+        FileStatusExtension.Success, FileStatusExtension.Failed,
+        FileStatusExtension.CompileError
+    ], file_name_filter, only_newest)
```
vulder: It could make sense to implement a function on `FSE` to return all FSEs, which represent `real` files compared to extensions like missing or blocked.
This allows us to adapted the list when caching FSE without having to find all locations where something like this is used.

But not sure what a good name for this is.

varats/tools/driver_casestudy.py
```diff
@@ -290,5 +304,33 @@ def __casestudy_package(args: tp.Dict[str, tp.Any],
...
+        f"Found {len(result_files)} matching result files (newest to oldest):")
+    for idx, result_file in enumerate(result_files):
+        status = (result_file_type.get_status_from_result_file(
+            result_file.name)).get_colored_status()
+        print(f"  {idx}. [{status}] {result_file.name}")
+    try:
+        choice = ""
+        while not choice.startswith("q"):
+            choice = input("Enter a number to open a file or 'q' to quit "
+                           "(default=0): ")
```
vulder: Does it make sense to extract this to cli_utils? I have a similar function there for y/n choices

varats/tools/driver_casestudy.py
```diff
@@ -290,5 +304,33 @@ def __casestudy_package(args: tp.Dict[str, tp.Any],
...
+            if not choice:
+                choice = "0"
+            if choice.isdigit() and int(choice) < len(result_files):
+                editor_name = local.env["EDITOR"]
+                if not editor_name:
+                    editor_name = "vim"
+                editor = local[editor_name]
+                editor[str(result_files[int(choice)])] & FG
```
vulder: add auto quit if now files where found

varats/tools/driver_casestudy.py
```diff
@@ -186,6 +185,19 @@ def add_common_args(sub_parser: ArgumentParser) -> None:
...
+                             help="Report type of the result files.",
+                             choices=MetaReport.REPORT_TYPES.keys(),
+                             type=str)
+    view_parser.add_argument("project",
+                             help="Project to view result files for.")
+    view_parser.add_argument("commit_hash",
+                             help="Commit hash to view result files for.")
+    view_parser.add_argument("--result-folder",
+                             help="Folder in which to search for result files.")
+
```
vulder: is this currently used?
PS: in case of the impl it should default to the CFG results folder

varats/paper/paper_config_manager.py
```diff
@@ -97,9 +99,23 @@ def get_revision_list(case_study: CaseStudy) -> str:
...
-def get_occurrences(
-        status_occurrences: tp.DefaultDict[FileStatusExtension, tp.Set[str]],
-        use_color: bool = False) -> str:
+def get_result_files(result_file_type: MetaReport, project_name: str,
+                     commit_hash: str) -> tp.List[Path]:
+
+    def file_name_filter(file_name: str) -> bool:
+        file_commit_hash = MetaReport.get_commit_hash_from_result_file(
+            file_name)
+        return commit_hash != file_commit_hash
```
vulder: Currently the hash needs to be a identical match, maybe a imprecise check would be better, like `startswith`
So one can type less chars :)

varats/tools/driver_casestudy.py
```diff
@@ -290,5 +304,33 @@ def __casestudy_package(args: tp.Dict[str, tp.Any],
             "Please do not provide any other file type extension than .zip")
 
 
+def __casestudy_view(args: tp.Dict[str, tp.Any]) -> None:
```
vulder, line 321: Should we also add a option to the tool that make the hash optional?
In that case, maybe show all files? OR Show an additional hash selection dialog?
This could make in easier to search for files if one does not 100% know the commit hash.

boehmseb: I fear that this will get confusing quickly for larger case studies. 
One common workflow I have in mind is that one uses `vara-cs status` and finds a commit hash of interest and then uses this tool.
In which situation do you think this could be useful?

vulder: Yeah my idea was that I can do this without looking at `vara-cs status` because for larger case studies this get's quite big.

vulder (approved):

vulder (approved):

PR #288, opened 2020-05-15
Implements the missing logging cases

vulder:
resolves se-passau/VaRA#589

### DIFF
--- a/varats/projects/c_projects/coreutils.py
+++ b/varats/projects/c_projects/coreutils.py
@@ -170,7 +170,7 @@ class Coreutils(prj.Project, CVEProviderHook):  # type: ignore
             run(make["-j", int(BB_CFG["jobs"])])
             for binary in self.binaries:
                 if not Path("{binary}".format(binary=binary)).exists():
-                    print("Shit is fucked up ", binary)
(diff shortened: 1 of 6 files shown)


varats/projects/c_projects/coreutils.py
```diff
@@ -170,7 +170,7 @@ def compile(self) -> None:
             run(make["-j", int(BB_CFG["jobs"])])
             for binary in self.binaries:
                 if not Path("{binary}".format(binary=binary)).exists():
-                    print("Shit is fucked up ", binary)
+                    print("Could not find {binary}")
```
boehmseb, line 173: Shouldn't this also be a warning?

vulder: In general yes but I'm not sure.
How should we handle logging for projects? They are executed with BB so our logger cannot log the warnings. The only options, besides print, is that we get a way to access the BB logger an log it their. Up until now, I always used print for project/experiment specific things.

boehmseb: Hm, then I think it is best to be consistent and leave it as is.

boehmseb (approved):

PR #293, opened 2020-05-20
Extends vara-cs view to allow dynamic commit selection

vulder:


varats/tools/driver_casestudy.py
```diff
@@ -370,10 +371,54 @@ def __casestudy_package(
...
+            available_commit_hashes.extend(
+                case_study.processed_revisions(result_file_type)
+            )
+
+        def set_commit_hash(choice: str) -> None:
+            nonlocal commit_hash
+            commit_hash = choice[:10]
+
+        try:
+            cli_list_choice(
```
boehmseb, line 397: What do you do if there is a case study that contains all revisions of, let's say git?

boehmseb (changes requested):

varats/tools/driver_casestudy.py
vulder: I added a cap to truncate the list

boehmseb (approved):

PR #296, opened 2020-05-25
Extends the Extract interface with error handling

vulder:
The interface of Extract now supports error handling. The caller can
provide a custom error handler to specify how error handling inside the
extract function should be done.


boehmseb (approved):
lgtm

PR #305, opened 2020-05-31
Add option to vara-cs ext to add revisions in a new stage.

boehmseb:
Resolves se-passau/VaRA#504


vulder (approved):
much better alternative to '+' shenanigans

PR #307, opened 2020-06-01
Add short development how-to to docs

boehmseb:


docs/source/index.rst
```diff
@@ -36,23 +36,8 @@ VaRA-TS Docs
 
    vara-ts/vara-buildsetup
    vara-ts/benchbuild
-   vara-ts/testing
```
vulder, line 39: Where did testing go? Did you delete it on purpose?

vulder (approved):

docs/source/index.rst
boehmseb: I merged this with the new `vara-ts/development` as they fit well together and are both quite short on their own.

PR #317, opened 2020-07-01
Changes build folder symlink to a relative path

vulder:
A relative path is better here because we depend less on system
specific parts and only describe inter relationships between project
folders.


alexmeinhold (approved):
Fixes a problem I had with building Vara.

PR #320, opened 2020-07-13
Replaces broken python resource ulimit with prlimit wrapper call

vulder:
Python resource is not able to correctly set the max stacksize in our
cluster environment, therefore, we replace it by a prlimit call to
execute commands with max stack size.

### DIFF
--- a/varats/utils/experiment_util.py
+++ b/varats/utils/experiment_util.py
@@ -138,9 +138,7 @@ def get_default_compile_error_wrapped(
     )
 
 
-def wrap_unlimit_stack_size(
-    cmd: tp.Callable[..., tp.Any]
-) -> tp.Callable[..., tp.Any]:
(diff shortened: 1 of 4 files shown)


varats/utils/experiment_util.py
```diff
@@ -137,23 +138,19 @@ def get_default_compile_error_wrapped(
...
-              self).__init__(obj=project, action_fn=self.__call__)
+    Args:
+        cmd: command that should be executed with max stack size
 
-    def __call__(self) -> StepResult:
-        """Same as 'ulimit -s 16777216' in a shell."""
-        resource.setrlimit(resource.RLIMIT_STACK, (16777216, 16777216))
+    Returns: wrapped command
+    """
+    return prlimit["--stack=17179869184:", cmd]
```
boehmseb: Where does this number come from? A short notice in the docstring would be nice.

boehmseb (changes requested):

boehmseb (approved):

PR #361, opened 2020-09-14
Fix output of vara-pc select and vara-pc list

boehmseb:


vulder:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- tests/tools/test_driver_paper_config.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/se-passau/VaRA-Tool-Suite/pullRequest?prid=6174055&bid=20155736)

varats/varats/paper/paper_config.py
```diff
@@ -241,7 +241,7 @@ def load_paper_config(config_path: tp.Optional[Path] = None) -> None:
         if vara_cfg()["paper_config"]["folder"].value is None or \
                 vara_cfg()["paper_config"][
                     "current_config"].value is None:
-            raise Exception(
+            raise RuntimeError(
```
vulder:
```suggestion
            raise LookupError(
```
Maybe lookup error is better for this case, we try to lookup the paper config and cannot find it.

varats/varats/tools/driver_paper_config.py
```diff
@@ -157,13 +158,14 @@ def set_pc_path(choice: str) -> None:
...
 
+        current_config = None
         try:
-            if is_paper_config_loaded():
-                current_config: tp.Optional[str] = get_loaded_paper_config(
-                ).path.name
-            else:
-                current_config = None
+            current_config = get_paper_config().path.name
+        except RuntimeError:
```
vulder: If changed above, also change here.

varats/varats/tools/driver_paper_config.py
```diff
@@ -210,12 +212,14 @@ def __pc_list(args: tp.Dict[str, tp.Any]) -> None:
...
-    if is_paper_config_loaded():
-        current_config: tp.Optional[str] = get_loaded_paper_config().path.name
-    else:
-        current_config = None
-
-    for paper_config in __get_paper_configs(pc_folder_path):
+    current_config = None
+    try:
+        current_config = get_paper_config().path.name
+    except RuntimeError:
```
vulder: same here

tests/tools/test_driver_paper_config.py
```diff
@@ -0,0 +1,78 @@
...
+    ):
+        """Test the vara-pc select subcommand."""
+
+        stdin.return_value = "1"
+        mock_get_paper_configs.return_value = ["foo", "bar", "baz"]
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tmppath = Path(tmpdir)
+            pc_path = tmppath / "paper_configs"
+            pc_path.mkdir()
+            for pc in mock_get_paper_configs.return_value:
```
vulder, line 70: `pc` in our context makes sense to use as a variable, we should add it to the allowed names in pylint rc.

vulder (changes requested):
Fix looks good, added only a small question related to the exception.
Tests look good but some small thing seems wrong because travis breaks when running tests.

varats/varats/paper/paper_config.py
boehmseb: According to the [documentation](https://docs.python.org/3/library/exceptions.html#LookupError), `LookupError` is used for missing keys or indices in mappings and sequences. That does not quite match our use case.
The "right" way would be to define our own exception, but i don't think it's worth it.

vulder (approved):

PR #409, opened 2020-10-30
Update BlameReportFunctionEntries with new instruction data

vulder:


tests/data/test_blame_report.py
```diff
@@ -255,6 +270,11 @@ def test_demangled_name(self):
             self.func_entry_cxx.demangled_name, 'doStuff(int, int)'
         )
 
+    def test_demangled_name(self):
```
boehmseb: Fix the test name.

boehmseb (changes requested):

boehmseb (approved):

PR #410, opened 2020-11-03
Fix wrong alignment of some statuses in vara-cs view

cormensratio:
Fixes wrong alignment of some statuses in vara-cs view
 closes: #684

### DIFF
--- a/varats-core/varats/report/report.py
+++ b/varats-core/varats/report/report.py
@@ -39,6 +39,10 @@ class FileStatusExtension(Enum):
         color."""
         return tp.cast(str, self.status_color[self.name])
 
+    def num_color_characters(self) -> int:
+        """Returns the number of non printable color characters."""
+        return len(re.sub(self.name, '', self.get_colored_status()))
+
     @staticmethod
     def get_physical_file_statuses() -> tp.Set['FileStatusExtension']:
         """Returns the set of file status extensions that are associated with
--- a/varats-core/varats/report/report.py
+++ b/varats-core/varats/report/report.py
@@ -41,7 +41,7 @@ class FileStatusExtension(Enum):
 
     def num_color_characters(self) -> int:
         """Returns the number of non printable color characters."""
-        return len(re.sub(self.name, '', self.get_colored_status()))
+        return len(self.status_color[''])
 
     @staticmethod
     def get_physical_file_statuses() -> tp.Set['FileStatusExtension']:
--- a/varats/varats/tools/driver_casestudy.py
+++ b/varats/varats/tools/driver_casestudy.py
@@ -436,16 +436,9 @@ def __init_commit_hash(args: tp.Dict[str, tp.Any]) -> str:
         def result_file_to_list_entry(
             commit_status_pair: tp.Tuple[str, FileStatusExtension]
         ) -> str:
-
-            status_name_length = len(commit_status_pair[1].name)
-            status_color_length = len(
-                commit_status_pair[1].get_colored_status()
-            )
-            status_length = status_color_length + \
-                            longest_file_status_extension - status_name_length
-
             status = commit_status_pair[1].get_colored_status().rjust(
-                status_length, " "
+                longest_file_status_extension +
+                commit_status_pair[1].num_color_characters(), " "
             )
 
             return f"[{status}] {commit_status_pair[0][:10]}"
@@ -506,14 +499,12 @@ def __casestudy_view(args: tp.Dict[str, tp.Any]) -> None:
         file_status = result_file_type.get_status_from_result_file(
             result_file.name
         )
-        status_name_length = len(file_status.name)
-
-        status_color_length = len(file_status.get_colored_status())
-
-        status_length = status_color_length + longest_file_status_extension - \
-                        status_name_length
-
-        status = (file_status.get_colored_status().rjust(status_length, " "))
+        status = (
+            file_status.get_colored_status().rjust(
+                longest_file_status_extension +
+                file_status.num_color_characters(), " "
+            )
+        )
         return f"[{status}] {result_file.name}"
 
     def open_in_editor(result_file: Path) -> None:
--- a/varats/varats/tools/driver_casestudy.py
+++ b/varats/varats/tools/driver_casestudy.py
@@ -437,15 +437,15 @@ def __init_commit_hash(args: tp.Dict[str, tp.Any]) -> str:
             commit_status_pair: tp.Tuple[str, FileStatusExtension]
         ) -> str:
 
-            status_length = len(commit_status_pair[1].name)
+            status_name_length = len(commit_status_pair[1].name)
             status_color_length = len(
                 commit_status_pair[1].get_colored_status()
             )
-            padding = status_color_length + longest_file_status_extension - \
-                      status_length
+            status_length = status_color_length + longest_file_status_extension - \
+                      status_name_length
 
             status = commit_status_pair[1].get_colored_status().rjust(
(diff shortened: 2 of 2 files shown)


cormensratio:
The statuses now look like this:
```
9.  [     Success] 24e0406c0f
10. [     Missing] de678e0c92
11. [CompileError] 25fe729532
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -425,17 +425,29 @@ def set_commit_hash(
...
         ])
 
         def result_file_to_list_entry(
             commit_status_pair: tp.Tuple[str, FileStatusExtension]
         ) -> str:
+
+            status_length = len(commit_status_pair[1].name)
+            status_color_length = len(
+                commit_status_pair[1].get_colored_status()
+            )
```
vulder: Why do we need these extra values in the max length?

varats/varats/tools/driver_casestudy.py
```diff
@@ -425,17 +425,29 @@ def set_commit_hash(
...
+            padding = status_color_length + longest_file_status_extension - \
+                      status_length
+
             status = commit_status_pair[1].get_colored_status().rjust(
-                longest_file_status_extension, " "
+                padding, " "
```
vulder: I would not call this padding, `rjust` takes the length of the string that should be returned and not a padding size.

varats/varats/tools/driver_casestudy.py
```diff
@@ -482,15 +494,26 @@ def __casestudy_view(args: tp.Dict[str, tp.Any]) -> None:
...
+            result_file.name
+        )
+        status_length = len(file_status.name)
+
+        status_color_length = len(file_status.get_colored_status())
+
+        padding = status_color_length + longest_file_status_extension - \
+                  status_length
+
+        status = (file_status.get_colored_status().rjust(padding, " "))
```
vulder: Same as above.

vulder (commented):
Overall, this seems to fix the bug but I'm not sure if does not introduces other problems, see comments attached.

varats/varats/tools/driver_casestudy.py
cormensratio: The selected code part confuses me in combination with the question, but I assume you mean why I included the virtual file statues in the max length, too.
In the previous implementation we were kind of "lucky" that the longest status name (without color string) was included in the physical file statuses. If it would have been located in the virtual file statuses, we would have missed the status name with the actual longest string. Since both, virtual and physical statuses, can be displayed in vara-cs view, I searched for the longest string in total.
To give more details on the bug:
The main problem was that the color code string of "plumbum.colors.orange3" of the "Missing" status, was longer than any  other color code of the remaining statuses, but the plain word "Missing" (without the color string) is not the longest word, but "CompileError" currently is.
The code part referenced by you is used to get the correct alignment for every status, by considering the plain word length and the length of its color string.

vulder: No I did not. That one seems correct to fix the observed bug and is the reason why `Missing` was not correctly aligned.

But do do we have to consider both when we just can take the one that is always longer?
I mean, your code actually does this in a sense.
1) You use the length of the name
2) to compute the status_length, you use the `colored_status` length add the longest `longest_file_status_extensions` and subtract the name

Remark: The length of a colored output is always >= the length of just the name (i.e., extra color tokens can only increase the length not shorten it). Therefore, if we compute the max based on the `colored_status` we always have the size of the biggest possible status extension. This in combination with how `rjust` works should be enough to always add enough padding.

cormensratio: The addition of the virtual_file_statuses alone was not the solution to the 'Missing' bug.

If I understand you correctly you mean that we should add the virtual_file_statuses to our search of the maximum and change the code for the ``longest_file_status_extension`` to:

longest_file_status_extension = max([
            len(status.**get_colored_status()**) for status in statuses
        ])

and use this max length as value for the ``rjust``.
The output for this scenario is the following:
```
9.  [      Success] 24e0406c0f
10. [Missing] de678e0c92
11. [ CompileError] 25fe729532
```
The explanation for this misalignment is that the "Missing" tag has the longest color string but as soon as the color string is shown as an actual color, it shrinks to the size of the word "Missing". Regarding the other colors, let's assume that the 'Missing' color string is **n** chars long, and the 'Success' color string is **n-6** chars long. 'Missing' and 'Success' have the same word length. When getting displayed in color, the 'Success' string shrinks to its word length but since `rjust` got the value **n**, the remaining 6 chars get displayed as whitespace padding.
Therefore i think we have to consider both, the length of the word and the length of the color string.

vulder: So only using the non color version is also not enough?

PS: But even when we need to consider both we could put this in the max calc for `longest_file_status_extension`? So, like adding a max(.get_clored, name) into the loop to compute everything in one place.

cormensratio: I do not think that it is enough, because by only using the non color version we would ignore the statuses that are longer in their word length than 'Missing' or 'Success'. E.g., just using the word length for `rjust` leads to:
```
9.  [Success] 24e0406c0f
10. [Missing] de678e0c92
11. [CompileError] 25fe729532
```
I do not quite understand why we should put both in the loop, because the loop currently calculates the maximum word length of all statuses and we do not need to know the maximum status color length. We need to know the specific length for every status by considering its own `status_name_length`, its own `status_color_length` and the length of the longest status name.
But maybe it would be beneficial for the performance if we move the calculation of the `rjust` value under the `longest_file_status_extension`-calculation and create a dict, which saves the resulting `rjust`-values in relation to the `FileStatusExtension`s. By doing that we would not need to access the `commit_status_pair`s each time and could pass `rjust` the correct value by accessing the precalculated dict.

vulder: I took a closer look and debugged it my self. The issue I have with the `status_name_length` code is that it's not straight forward explaining what it does but just computes a relative metric that get's the right number in the end, solving the problem but making it hard to understand.
The problem at hand is that we need to compute the maximum name length for `rjust` to align the `status_names` but we need to adjust the `max` amount by the number of non printable characters, specific to the color used. So that in the end visible letters are aligned (but to different bases).
From the calculation of computing the maximum, subtracting the own name and readding the colored name, we get the correct value but it's hard to figure out why.

My proposed adaption:
* max the `status.name` as is
* extend the `FileStatusExtension` class with a method `num_color_characters` (which computes the number of non printable color characters)
* adapt `rjust` to `.rjust(longest_file_status_extension + commit_status_pair[1].num_color_characters() ," ")`

This should make it clearer that we need to adapt the maximum because of the non printable color characters.

cormensratio: I agree. This makes it clearer and easier to read :+1: .

varats-core/varats/report/report.py
```diff
@@ -39,6 +39,10 @@ def get_colored_status(self) -> str:
         color."""
         return tp.cast(str, self.status_color[self.name])
 
+    def num_color_characters(self) -> int:
+        """Returns the number of non printable color characters."""
+        return len(re.sub(self.name, '', self.get_colored_status()))
```
vulder:
```suggestion
        return len(self.status_color[''])
```
It's easier to just use the color and a empty string to get the non visible colors

varats/varats/tools/driver_casestudy.py
vulder: Looks nice, thx :+1:

PR #411, opened 2020-11-12
Table for displaying bug metrics

SimonFedick:
Adds a new Table subclass BugOverviewTable able to display bug data for a given project.

### DIFF
--- a/varats/varats/projects/test_projects/bug_provider_test_repos.py
+++ b/varats/varats/projects/test_projects/bug_provider_test_repos.py
@@ -1,0 +1,47 @@
+"""Projects in vara-test-repos used for testing the bug provider."""
+import typing as tp
+
+import benchbuild as bb
+from plumbum import local
+
+from varats.paper_mgmt.paper_config import project_filter_generator
+from varats.project.project_util import (
+    VaraTestRepoSource,
+    ProjectBinaryWrapper,
+    wrap_paths_to_binaries,
+    BinaryType,
+)
+
+
+class BasicTestRepo(bb.Project):  # type: ignore
+    """Example project demonstrating how to use a repo from the vara-test-
+    repos."""
+
+    NAME = 'basic_test_repo'
+    DOMAIN = 'testing'
+    GROUP = 'test_projects'
+
+    SOURCE = [
+        VaraTestRepoSource(
+            remote="BugDetectionRepos/BasicTestRepo",
+            local="basic_test_repo",
+            refspec="HEAD",
+            limit=None,
+            version_filter=project_filter_generator("basic_test_repo")
+        )
+    ]
+
+    @property
+    def binaries(self) -> tp.List[ProjectBinaryWrapper]:
+        return wrap_paths_to_binaries([("main", BinaryType.executable)])
+
+    def run_tests(self) -> None:
+        pass
+
+    def compile(self) -> None:
+        """Compile the example project."""
+        source = local.path(self.source_of_primary)
+
+        c_compiler = bb.compiler.cc(self)  # type: ignore
+        with local.cwd(source):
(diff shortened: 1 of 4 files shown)


vulder (commented):
looks good :+1: 
a small test would help, we should talk on Monday about creating a small test repository with bug tags, which we can use for testing.

varats/varats/tables/bug_overview_table.py
```diff
@@ -0,0 +1,67 @@
...
+        bug_df = pd.DataFrame(columns=variables, data=np.array(data_rows))
+
+        if self.format in [
+            TableFormat.latex, TableFormat.latex_raw, TableFormat.latex_booktabs
+        ]:
+            tex_code = bug_df.to_latex(bold_rows=True, multicolumn_format="c")
+            return str(tex_code) if tex_code else ""
+        return tabulate(bug_df, bug_df.columns, self.format.value)
+
+    def save(
```
boehmseb: You do not need to override `save()` here. You only need to do that in certain cases (e.g., when your table uses multiple projects at the same time). But in your case, the default `save()` should suffice.

varats/varats/tables/bug_overview_table.py
```diff
@@ -0,0 +1,67 @@
...
+
+    NAME = "bug_overview"
+
+    def __init__(self, **kwargs: tp.Any):
+        super().__init__(self.NAME, **kwargs)
+
+    def tabulate(self) -> str:
+        project_name = self.table_kwargs["project"]
+
+        with BugProvider.create_provider_for_project(
```
boehmseb: Why are you using a context manager here? Providers are not implemented as such and do not need to be. I don't even know what the defined behaviour is in that case.

varats/varats/tables/bug_overview_table.py
```diff
@@ -0,0 +1,67 @@
...
+            get_project_cls_by_name(project_name)
+        ) as provider:
+            if provider:
+                bug_provider = provider
+            else:
+                bug_provider = BugProvider.create_default_provider(
+                    get_project_cls_by_name(project_name)
+                )
```
boehmseb: You do not need to do this. Just call `BugProvider.get_provider_for_project(...)`. This takes care of constructing a default provider if necessary. That's why we have them in the first place.

boehmseb (changes requested):

varats/varats/projects/test_projects/bug_provider_test_repos.py
```diff
@@ -0,0 +1,47 @@
...
+from varats.paper_mgmt.paper_config import project_filter_generator
+from varats.project.project_util import (
+    VaraTestRepoSource,
+    ProjectBinaryWrapper,
+    wrap_paths_to_binaries,
+    BinaryType,
+)
+
+
+class BasicTestRepo(bb.Project):  # type: ignore
```
boehmseb: Please choose a name (class name and project `NAME`) that reflects that this project is for testing bug detection. In many situations, the project name appears without the module name and then it will be hard to distinguish this from other test projects.

boehmseb (changes requested):
Please have a look at the name of the test project. Otherwise looks good.

varats/varats/projects/test_projects/bug_provider_test_repos.py
SimonFedick: Is `BasicBugDetectionTestRepo` alright or would that be too long of a name?

boehmseb: Should be fine. We have test projects with longer names ;)

boehmseb (approved):

vulder (approved):

PR #413, opened 2020-11-13
Add new plot types for multiple libraries: b_interaction_degree_multi_lib, b_interaction_fraction_overview, b_library_interactions

cormensratio:
Adds new plot types to improve the analyzability of library interactions between multiple libraries.
Adds: b_interaction_degree_multi_lib, b_interaction_fraction_overview, b_multi_lib_interaction_sankey_plot


varats/varats/data/reports/blame_report.py
```diff
@@ -521,6 +516,80 @@ def generate_degree_tuples(
...
+def generate_lib_dependent_degrees(
+    report: tp.Union[BlameReport, BlameReportDiff]
+) -> tp.Dict[str, tp.List[tp.Tuple[int, int]]]:
+    """
+    Args:
+        report: The blame report
+
+    Returns:
+        Lists of tuples (degree, amount) categorised by their corresponding
+        library name.
```
vulder: Map of ...

varats-core/varats/utils/git_util.py
```diff
@@ -475,8 +475,9 @@ def calc_code_churn(
     if not churn_config.include_everything:
         diff_base_params.append("--")
         # builds a regrex to select files that git includes into churn calc
+        diff_base_params.append(":")
```
cormensratio: For my system it works if there is either a whitespace after the ":" or a "." instead of the ":", followed by a whitespace.

varats/varats/data/databases/blame_interaction_degree_database.py
```diff
@@ -93,36 +110,96 @@ def create_data_frame_for_report(
...
+                sum_amounts=tp.cast(int, total_max_time_amounts)
+            )
+
+            # Append avg_time rows
+            append_rows_of_degree_type(
+                degree_type=DegreeType.avg_time,
+                degrees=tp.cast(tp.List[int], avg_time_buckets),
+                amounts=tp.cast(tp.List[int], avg_time_amounts),
+                sum_amounts=tp.cast(int, total_avg_time_amounts)
+            )
```
vulder: The idea behind type annotations is that we can verify that expected type is what we get. Casting everything to the type we expect defeats the purpose of type annotations.
If the types do not match, please find out where and why they don't match. In some cases, we can create variables with the correct type, or in other cases we can at least make sure at the point of variable assignment/creation that we have the correct type.

varats/varats/data/reports/blame_report.py
```diff
@@ -516,6 +516,62 @@ def generate_degree_tuples(
...
+
+                if inter_hash_repo_name \
+                        not in tmp_degree_of_libs:
+                    tmp_degree_of_libs[inter_hash_repo_name] = 1
+                else:
+                    tmp_degree_of_libs[inter_hash_repo_name] += 1
+
+            for repo_name, degree in tmp_degree_of_libs.items():
+                if degree not in categorised_degree_dict[base_repo_name][
+                    repo_name]:
```
vulder: Wrong/strange line continuation, warp the condition into `()` so you can easier add a line break.

varats/varats/data/reports/blame_report.py
```diff
@@ -516,6 +516,62 @@ def generate_degree_tuples(
...
+            for inter_hash in interaction.interacting_commits:
+                inter_hash_repo_name = inter_hash.repository_name
+
+                if inter_hash_repo_name \
+                        not in categorised_degree_dict[base_repo_name]:
+                    categorised_degree_dict[base_repo_name][inter_hash_repo_name
+                                                           ] = {}
+
+                if inter_hash_repo_name \
+                        not in tmp_degree_of_libs:
```
vulder:
```suggestion
                if inter_hash_repo_name not in tmp_degree_of_libs:
```

varats/varats/data/reports/blame_report.py
```diff
@@ -516,6 +516,62 @@ def generate_degree_tuples(
...
+            tmp_degree_of_libs: tp.Dict[str, int] = {}
+
+            if not categorised_degree_dict:
+                categorised_degree_dict[base_repo_name] = {}
+
+            for inter_hash in interaction.interacting_commits:
+                inter_hash_repo_name = inter_hash.repository_name
+
+                if inter_hash_repo_name \
+                        not in categorised_degree_dict[base_repo_name]:
```
vulder:
```suggestion
                if (inter_hash_repo_name 
                        not in categorised_degree_dict[base_repo_name]):
```
With `()` you get sometimes better line breaks for long conditions.

PS: run format after applying suggestion

varats/varats/data/reports/blame_report.py
```diff
@@ -516,6 +516,62 @@ def generate_degree_tuples(
...
+) -> tp.Dict[str, tp.Dict[str, tp.List[tp.Tuple[int, int]]]]:
+    """
+    Args:
+        report: The blame report
+
+    Returns:
+        Map of tuples (degree, amount) categorised by their corresponding
+        library name to their corresponding base library name.
+    """
+    categorised_degree_dict: tp.Dict[str, tp.Dict[str, tp.Dict[int, int]]] = {}
```
vulder: Maybe we should add a type def with a describing name for this type. This would make it easier to understand, instead of dict in dict without context.

varats/varats/data/reports/blame_report.py
```diff
@@ -516,6 +516,62 @@ def generate_degree_tuples(
     return list(degree_dict.items())
 
 
+def generate_lib_dependent_degrees(
+    report: tp.Union[BlameReport, BlameReportDiff]
+) -> tp.Dict[str, tp.Dict[str, tp.List[tp.Tuple[int, int]]]]:
+    """
+    Args:
+        report: The blame report
```
vulder:
```suggestion
        report: blame report
```

varats/varats/data/reports/blame_report.py
```diff
@@ -516,6 +516,62 @@ def generate_degree_tuples(
...
+                    degree] += interaction.amount
+
+    # Transform to tuples (degree, amount)
+    result_dict: tp.Dict[str, tp.Dict[str, tp.List[tp.Tuple[int, int]]]] = {}
+    for base_repo_name, inter_lib_dict in categorised_degree_dict.items():
+        result_dict[base_repo_name] = {}
+        for inter_lib_name, degree_amount_dict in inter_lib_dict.items():
+            result_dict[base_repo_name][inter_lib_name] = tp.cast(
+                tp.List[tp.Tuple[int, int]], list(degree_amount_dict.items())
+            )
```
vulder: Why do you need a cast here?

varats/varats/data/databases/blame_interaction_degree_database.py
```diff
@@ -93,36 +110,96 @@ def create_data_frame_for_report(
...
+                for inter_lib_name, degree_amount_tuples in \
+                        inter_lib_dict.items():
+
+                    inter_degrees, inter_amounts = map(
+                        list, zip(*degree_amount_tuples)
+                    )
+
+                    for i, _ in enumerate(inter_degrees):
+                        degree = tp.cast(tp.List[int], inter_degrees)[i]
+                        lib_amount = tp.cast(tp.List[int], inter_amounts)[i]
```
vulder: more casts

varats/varats/data/databases/blame_interaction_degree_database.py
```diff
@@ -56,22 +58,37 @@ def _load_dataframe(
...
             report = load_blame_report(report_path)
-            list_of_degree_occurrences = generate_degree_tuples(report)
-            degrees, amounts = map(list, zip(*list_of_degree_occurrences))
-            total = sum(amounts)
 
-            list_of_author_degree_occurrences = generate_author_degree_tuples(
-                report, commit_lookup
-            )
+            categorised_degree_occurrences = \
+                generate_lib_dependent_degrees(report)
```
vulder:
```suggestion
            categorised_degree_occurrences = generate_lib_dependent_degrees(report)
```
+ format

varats/varats/data/databases/blame_interaction_degree_database.py
```diff
@@ -56,22 +58,37 @@ def _load_dataframe(
...
+
+            def calc_total_amounts() -> int:
+                total = 0
+
+                for _, lib_dict \
+                        in categorised_degree_occurrences.items():
```
vulder:
```suggestion
                for _, lib_dict in categorised_degree_occurrences.items():
```

varats/varats/data/databases/blame_interaction_degree_database.py
```diff
@@ -56,22 +58,37 @@ def _load_dataframe(
...
+                    for _, tuple_list in lib_dict.items():
+                        for degree_amount_tuple in tuple_list:
+                            total += degree_amount_tuple[1]
+                return total
+
+            total_amounts_of_all_libs = calc_total_amounts()
+
+            list_of_author_degree_occurrences = \
+                generate_author_degree_tuples(report, commit_lookup
+                                              )
```
vulder:
```suggestion
            list_of_author_degree_occurrences = generate_author_degree_tuples(report, commit_lookup)
```
+ format

varats-core/varats/utils/git_util.py
vulder: Can you please find out why you need a white space here? From the git spec and bash this is not required an the select should. We need to figure out why this is problematic because if not, we don't now which other systems we break if we change something here.

varats/varats/data/reports/blame_report.py
```diff
@@ -516,6 +516,62 @@ def generate_degree_tuples(
     return list(degree_dict.items())
 
 
+def generate_lib_dependent_degrees(
```
vulder, line 522: We should add a few tests for this to make sure that the data is correctly calcualted.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+            plt.setp(
+                x_axis.get_xticklabels(),
+                fontsize=plot_cfg['xtick_size'],
+                fontfamily='monospace',
+                rotation=270
+            )
+
+        _gen_multi_lib_degree_plot()
+
+    def _fraction_overview_plot(
```
vulder, line 758: I think you put some of the local functions, declared local in this func, as private functions into the class. The function is quite long and hard to understand. Maybe we could also split up this function a bit more, it's quite long.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+                                       ] = _get_distinct_base_lib_names(
+                                           dataframes_per_revision[revision]
+                                       )
+            inter_lib_names_per_revision[revision
+                                        ] = _get_distinct_inter_lib_names(
+                                            dataframes_per_revision[revision]
+                                        )
+
+        def _calc_fractions() -> tp.Tuple[tp.Dict[str, tp.List[float]], tp.Dict[
+            str, tp.List[float]]]:
```
vulder: Maybe, the long type should have a name.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+        for idx, row in interaction_plot_df.iterrows():
+            total_amount = summed_df['amount'].loc[row['revision']]
+            interaction_plot_df.at[idx,
+                                   'fraction'] = row['amount'] / total_amount
+
+        unique_revisions, sub_df_list = _filter_data_frame(
+            degree_type, interaction_plot_df, commit_map
+        )
+
+        def _gen_multi_lib_degree_plot() -> None:
```
vulder: Maybe this function could be moved out.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+                        ].amount.sum(), total_amount_per_revision[rev]
+                    )
+                    inter_fractions[inter_name].append(current_fraction)
+
+            return base_fractions, inter_fractions
+
+        base_lib_fractions, inter_lib_fractions = _calc_fractions()
+
+        def _collect_plot_data(
+        ) -> tp.Tuple[tp.List[tp.List[float]], tp.List[tp.List[float]]]:
```
vulder: same here

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+            if with_churn:
+                out_axis = fig.add_subplot(grid_spec[0, :])
+                out_axis.get_xaxis().set_visible(False)
+                in_axis = fig.add_subplot(grid_spec[1, :])
+                in_axis.get_xaxis().set_visible(False)
+                churn_axis = fig.add_subplot(grid_spec[-1, :], sharex=out_axis)
+                x_axis = churn_axis
+            else:
+                out_axis = fig.add_subplot(grid_spec[0, :])
+                in_axis = fig.add_subplot(grid_spec[1, :])
```
vulder: can be move out of the if, the same condition is in the then branch

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+            return base_lib_fractions_list, inter_lib_fractions_list
+
+        base_plot_fraction_data, inter_plot_fraction_data = _collect_plot_data()
+
+        def _gen_fraction_overview_plot() -> None:
+            fig = plt.figure()
+            grid_spec = fig.add_gridspec(3, 1)
+
+            if with_churn:
+                out_axis = fig.add_subplot(grid_spec[0, :])
```
vulder: see below

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+                out_axis.get_xaxis().set_visible(False)
+                in_axis = fig.add_subplot(grid_spec[1, :])
```
vulder: see below

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+
+            ingoing_plot_lines += in_axis.stackplot(
+                unique_revisions,
+                inter_plot_fraction_data,
+                linewidth=plot_cfg['linewidth'],
+                colors=colormap,
+                edgecolor=plot_cfg['edgecolor'],
+                alpha=0.7
+            )
+            legend_in = in_axis.legend(
```
vulder: maybe setting up the legend could be a separate function

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+            plt.setp(
+                x_axis.get_xticklabels(),
+                fontsize=plot_cfg['xtick_size'],
+                fontfamily='monospace',
+                rotation=270
+            )
+
+        _gen_fraction_overview_plot()
+
+    def _library_interactions(
```
vulder: The name could be clearer, "what" library interactions? Plotting, calculating, ...?

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+
+            name_dict["all_distinct_lib_names"] = sorted(
+                set(name_dict["all_lib_names"])
+            )
+            return name_dict
+
+        lib_name_dict = _gen_lib_name_dict(interaction_plot_df)
+
+        def _build_color_mappings(
+        ) -> tp.Tuple[tp.Dict[str, tp.Any], tp.Dict[str, tp.Dict[int, str]]]:
```
vulder: looks like the same large type

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -185,6 +206,567 @@ def _degree_plot(
...
+            highest_degree = interaction_plot_df["degree"].max()
+
+            if len(lib_name_dict["all_distinct_lib_names"]) > num_colormaps:
+                LOG.warning(
+                    "Not enough colormaps for all libraries provided. "
+                    "Colormaps will be reused."
+                )
+
+            for lib_idx, lib_name in enumerate(libs_to_shades):
+
```
```suggestion
```

vulder (changes requested):
I made a first pass over the code. Looks mostly fine but some of the function/code could be split up a bit more to make it better understandable.

varats/varats/data/reports/blame_report.py
cormensratio: I first thought that mypy falsely gets a tp.List[int] instead of tp.List[tp.Tuple[int,int]], but it seems that my typechecker is the only one complaining ;). I'll remove the cast.

varats/varats/data/reports/blame_report.py
cormensratio: Added: DegreeAmountMappingTy = tp.Dict[int, int]

varats/varats/plots/blame_interaction_degree.py
cormensratio: Added: `BaseAndInterLibToFractionListMappingsTuple = tp.Tuple[tp.Dict[str, tp.List[float]], tp.Dict[str, tp.List[float]]]`

varats/varats/plots/blame_interaction_degree.py
cormensratio: Added: `BaseAndInterLibFractionListsTuple = tp.Tuple[tp.List[tp.List[float]],tp.List[tp.List[float]]]`

varats/varats/plots/blame_interaction_degree.py
cormensratio: Not the same large type, but another.
Added: `LibsToColormapsAndLibsToIndexWithShadesMappingsTuple = tp.Tuple[tp.Dict[str, tp.Any], tp.Dict[str,tp.Dict[int, str]]]`

varats/varats/plots/blame_interaction_degree.py
cormensratio: I wrongly named some inner function names with "_", if you meant that. If we decide to keep the function after the next meeting i will try to split it further.

varats-core/varats/utils/git_util.py
cormensratio: I agree. I will investigate it further.

varats/varats/data/databases/blame_interaction_degree_database.py
```diff
@@ -56,73 +58,162 @@ def _load_dataframe(
...
+                return total
+
+            total_amounts_of_all_libs = calc_total_amounts()
 
             list_of_author_degree_occurrences = generate_author_degree_tuples(
                 report, commit_lookup
             )
-            author_degrees, author_amounts = map(
+
+            author_degrees_untyped, author_amounts_untyped = map(
```
boehmseb: All these map calls will fail if the `list_of_foo` is empty.
To fix this, you can do something like this:

```python
degrees = []
amounts = []
list_of_degree_occurrences = generate_degree_tuples(report)
if list_of_degree_occurrences:
degrees, amounts = map(list, zip(*list_of_degree_occurrences))
total = sum(amounts)
```

boehmseb (changes requested):
There is a problem in the interaction degree database.

varats-core/varats/utils/git_util.py
```diff
@@ -171,6 +171,28 @@ def get_extensions_repr(self, sep: str = ", ") -> str:
 
         return concat_str
 
+    def get_modified_file_extensions(self,
```
vulder: Why did you not directly modify the `get_extensions_repr` method? I think we only need one of the two and can just replace the old one.

varats/varats/data/databases/blame_interaction_degree_database.py
vulder: This also gives one the possibility to define the types of the different lists in advance so we should not need the cast later.

vulder (approved):
For the `unique_revisions` lists, if you don't need the ordering you could use a set here.

The rest overall is now better understandable.

varats/varats/data/databases/blame_interaction_degree_database.py
cormensratio: Refactored it to function: `_get_tuple_of_degree_amount_lists()`

varats-core/varats/utils/git_util.py
cormensratio: I wasn't sure if we need the original functionality of the method in other code parts. I will adapt the method and change the usages to the new one.

cormensratio:
> For the `unique_revisions` lists, if you don't need the ordering you could use a set here.

I think that the set would not work because e.g., the passed data for the stackplot must be hashable. As already mentioned by you the order of the unique revisions is in fact important because the revisions are sorted by their time.
Therefore I would stay with the unique list.

vulder (approved):
lgtm

boehmseb (approved):

PR #414, opened 2020-11-16
Migrate table discovery to match plots

boehmseb:


vulder (commented):
Why did you move table.py and tables.py into different modules in core?
+ Why did you move the full tables module? Core should only have the central elements, the specific implementations should stay in varats.

Maybe we should/can build one table folder in core? Would make it less confusing with regard to table vs tables.

boehmseb:
> Why did you move table.py and tables.py into different modules in core?
> 
>     * Why did you move the full tables module? Core should only have the central elements, the specific implementations should stay in varats.

Well, I refactored the layout such that it matches that of the plot classes.
> 
> 
> Maybe we should/can build one table folder in core? Would make it less confusing with regard to table vs tables.

There is more to it and I suggest, we discuss this in person.

vulder (approved):

vulder (approved):

PR #415, opened 2020-11-25
Latex Document Wrapper for Tables

SimonFedick:
Adds a --wrap-document option to the vara-table command and adds a corresponding wrapper for optionally wrapping latex tables in a document. Wrapper requires the pylatex library.


varats-core/varats/table/table.py
```diff
@@ -160,5 +169,44 @@ def save(
         else:
             table_dir = path
 
-        with open(table_dir / self.table_file_name(), "w") as outfile:
-            outfile.write(table)
+        if wrap_document and self.__format in [
+            TableFormat.latex, TableFormat.latex_raw, TableFormat.latex_booktabs
+        ]:
+            # landscape mode for bug tables because of long entries
+            landscape_mode = "true" if self.__name == "bug_overview" else "false"
```
boehmseb: Please no subclass-specific code in the base class.

varats-core/varats/table/table.py
```diff
@@ -160,5 +169,44 @@ def save(
...
+                default_filepath=f"{table_dir}/{self.table_file_name(False)}",
+                documentclass="scrbook",
+                document_options="paper=a4",
+                geometry_options={
+                    "margin": "1.5cm",
+                    "landscape": landscape_mode
+                }
+            )
+            wrap_table_in_document(table, doc)
+            doc.generate_tex()
```
boehmseb: Move this code to `wrap_table_in_document` and make that function return a `str` (I hope that is easily possible).
That way, we can do:
```python
if wrap_table:
    table = wrap_table_in_document(table)
```

To make things like landscape mode easily configurable, split the wrap function into two functions:
one function that takes the options as parameters, and an abstract function that does not take the parameters and calls the first one with the desired parameters. Subclasses then only need to override the second function with its own parameters.

varats-core/varats/table/table.py
```diff
@@ -139,19 +145,22 @@ def table_file_name(self) -> str:
...
     ) -> None:
         """
         Save the current table to a file.
 
         Args:
             path: The path where the file is stored (excluding the file name).
-            file_name: the file name of the table; this overrides automatic
-                       file name construction.
+            wrap_document: flags whether to wrap the (latex) table code into a
+                whole document.
```
boehmseb:
```suggestion
                complete document.
```

varats/varats/tables/diff_correlation_overview_table.py
```diff
@@ -63,5 +68,18 @@ def save(self, path: tp.Optional[Path] = None) -> None:
...
+            TableFormat.latex_raw, TableFormat.latex_booktabs, TableFormat.latex
+        ]:
+            doc = Document(
+                default_filepath=f"{table_dir}/{self.name}",
+                documentclass="scrbook",
+                document_options="paper=a4",
+                geometry_options={"margin": "1.5cm"}
+            )
+            wrap_table_in_document(table, doc)
+            doc.generate_tex()
```
boehmseb: This should collapse into a simple call to `wrap_table_in_document(table)` once you addressed my other comments.

boehmseb (changes requested):

varats-core/varats/table/table.py
```diff
@@ -110,10 +112,14 @@ def supports_stage_separation() -> bool:
...
         """Build the table using tabulate."""
 
-    def table_file_name(self) -> str:
+    def table_file_name(self, include_filetype: bool = True) -> str:
         """
         Get the file name this table; will be stored to when calling save.
 
+        Args:
+            include_filetype: flags whether to include the file extension at the
+                end of the filename.
```
vulder:
```suggestion
            include_filetype: flags whether to include the file extension at the
                              end of the filename.
```

varats-core/varats/table/table.py
```diff
@@ -139,19 +145,22 @@ def table_file_name(self) -> str:
         ) and self.table_kwargs.get('sep_stages', None):
             sep_stages = 'S'
 
-        return f"{table_ident}{self.name}{sep_stages}.{filetype}"
+        if include_filetype:
+            return f"{table_ident}{self.name}{sep_stages}.{filetype}"
+        return f"{table_ident}{self.name}{sep_stages}"
```
vulder:
```suggestion
        table_file_name = f"{table_ident}{self.name}{sep_stages}"
        if include_filetype:
            table_file_name += f".{filetype}"
        return table_file_name
```

varats-core/varats/table/table.py
vulder: And also align the word `complete` with flags so it forms a block

vulder (changes requested):

varats-core/varats/table/table.py
SimonFedick: This now happens in an abstract method `wrap_table` that calls `wrap_table_in_document` with `self.tabulate` and its desired parameters for each subclass respectively.

varats/setup.py
```diff
@@ -35,6 +35,7 @@
         "pygtrie",
         "pyzmq>=19.0.0",
         "PyGithub>=1.47",
+        "pylatex>=1.4.1"
```
boehmseb:
```suggestion
        "pylatex>=1.4.1",
```

varats-core/varats/table/table.py
```diff
@@ -139,21 +145,38 @@ def table_file_name(self) -> str:
...
             sep_stages = 'S'
 
-        return f"{table_ident}{self.name}{sep_stages}.{filetype}"
+        table_file_name = f"{table_ident}{self.name}{sep_stages}"
+        if include_filetype:
+            table_file_name += f".{filetype}"
+        return table_file_name
+
+    @abc.abstractmethod
+    def wrap_table(self) -> str:
```
boehmseb: Make the table content an argument of this function so that the save function always calls `tabulate()` and the `wrap_table()` function only does the wrapping.

boehmseb (changes requested):

boehmseb:
The rest looks good now btw.

boehmseb (approved):

vulder (approved):
lgtm

vulder:
@SimonFedick Hi, could you please resolve the conflicts, I would merge the PR after wards. :smiley:

PR #416, opened 2020-11-26
Let `vara-art generate` create a HTML overview

boehmseb:
This resolves se-passau/VaRA#644.


varats/varats/tools/html_utils.py
```diff
@@ -0,0 +1,83 @@
+"""Utility module for generating html files."""
+
+import typing as tp
+
+__HTML_BASE_TEMPLATE = """<!DOCTYPE html>
```
vulder: Would it be better to put these "templates" into a separate template file? Similar to how BB renders slurm scripts. With the template engine replacing special values is trivial :smile:

varats/varats/tools/driver_artefacts.py
```diff
@@ -281,11 +273,102 @@ def generate_html_plot_overview(
...
+    title = f"Artefacts for paper config \"{get_paper_config().path.name}\""
+    content = __INDEX_LINKS
+    content += __COLUMN_TEMPLATE.format("\n".join(columns))
+    html = html_page(title, content, [CSS_COMMON, CSS_TABLE])
 
     with open(outfile, "w") as file:
         file.write(html)
 
 
+def _generate_artefact_html(artefact: Artefact,
```
vulder, line 314: Maybe we should add simple tests for the html generation of artefacts/table methods, so we can make sure they stay correct.

vulder (changes requested):
A few things but overall looks good :)

varats/varats/tools/html_utils.py
boehmseb: I don't see the benefit right now, but that might change if we use this in other places in the future.
As I use some of these snippets in different situations and combinations, I decided against this for now. Instead, I started this module.

varats/varats/tools/driver_artefacts.py
boehmseb: Yes, you are right. I will add some tests.

varats/varats/tools/html_utils.py
```diff
@@ -0,0 +1,83 @@
+"""Utility module for generating html files."""
```
boehmseb: @vulder do you think this is the right place for this module? Or should we find a better place?

vulder: Hmm, good point. Normally we put all utils code into core so other people could work with them. Is the html overview part generic enough for that?
If not I see two options.
1) keep it in tools as is
2) create another folder for specific tool suite utils
Not sure what is better here.

varats/varats/tools/html_utils.py
vulder: Reuse is one thing, having better separation and easier insertion are the other two.
+ editing/testing the files in python is a bit cumbersome, e.g., simple viewing the current html in a browser or testing different css foo.

varats/varats/tools/html_utils.py
boehmseb: What is in the module right now *could* be generic enough for core (that's why I left the other more specific templates in the driver module), but I'm not sure whether this is actually interesting for core.

I also thought of your two suggestions. In my opinion, it would fit better inside a `utils` package/folder, but I was not sure if we tried to get rid of this during the split of core, or whether the utils modules all moved to core by chance.

varats/varats/tools/html_utils.py
boehmseb: A disadvantage of having separate files would be that it takes more work to make the templates available to the user (e.g., what templates or styles are available).
And editing is not cumbersome at all if you have an IDE that can interpret strings as any language you want :wink: 

As I said, at least for now, I don't see the benefit. This will of course change if we add more fancy CSS/js, etc., but we can still change that then.

vulder: For now, ok. But we should later fix this.

Regarding the editing, the problem is that I cannot open it with chrome and edit HTML/css directly to see the impact. Dose pyChar render the HTML for you in the IDE from a string?

varats/varats/tools/html_utils.py
vulder: It was by chance because everything was reusable/used in core code.

Then just create a special name `utils` package for varats.

varats/varats/tools/html_utils.py
boehmseb: You can inject almost any language in any string literal in PyCharm and others. That means I can use all HTML-editing capabilities of my IDE inside any String literal I want.

vulder: That part I know, but I want it to be rendered and the rendering on the fly updatable like in chrome with dev tools.

boehmseb: I just checked and it seems like this is not possible. You can edit the string like it is its own file (you even get a new tab for this), but it seems this file is only virtual.

varats/varats/utils/html_utils.py
```diff
@@ -0,0 +1,83 @@
+"""Utility module for generating html files."""
```
vulder: You cannot/should directly name the fold `utils`. This is the same as in `varats-core` namespace package, hence we can get a clash/race between the two inits when importing 'varats.utils.html_utils.py`.
We need to find a unique name for the package folder.

vulder (changes requested):
Except for the folder name, this looks good.

vulder (approved):

PR #418, opened 2020-11-30
Extend VaraTestRepoSources for git submodule usage

cormensratio:
Introduces handling of local libraries used as GitSubmodules in benchbuild projects.
Also closes: se-passau/VaRA#677


varats-core/varats/project/project_util.py
```diff
@@ -254,11 +255,64 @@ def wrap_paths_to_binaries(
     ])
 
 
-# ignore type as we do not have appropriate type information from benchbuild
+def rename_to_git():
```
vulder: Could you please try to replace find with our renaming impl from VaRA. `find`, used like this` has the problem of creating unwanted files.
see se-passau/VaRA#677

varats-core/varats/project/project_util.py
```diff
@@ -57,7 +57,8 @@ def get_local_project_git_path(
     else:
         source = get_primary_project_source(project_name)
 
-    if hasattr(source, "fetch"):
+    if not isinstance(source,
+                      VaraTestSubmoduleSource) and hasattr(source, "fetch"):
         source.fetch()
```
vulder: Why is it wrong to `fetch` here? Checking for a specific class looks a bit strange from a design point of view. Maybe we should adapt the `fetch` method if it does not do the right thing.

varats-core/varats/project/project_util.py
```diff
@@ -254,11 +255,64 @@ def wrap_paths_to_binaries(
...
+
+    def __init__(
+        self,
+        remote: str,
+        local: str,
+        clone: bool = True,
+        limit: int = 10,
+        refspec: str = 'HEAD',
+        shallow: bool = True,
+        version_filter: tp.Callable[[str], bool] = lambda version: True
```
vulder: If you don't change the defaults here, just forward the *args or *kw args that the user gave instead of copy pasting the current state of the super class.

varats-core/varats/project/project_util.py
```diff
@@ -254,11 +255,64 @@ def wrap_paths_to_binaries(
...
+    def __init__(
+        self,
+        remote: str,
+        local: str,
+        clone: bool = True,
+        limit: int = 10,
+        refspec: str = 'HEAD',
+        shallow: bool = True,
+        version_filter: tp.Callable[[str], bool] = lambda version: True,
+        submodules: tp.Optional[tp.List[VaraTestSubmoduleSource]] = None
```
vulder: same

varats-core/varats/project/project_util.py
```diff
@@ -278,48 +332,46 @@ def fetch(self) -> pb.LocalPath:
...
-            )
-            find(
-                ".", "-name", "gitignore", "-execdir", "mv", "-i", "{}",
-                ".gitignore", ";"
-            )
+        def prepare_submodules():
+            for submodule in tp.cast(
+                tp.List[VaraTestSubmoduleSource], self.submodules
+            ):
+                submodule_path = vara_test_repos_path + "/" + submodule.remote
```
vulder: Just make vara_test_repos_path a actual Path and just use `/` to concat paths together

varats/varats/projects/cpp_projects/two_libs_one_project_interaction_discrete_libs_single_project.py
```diff
@@ -69,6 +59,24 @@ class TwoLibsOneProjectInteractionDiscreteLibsSingleProject(
...
+        version_filter=project_filter_generator(
+            "TwoLibsOneProjectInteractionDiscreteLibsSingleProject"
+        ),
+        submodules=submodules
+    )
+
+    SOURCE.append(main_repo)
+    for submodule in submodules:
+        SOURCE.append(submodule)
+
```
vulder: a) why can you not just add everything into the SOURCE list like BB requests?
b) executing code at class parse time for filling class variables looks like bad design to me, why do you need this?

vulder (changes requested):

varats-core/varats/project/project_util.py
```diff
@@ -57,25 +57,25 @@ def get_local_project_git_path(
     else:
         source = get_primary_project_source(project_name)
 
-    if hasattr(source, "fetch"):
+    if not isinstance(source, GitSubmodule) and hasattr(source, "fetch"):
```
boehmseb: Why not fetch submodules here?

Also, use `is_git_source()` here:
```suggestion
    if not isinstance(source, GitSubmodule) and is_git_source(source):
```

varats-core/varats/project/project_util.py
```diff
@@ -275,32 +307,49 @@ def fetch(self) -> pb.LocalPath:
...
+
+                with local.cwd(submodule_target):
+                    rename_to_git()
+
+        extract_main_repo()
+        extract_libraries()
+
+        return main_tgt_path
+
+    def version(self, target_dir: str, version: str = 'HEAD') -> pb.LocalPath:
```
boehmseb, line 378: Why do you need to override this?

varats-core/varats/project/project_util.py
```diff
@@ -275,32 +307,49 @@ def fetch(self) -> pb.LocalPath:
...
+                return
+
+            for submodule in tp.cast(tp.List[GitSubmodule], self.submodules):
+                submodule_path = vara_test_repos_path / Path(submodule.remote)
+                submodule_target = local.path(target_prefix()) / submodule.local
+                mkdir("-p", submodule_target)
+                cp("-r", submodule_path + "/.", submodule_target)
+
+                with local.cwd(submodule_target):
+                    rename_to_git()
```
boehmseb: Doesn't this break the existing Git/Submodule abstraction, since the main git now also takes care to fetch its submodules as separate repos?
I think it would be better to also have a class `VaraTestRepoSubmodule(GitSubmodule)`. It might be that we then need a better way to distinguish Git sources from GitSubmodule sources, but that can be handled by some function similar to `is_git_source()`.

@vulder what's your take on this?

varats/varats/projects/c_projects/gzip.py
```diff
@@ -29,6 +30,7 @@ class Gzip(bb.Project, ReleaseProviderHook, CVEProviderHook):  # type: ignore
     NAME = 'gzip'
     GROUP = 'c_projects'
     DOMAIN = 'compression'
+    CONTAINER = ContainerImage()
```
boehmseb: You should do this in all project classes since they all will break with a BB upgrade.

Which reminds me: we need to upgrade the BB dependency. @vulder maybe we want a new BB release for this?

boehmseb (changes requested):

varats/varats/projects/c_projects/gzip.py
cormensratio: I had not in mind that this PR should add a container for all projects, but i can of course do that.

varats-core/varats/project/project_util.py
cormensratio: At least at one point in the report generation, the original version() call would produce an error similar to this:
```
Stderr:       | fatal: repository 'LibraryAnalysisRepos/TwoLibsOneProjectInteractionDiscreteLibsSingleProject/Elementalist' does not exist
```
As in the overwritten fetch() of VaraTestRepoSource, we have to provide a correct path to the local repository in the version() call, which needs another kind of wrapper for the paths. Therefore I provided the missing path parts in the overwritten version() and removed the insignificant clone flags for a local repository clone.

varats-core/varats/project/project_util.py
boehmseb: But why do we not fetch submodules? I would assume that if we get the git path to a submodule, we want to access it and hence, it should be fetched.

varats/varats/projects/c_projects/gzip.py
boehmseb: We have to because of the implicit BB upgrade.

varats-core/varats/project/project_util.py
cormensratio: Since benchbuild's Git() handles and initializes its submodules and the actual GitSubmodule() only provides an entry point in code to distinguish between main repository and submodule, I ensured that the corresponding wrapper (VaraTestRepoSource) does the same with the local repositories.

varats-core/varats/project/project_util.py
cormensratio: I will try to replace it.

varats-core/varats/project/project_util.py
boehmseb: Ah, I see. Seems like git still checks whether the remote repository exists despite the `--dissociate` and `--reference`.

varats-core/varats/project/project_util.py
boehmseb: Yes, you are right. Somehow I overlooked that obvious problem.
But still, I think we should have this extra class for submodules as I said above. You can then simply call `fetch()` on each submodule and copy it afterwards. Depending on the repo, it could even be that `git submodule --init` will work afterwards.

varats-core/varats/project/project_util.py
cormensratio: Your question is related to one question of @vulder regarding the SOURCES list. I'll test if your proposed idea can solve both problems and let you know.

varats/varats/projects/c_projects/gzip.py
simbuerg: I think that should be a bug-report for benchbuild. An empty container declaration should not be necessary.

varats-core/varats/project/project_util.py
cormensratio: The main problem is that we need the initialized submodules already when the main repo gets initialized. This is also the reason why I built the submodule list in the SOURCES before the VaraTestRepoSource. Considering that, I saw no reason to fetch() the submodules, at the referred code part, again.
I could not figure out how i can circumvent the non-optimal definition of SOURCES and the execution of code at class parse time, which @vulder already mentioned. But your proposed wrapper (VaraTestRepoSubmodule) for the GitSubmodules enables us to call fetch() on a submodule without an path-error, therefore i can remove the ```if not isinstance(source, GitSubmodule)``` part.

varats-core/varats/project/project_util.py
cormensratio: As mentioned above, I added a wrapper for the GitSubmodules. Currently I replaced this code part with a loop over the submodules and call fetch on each one.

varats/varats/projects/cpp_projects/two_libs_one_project_interaction_discrete_libs_single_project.py
cormensratio: As I wrote in a lower comment to @boehmseb , we need the initialized submodules already when the main repo gets initialized. Currently I have found no other way to provide the main repo with the extracted submodules, therefore I initialize them at the same time, when we fetch and initialize the main repo. To do so, the main repo has to know which submodules it has to initialize, so i pass a list of its submodules. 
This would also mirror the behaviour of benchbuilds Git(), as it initializes/manages its submodules.
But i totally agree that this probably is a bad design. If you have a recommendation to improve this, let me know.

varats-core/varats/project/project_util.py
cormensratio: Done

varats-core/varats/project/project_util.py
cormensratio: Done

varats-core/varats/project/project_util.py
cormensratio: I now have adapted the fetch of the VaraTestRepoSubmodule. See lower conversations with boehmseb.

varats-core/varats/project/project_util.py
boehmseb: Those are actually two different problems. 
For our test repo sources, we probably will not be able to get rid of this forward-reference.
This comment here and the wrapper around `GitSubmodule` are needed to preserve the submodule-abstraction built into benchbuild.

varats/varats/projects/cpp_projects/two_libs_one_project_interaction_discrete_libs_single_project.py
boehmseb: ~~I agree with Simon here. I fear we will not be able to get rid of this forward-reference for our test repos.~~
Actually, I just realized that we have a larger problem here. I think it would be best to discuss that in our next meeting.

vulder: I understand the problem but we have to find another way to specify this. In the current state this requires to much knowledge between the different Git/Submodules becoming brittle if either the project or something else changes.

In general, we need to find a way so that our normal VaraTestGit can do a submodule initialization without the VaraTestGitSubmodules. All needed information is already present in the .gitted/.gitsubmouldes files we just need to find a way to map this back to our TestRepo structure, so we can initialize the submodules -- as git would to with submodule init -- without a relation to the VaraTestGitSubmodule class. But I'm not sure how easy this is to implement.

Discussing this in detail in the next meeting sound good to me.

cormensratio: :+1:

varats-core/varats/project/project_util.py
cormensratio: Done

varats-core/varats/project/project_util.py
```diff
@@ -254,16 +255,82 @@ def wrap_paths_to_binaries(
...
+    __vara_test_repos_git = Git(
+        remote="https://github.com/se-passau/vara-test-repos",
+        local="vara_test_repos",
+    )
+
+    def fetch(self) -> pb.LocalPath:
+        """
+        Overrides ``GitSubmodule`` s fetch to
+          1. fetch the vara-test-repos repo
+          2. extract the specified submodule from the vara-test-repos repo
```
vulder, line 322: Should also mention that we do the .gitted -> .git rewriting

varats/varats/projects/c_projects/gzip.py
vulder: @cormensratio If you already filed the bug report, just add a comment here:
```
TODO (PolyJIT/benchbuild#42): remove after BB provides implicit default
```

varats/varats/projects/cpp_projects/two_libs_one_project_interaction_discrete_libs_single_project.py
```diff
@@ -28,6 +29,8 @@ class TwoLibsOneProjectInteractionDiscreteLibsSingleProject(
     NAME = 'TwoLibsOneProjectInteractionDiscreteLibsSingleProject'
     GROUP = 'cpp_projects'
     DOMAIN = 'library-testproject'
+    CONTAINER = ContainerImage()
```
vulder: same

vulder (approved):
Looks good, found just a few nits.
Can you afterwards in a new PR also clean up VaRATestRepo source with the new `copy_renamed...` method and checkout if the issue is fixed? thanks :smile: 

@boehmseb I think we need a BB version bump for this, do we already have a BB release that includes this or are we waiting for the container fix?

boehmseb:
> @boehmseb I think we need a BB version bump for this, do we already have a BB release that includes this or are we waiting for the container fix?

The following issues on benchbuild side need to be resolved first: PolyJIT/benchbuild#345, PolyJIT/benchbuild#346, and PolyJIT/benchbuild#347. The last one asks for a new release.

varats-core/varats/project/project_util.py
```diff
@@ -254,16 +255,82 @@ def wrap_paths_to_binaries(
...
+
 class VaraTestRepoSource(Git):  # type: ignore
     """A project source for repositories stored in the vara-test-repos
     repository."""
 
     __vara_test_repos_git = Git(
         remote="https://github.com/se-passau/vara-test-repos",
         local="vara_test_repos",
-        refspec="HEAD",
-        limit=1
```
boehmseb: Is there any specific reason why you removed this?
Yes, in practice, we do not use those variables, since we do not call `versions()` on this repo. But technically, we only want the first revision of HEAD of this repo.

varats-core/varats/project/project_util.py
```diff
@@ -254,16 +255,82 @@ def wrap_paths_to_binaries(
...
+                os.rename(
+                    os.path.join(root, name),
+                    os.path.join(root, ".gitattributes")
+                )
+            elif name == "gitignore":
+                os.rename(
+                    os.path.join(root, name), os.path.join(root, ".gitignore")
+                )
+            elif name == ".gitted":
+                os.rename(os.path.join(root, name), os.path.join(root, ".git"))
```
boehmseb, line 299: Can this case happen? Isn't `.git` (and hence, `.gitted`) always a dir?

boehmseb (commented):
Two quick questions. Otherwise looks good.

varats-core/varats/project/project_util.py
```diff
@@ -275,32 +342,56 @@ def fetch(self) -> pb.LocalPath:
...
+        cp("-r", main_repo_src_local + "/.", tgt_loc)
+
+        # Extract submodules
+        with pb.local.cwd(tgt_loc):
+
+            # Get submodule entries
+            submodule_url_entry_list = git(
+                "config", "--file", ".gitmodules", "--name-only",
+                "--get-regexp", "url"
+            ).split('\n')
```
boehmseb, line 405: This fails for repos without a `.gitmodules`. You should check first if this file exists and skip this whole step if not.

Something like this before the `cwd` should do:
```python
if not Path(tgt_loc / ".gitmodules").exists():
    return tgt_loc
```

boehmseb (changes requested):

varats-core/varats/project/project_util.py
cormensratio: Added 
```3. rename files that were made git_storable (e.g., .gitted) back to their original name (e.g., .git)```

File: varats-core/varats/project/project_util.py (hunk 13 above)
Comment (cormensratio) [reply varats-core/varats/project/project_util.py]: The default value for "refspec" in BBs Git is already "HEAD", but I will add it again to make it clear. The deletion of limit was unintentional.

File: varats-core/varats/project/project_util.py (hunk 14 above)
Comment (cormensratio) [reply varats-core/varats/project/project_util.py:299]: It seems that git can create .git files. 
See a prior discussion: https://github.com/se-passau/VaRA/pull/668#discussion_r486217970

File: varats-core/varats/project/project_util.py (hunk 15 above)
Comment (cormensratio) [reply varats-core/varats/project/project_util.py:405]: You are right. Always talking about submodules makes you forget that not every repo has them ;).

File: varats/varats/projects/cpp_projects/two_libs_one_project_interaction_discrete_libs_single_project.py (hunk 12 above)
Comment (cormensratio) [reply varats/varats/projects/cpp_projects/two_libs_one_project_interaction_discrete_libs_single_project.py]: Done

File: varats-core/varats/project/project_util.py (hunk 14 above)
Comment (boehmseb) [reply varats-core/varats/project/project_util.py:299]: Ok

File: varats/varats/projects/test_projects/bug_provider_test_repos.py [hunk 16]
```diff
@@ -26,7 +26,7 @@ class BasicBugDetectionTestRepo(bb.Project):  # type: ignore
             remote="BugDetectionRepos/BasicTestRepo",
             local="basic_test_repo",
             refspec="HEAD",
-            limit=None,
+            limit=1,
```
boehmseb: For the projects, the limit needs to stay `None`, otherwise only the newest revision gets cloned. We want this only for the `vara-test-repos` repo.

Easiest fix is to revert 52fdb07cf5664dd70fb8a1447ac862b18092766e

varats-core/varats/project/project_util.py
```diff
@@ -257,7 +258,79 @@ def wrap_paths_to_binaries(
...
+    original git name and stores the renamed copy at the destination path. The
+    original files stay untouched. Renaming and copying will be skipped if the
+    dest_dir already exists.
+
+    Args:
+        src_dir: path to the source directory
+        dest_dir: path to the destination directory
+    """
+    if os.path.isdir(dest_dir):
+        return
```
vulder: This looks like we fail silently here, maybe we should write at least a error message to the log. Providing a wrong `dest_dir` seems like a clear error/misshap to me.

varats-core/varats/project/project_util.py
```diff
@@ -274,36 +347,66 @@ def fetch(self) -> pb.LocalPath:
...
+
+        main_repo_src_local = self.fetch()
+        tgt_loc = pb.local.path(target_dir) / self.local
+        vara_test_repos_path = self.__vara_test_repos_git.fetch()
+        main_repo_src_remote = vara_test_repos_path / self.remote
+
+        mkdir('-p', tgt_loc)
+
+        # Extract main repository
+        cp("-r", main_repo_src_local + "/.", tgt_loc)
```
vulder, line 390: Is the `"/."` needed here? I would assume that `cp` defaults to the dir.

varats/varats/projects/c_projects/gzip.py
```diff
@@ -30,6 +31,9 @@ class Gzip(bb.Project, ReleaseProviderHook, CVEProviderHook):  # type: ignore
     GROUP = 'c_projects'
     DOMAIN = 'compression'
 
+    # TODO (PolyJIT/benchbuild#345): remove after BB provides implicit default
+    CONTAINER = ContainerImage()
```
vulder: Can be fixed with the next BB release.

vulder (changes requested):
@boehmseb Do we have a smoke test for our own GitWrappers?
If no, maybe we should add a call that would trigger an error if something breaks.

Overall this seems fine to me :smile: 
After cleaning up with the BB release and small fixes we can merge this.

varats-core/varats/project/project_util.py
cormensratio: You are right. ~~I will add a log error message.~~ I added a log error message.

varats-core/varats/project/project_util.py
cormensratio: It is needed. Your assumption is correct, but since the target directory got already created in line 377, we only want to copy the contents to avoid nesting e.g., the Elementalist folder twice.

varats/varats/projects/c_projects/gzip.py
cormensratio: I added the new bb release to the requirements.txt, setup.py and removed the todos.

boehmseb (approved):
lgtm

requirements.txt
```diff
@@ -1,5 +1,5 @@
 argparse-utils>=1.2.0
-benchbuild>=5.2
+benchbuild>=6.0.1
```
cormensratio, line 2: PR https://github.com/se-passau/VaRA/issues/488, which got merged in vara-dev, uses the benchbuild version `"benchbuild>=5.2,<6.0.0"`. I do not know if this version range is required for the PR to work, but a bb version >= 6.0.1 is required for this PR.

varats-core/varats/project/project_util.py
```diff
@@ -274,36 +357,67 @@ def fetch(self) -> pb.LocalPath:
...
+
+        # Skip submodule extraction if none exist
+        if not Path(tgt_loc / ".gitmodules").exists():
+            return tgt_loc
```
boehmseb: You still have to do the checkout, even if no submodules are present.

```suggestion
        if not Path(tgt_loc / ".gitmodules").exists():
            with pb.local.cwd(tgt_loc):
                git("checkout", "--detach", version)
            return tgt_loc
```

requirements.txt
boehmseb: That was to pin the version to pre-6.0 since no migration to the new BB version was done there. But we have done the migration here, so 6.0.1 is fine.

boehmseb (approved):
Found one small bug (maybe we should add a test for that). Otherwise looks good

varats-core/varats/project/project_util.py
```diff
@@ -257,7 +261,86 @@ def wrap_paths_to_binaries(
...
+            elif name == ".gitted":
+                os.rename(os.path.join(root, name), os.path.join(root, ".git"))
+
+        for name in dirs:
+            if name == ".gitted":
+                os.rename(os.path.join(root, name), os.path.join(root, ".git"))
+
+
+# TODO (se-passau/VaRA#717): Remove pylint's disable when issue is fixed
+# pylint: disable=too-many-ancestors
```
vulder: This disable is wrong, this would disable all too-many-ancestors errors in the whole file

vulder (approved):

varats-core/varats/project/project_util.py
cormensratio: I changed it to target the corresponding line only.

PR #419, opened 2020-12-02
Revert mypy hack

boehmseb:
Mypy released a new version so we can go back to using the normal releases.
Also, we no longer have to ignore rich during type checks.


vulder:
Can/should we adapt this for the new CI pipeline?

boehmseb:
> Can/should we adapt this for the new CI pipeline?

If you haven't already tested whether the new CI pipeline works without the "uninstall rich" hack, we can test it here. If you have already tested this and it does not work, then you can close this PR.

vulder:
> > Can/should we adapt this for the new CI pipeline?
> 
> If you haven't already tested whether the new CI pipeline works without the "uninstall rich" hack, we can test it here. If you have already tested this and it does not work, then you can close this PR.

Did not work for me locally without rich uninstall. Then I close this for now, maybe a future update will fix this. But at least we can fall back to the release mypy version :wink:

PR #420, opened 2020-12-07
Result folder cleanup utility resolves se-passau/VaRA#488

Sinerum:
Implemented the cleanup utility for old files and user specified regex as described in  se-passau/VaRA#488


varats/varats/tools/driver_casestudy.py
```diff
@@ -561,8 +568,32 @@ def find_result_dir_paths_of_projects() -> tp.List[Path]:
...
+            newer_files: tp.Dict[str, Path] = dict()
+            result_dir_cs = result_dir/case_study.project_name
+            if not result_dir_cs.exists():
+                continue
+            for opt_res_file in result_dir_cs.iterdir():
+                commit_hash = MetaReport.get_commit_hash_from_result_file(opt_res_file.name)
+                if case_study.has_revision(commit_hash):
+                    current_file = newer_files.get(commit_hash)
+                    if current_file is None:
+                        newer_files.__setitem__(commit_hash,opt_res_file)
```
boehmseb:
```suggestion
                        newer_files[commit_hash] = opt_res_file
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -561,8 +568,32 @@ def find_result_dir_paths_of_projects() -> tp.List[Path]:
...
+                    else:
+                        if (
+                                current_file.stat().st_mtime <
+                                opt_res_file.stat().st_mtime
+                        ):
+                            nnewer_files.__setitem__(commit_hash,opt_res_file)
```
boehmseb: Typo?

```suggestion
                            newer_files[commit_hash] = opt_res_file
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -561,8 +568,32 @@ def find_result_dir_paths_of_projects() -> tp.List[Path]:
         return existing_paper_config_result_dir_paths
 
     def remove_old_result_files() -> None:
-        pass
-        # TODO: Implement the removal of old result files.
+        paper_config = get_paper_config()
+        result_dir = Path(str(vara_cfg()['result_dir']))
+        for case_study in paper_config.get_all_case_studies():
+            old_files = []
```
boehmseb: add type annotation

varats/varats/tools/driver_casestudy.py
```diff
@@ -561,8 +568,32 @@ def find_result_dir_paths_of_projects() -> tp.List[Path]:
...
+                        if (
+                                current_file.stat().st_mtime <
+                                opt_res_file.stat().st_mtime
+                        ):
+                            nnewer_files.__setitem__(commit_hash,opt_res_file)
+                            old_files.append(current_file)
+                        else:
+                            old_files.append(opt_res_file)
+            for file in old_files:
+                if os.path.exists(file):
```
boehmseb: As you already have a `Path` here, prefer the pathlib API over `os`:
```suggestion
                if file.exists():
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
-        # TODO: Implement the removal of files specified by the user
-        #  https://github.com/se-passau/VaRA/issues/488
+        filter = args["filter"]
+        if filter == "":
+            parser.error("Specify a regex filter with --filter or -f")
+        result_dir_paths = find_result_dir_paths_of_projects()
+
+        for result_dir_path in result_dir_paths:
+            result_file_names = os.listdir(result_dir_path)
+            todelete = []
```
boehmseb: Add type annotation.

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
                     #  deleted files.
 
     def remove_result_files_by_regex() -> None:
-        pass
-        # TODO: Implement the removal of files specified by the user
-        #  https://github.com/se-passau/VaRA/issues/488
+        filter = args["filter"]
+        if filter == "":
```
boehmseb: Empty strings are `False` in python.

```suggestion
        if not filter:
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
+            for result_file_name in result_file_names:
+               match =  re.match(filter,result_file_name)
+               if match is not None:
+                   todelete.append(result_file_name);
+            if not todelete:
+                print(f"No matching result files in {result_dir_path} found.")
+                continue
+
+            for file_name in todelete:
+                print(f"{file_name}")
```
boehmseb: I'm not sure if we want to list all files here. What do you think @vulder?

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
+            print(f"Found {len(todelete)} matching result files in {result_dir_path}:")
+            try:
+                if cli_yn_choice("Do you whant to delte these files","n"):
```
boehmseb:
```suggestion
                if cli_yn_choice("Do you whant to delete these files","n"):
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
+                    for file_name in todelete:
+                        if os.path.exists(result_dir_path / file_name):
```
boehmseb: Use `Path.exists()`.

boehmseb (changes requested):
As this is your first PR, there is a bit more to say about our development guidelines:

Typing:
---------

We use mypy to statically typecheck our code. This can be donw from the commandline via `mypy --strict -p varats`. If you encounter lots of errors from benchbuild modules, just add a ~~` | grep -v "error: unused 'type: ignore' comment"`~~ (no longer needed). Please fix all type errors that are related to your code.

Linting:
----------

We use pylint for static code analysis. you should be able to run it with `pylint varats`. Again, please fix all errors related to your code if possible, or silence them otherwise.

Formatting:
---------------

We have a mechanism to install and manage git hooks that take care of formatting, import ordering and such. Simply run `pre-commit install` in your environment to activate those git hooks. You can run the hooks manually via `pre-commit run --all-files`.

Usually our CI would also enforce that a PR adheres to these requirements, but travis has ceased to support open source projects apparently and does not work currently.


As for the code itself:
----------------------------

I added some comments to address typos and some smaller mistakes.
Also, please make the inner functions of `__casestudy_cleanup()` top-level functions as they do not really depend on their outer function. You can make them private (one leading `_` indtead of two) so they can still be tested easily.

varats/varats/tools/driver_casestudy.py
```diff
@@ -543,7 +550,7 @@ def open_in_editor(result_file: Path) -> None:
         return
 
 
-def __casestudy_cleanup(args: tp.Dict[str, tp.Any]) -> None:
+def __casestudy_cleanup(args: tp.Dict[str, tp.Any],parser: ArgumentParser) -> None:
```
vulder: Formatting, should be automatically checked/fixed in the future by the precommit checker. To fix this just use yapf directly or integrate yapf into your IDE. Assuming PyCharm @boehmseb can help you on how to set it up.

varats/varats/tools/driver_casestudy.py
vulder: `cli_yn_choice` is not defined in this context, I guess there is a import missing.

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
+
+            for file_name in todelete:
+                print(f"{file_name}")
+            print(f"Found {len(todelete)} matching result files in {result_dir_path}:")
+            try:
+                if cli_yn_choice("Do you whant to delte these files","n"):
+                    for file_name in todelete:
+                        if os.path.exists(result_dir_path / file_name):
+                            os.remove(result_dir_path/file_name)
+            except:
```
vulder: This is a cache all without a specific exception type, which exception did you want to catch here?

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
+                EOFError
```
vulder: This statement has no effect, did you want to cache EOFErrors here?

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
+                continue
```
vulder: When does this happen? Just catching and skipping the exception without any logging/user message seems a bit strange to me.

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
                     #  deleted files.
 
     def remove_result_files_by_regex() -> None:
-        pass
-        # TODO: Implement the removal of files specified by the user
-        #  https://github.com/se-passau/VaRA/issues/488
+        filter = args["filter"]
```
vulder: `filter` is a python build in function, maybe rename this var to `filter_regex`?

varats/varats/tools/driver_casestudy.py
vulder:
```suggestion
        if not filter:
```

varats/varats/tools/driver_casestudy.py
vulder:
```suggestion
            files_to_delete: tp.List[?] = []
```
Maybe `files_to_delete` is a clearer name.
+ For empty list constructions we should add a type annotation to specify the type, so miss use later would be caught.

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
+            parser.error("Specify a regex filter with --filter or -f")
+        result_dir_paths = find_result_dir_paths_of_projects()
+
+        for result_dir_path in result_dir_paths:
+            result_file_names = os.listdir(result_dir_path)
+            todelete = []
+            for result_file_name in result_file_names:
+               match =  re.match(filter,result_file_name)
+               if match is not None:
+                   todelete.append(result_file_name);
```
vulder:
```suggestion
                match =  re.match(filter,result_file_name)
                if match is not None:
                    todelete.append(result_file_name);
```
Wrong indentation depth

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
+            if not todelete:
+                print(f"No matching result files in {result_dir_path} found.")
+                continue
+
+            for file_name in todelete:
+                print(f"{file_name}")
+            print(f"Found {len(todelete)} matching result files in {result_dir_path}:")
+            try:
```
vulder:
```suggestion
                print(f"{file_name}")
            print(f"Found {len(todelete)} matching result files in {result_dir_path}:")
            
            try:
```
I would add a space here, makes it a bit easier to spot where the user check happens.

varats/varats/tools/driver_casestudy.py
```diff
@@ -561,8 +568,32 @@ def find_result_dir_paths_of_projects() -> tp.List[Path]:
...
+                            nnewer_files.__setitem__(commit_hash,opt_res_file)
+                            old_files.append(current_file)
+                        else:
+                            old_files.append(opt_res_file)
+            for file in old_files:
+                if os.path.exists(file):
+                    os.remove(file)
 
     def remove_error_result_files() -> None:
         result_dir_paths = find_result_dir_paths_of_projects()
```
vulder: Is method was started by cormensratio but we should fix it an clean it up in this PR, so it works correctly.

a) `__contains__` should not me called directly
b) the contains check is not necessary use:
```
MetaReport.result_file_has_status_*
```
For error files we should check `error` and `cerror` here.

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
                     #  deleted files.
 
     def remove_result_files_by_regex() -> None:
```
vulder: @boehmseb What do you think, should be also restrict regex deletion (by default) to the current ~~case study~~ paper config?

varats/varats/tools/driver_casestudy.py
```diff
@@ -561,8 +568,32 @@ def find_result_dir_paths_of_projects() -> tp.List[Path]:
...
+            for opt_res_file in result_dir_cs.iterdir():
+                commit_hash = MetaReport.get_commit_hash_from_result_file(opt_res_file.name)
+                if case_study.has_revision(commit_hash):
+                    current_file = newer_files.get(commit_hash)
+                    if current_file is None:
+                        newer_files.__setitem__(commit_hash,opt_res_file)
+                    else:
+                        if (
+                                current_file.stat().st_mtime <
+                                opt_res_file.stat().st_mtime
```
vulder: If I remember correctly, we need to use `st_mtime_ns` here, with `st_time` we run into the strange floating point time comparison bug.
@boehmseb

varats/varats/tools/driver_casestudy.py
vulder: `nnewer_files` looks like a typo, because it's not defined in this scope.
Furthermore, to set a item in a map just use the `[]`operator instead of calling `__setitem__` cirectly.
```
foo_dict = {}
foo_dict["key"] = bar_value
```

varats/varats/tools/driver_casestudy.py
vulder: Should have a type annotation

varats/varats/tools/driver_casestudy.py
```diff
@@ -561,8 +568,32 @@ def find_result_dir_paths_of_projects() -> tp.List[Path]:
...
-        pass
-        # TODO: Implement the removal of old result files.
+        paper_config = get_paper_config()
+        result_dir = Path(str(vara_cfg()['result_dir']))
+        for case_study in paper_config.get_all_case_studies():
+            old_files = []
+            newer_files: tp.Dict[str, Path] = dict()
+            result_dir_cs = result_dir/case_study.project_name
+            if not result_dir_cs.exists():
+                continue
```
vulder:
```suggestion
                continue

```

varats/varats/tools/driver_casestudy.py
vulder: operator `[]`

varats/varats/tools/driver_casestudy.py
```diff
@@ -561,8 +568,32 @@ def find_result_dir_paths_of_projects() -> tp.List[Path]:
         return existing_paper_config_result_dir_paths
 
     def remove_old_result_files() -> None:
```
vulder: @boehmseb Maybe we should move these cleaning functions into a separate module, any suggestions where they would have a nice place to live and feel welcome?
Ideas:
* paper_mgmt
* report.report
* new one

varats/varats/tools/driver_casestudy.py
```diff
@@ -304,6 +304,13 @@ def __create_cleanup_parser(sub_parsers: _SubParsersAction) -> None:
         choices=["old", "error", "regex"],
         type=str
     )
+    cleanup_parser.add_argument(
+        "--filter",
+        help="Filter when unsing regex",
```
vulder:
```suggestion
        help="Regex to determine which report files should be deleted.",
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -304,6 +304,13 @@ def __create_cleanup_parser(sub_parsers: _SubParsersAction) -> None:
         choices=["old", "error", "regex"],
         type=str
     )
+    cleanup_parser.add_argument(
+        "--filter",
```
vulder: In `vara-cs status` we called it `--filter-regex`, using the same here would make the option consistent.
```suggestion
        "--filter-regex",
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -62,7 +62,7 @@ def main() -> None:
     elif args['subcommand'] == 'view':
         __casestudy_view(args)
     elif args['subcommand'] == 'cleanup':
-        __casestudy_cleanup(args)
+        __casestudy_cleanup(args,parser)
```
vulder:
```suggestion
        __casestudy_cleanup(args, parser)
```
Just formatting thins but running yapf once should automatically fix everything :smile:

varats/varats/tools/driver_casestudy.py
```diff
@@ -579,13 +610,42 @@ def remove_error_result_files() -> None:
...
+                continue
+
 
     # TODO: Implement case distinction for cleanup types
     if cleanup_type == "error":
         remove_error_result_files()
+    if cleanup_type == "old":
+        remove_old_result_files()
+    if cleanup_type == "regex":
+        remove_result_files_by_regex()
```
vulder: Maybe we should use an enum here to mark the different cases.

vulder (changes requested):
Overall functionality looks good and matches what I had in mind when writing the issue :+1: .
I remarked a few things that we should clean up/fix before merging.

A large part of the remarks would be automatically solved in the future when we fix your setup. If you have question just write me or @boehmseb.

vulder:
Oh boy :laughing: some of my comments overlap with @boehmseb , somehow github did not refresh for me and show me your comments^^

varats/varats/tools/driver_casestudy.py
Sinerum: EOFError occurs when the input to the confirmation is terminated. I would treat that as a no in this case and move or would you expect different behavior?

varats/varats/tools/driver_casestudy.py
Sinerum: i could not find error in the FileStatusExtension enum so i only check CompileError

varats/varats/tools/driver_casestudy.py
```diff
@@ -543,49 +554,109 @@ def open_in_editor(result_file: Path) -> None:
         return
 
 
-def __casestudy_cleanup(args: tp.Dict[str, tp.Any]) -> None:
+def __casestudy_cleanup(
+    args: tp.Dict[str, tp.Any], parser: ArgumentParser
+) -> None:
     cleanup_type = args['cleanup_type']
+
+    if cleanup_type == "error":
```
vulder: The enum is still missing

vulder (approved):
The enum fix is missing,otherwise, this looks good to me. Thanks for implementing this :smile:

varats/varats/tools/driver_casestudy.py
```diff
@@ -543,49 +554,109 @@ def open_in_editor(result_file: Path) -> None:
...
+    for result_dir_path in result_dir_paths:
+        result_file_names = os.listdir(result_dir_path)
+
+        for result_file_name in result_file_names:
+            if MetaReport.result_file_has_status(
+                result_file_name, FileStatusExtension.CompileError
+            ) and MetaReport.is_result_file(result_file_name):
+                os.remove(result_dir_path / result_file_name)
+                # TODO: Maybe implement listing of to be deleted or
+                #  deleted files.
```
boehmseb: I would say we don't need this, so the TODO can be removed

boehmseb (approved):

varats/varats/tools/driver_casestudy.py
```diff
@@ -621,12 +620,14 @@ def _remove_error_result_files() -> None:
         result_file_names = os.listdir(result_dir_path)
 
         for result_file_name in result_file_names:
-            if MetaReport.result_file_has_status(
-                result_file_name, FileStatusExtension.CompileError
+            if (
+                MetaReport.result_file_has_status(
```
vulder: Just a quick tipp here, you don't have to directly pass the FSE you want to check, there are helper methods that make calling code easier to read/write.
Just use:
```
MetaReport.result_file_has_status_failed(result_file_name)
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -621,12 +620,14 @@ def _remove_error_result_files() -> None:
...
+                    result_file_name, FileStatusExtension.CompileError
+                ) or MetaReport.result_file_has_status(
+                    result_file_name, FileStatusExtension.Failed
+                )
             ) and MetaReport.is_result_file(result_file_name):
```
vulder: Small side note: If you want to check if `result_file_name` is actually a result file, i.e., conforms to the filename formatting, you should do it before checking the status because checking the status on a wrong formatted file will most likely not work. This would in the end not cause an error because the status checks would return `False` and everything is OK. However, with check result file first it is "logically" the correct ordering and makes the checks in order a bit easier to understand.

vulder (approved):
I found 2 small nits, please clean them up before I merge it but there is no further review required afterwards :).
The tool looks good :+1:  thanks for implementing and fixing the issue.

vulder (approved):

PR #423, opened 2020-12-26
Enable new CI pipeline with github actions

vulder:


.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,41 @@
+name: reviewdog
+on: [pull_request]
```
boehmseb: Is there a difference between this syntax and the one used in `pre-commit.yml`?

boehmseb (approved):
Looks good to me. I'm just confused why the _pylint_ and _mypy_ reports are listed under _Python CI_ when they are actually executed by _reviewdog_. Why is this?

.github/workflows/reviewdog.yml
vulder: somehow they are always shown under the first action tab, I guess reviewdog posting action results somehow is responsible for this. It's the same for benchbuilds gh actions.

vulder: I tried different setup/naming schemas but the runs get always attributed to the first action... no idea why

PR #424, opened 2020-12-26
Fixes some mypy/pylint issues

vulder:


boehmseb (approved):
Finally I can drop the `grep -v "Unused type ignore comment"` when running mypy locally ;)

vulder:
> Finally I can drop the `grep -v "Unused type ignore comment"` when running mypy locally ;)

Yes, me too.
+ currently, it looks like it's just running/reporting results as if executed locally.

However, there is still a issue with ReviewDog that it shows the findings X times instead of once. Which seems to be a bug with review dog as far as I can tell because the underlying tool pylint only reports the error/warning once. :rofl:

PR #427, opened 2021-01-07
Add switch to select output format of vara-table

Sinerum:
Add an argument "output-format" to the vara-table tool to specify the format of the table. Closes se-passau/VaRA#699

### DIFF
--- a/varats/varats/tools/driver_table.py
+++ b/varats/varats/tools/driver_table.py
@@ -4,8 +4,11 @@ import argparse
 import logging
 import typing as tp
 
+from argparse_utils import enum_action
+
 from varats.data.discover_reports import initialize_reports
 from varats.projects.discover_projects import initialize_projects
+from varats.table.table import TableFormat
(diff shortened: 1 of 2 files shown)


varats/varats/tools/driver_table.py
```diff
@@ -58,6 +61,12 @@ def main() -> None:
         "Tables may ignore this option.",
         default="EmptyReport"
     )
+    parser.add_argument(
+        "--output-format",
+        help="The format the table should have",
+        action=enum_action(TableFormat),
+        default=TableFormat.fancy_grid
```
Sinerum: should fancy_grid be the default or latex_booktabs?
latex_booktabs is used as the default in the constructor for the class Table which is only used in tests,
fancy_grid was hard coded in build_table before the change

vulder (approved):

vulder (approved):

boehmseb (approved):

PR #431, opened 2021-01-25
Adds a check for VaRA major releases

cormensratio:
Adds a check for VaRA major releases when using `vara-buildsetup vara -u` or `vara-buildsetup vara -b`.
closes https://github.com/se-passau/VaRA/issues/565


varats/varats/tools/research_tools/vara.py
```diff
@@ -29,9 +31,22 @@
 LOG = logging.getLogger(__name__)
 
 
+class UnixColors(Enum):
```
vulder: Please don't build your own solution here, plumbum already supports colored console output.

varats/varats/tools/research_tools/research_tool.py
```diff
@@ -329,6 +338,10 @@ def setup(self, source_folder: tp.Optional[Path], **kwargs: tp.Any) -> None:
             source_folder: location to store the code base in
         """
 
+    @abc.abstractmethod
+    def is_up_to_date(self) -> None:
+        """Returns true if a newer major release of VaRA is available."""
```
vulder: This method should probably return a `bool` an not `None`.
Furthermore, if the method is called `is_up_to_data` returning `true` to indicate that there is a newer version is a bit confusing :wink:

varats/varats/tools/research_tools/vara.py
```diff
@@ -170,6 +207,77 @@ def setup(self, source_folder: tp.Optional[Path], **kwargs: tp.Any) -> None:
...
+                f"{UnixColors.OKGREEN.value}VaRA major release version is up "
+                f"to date!{UnixColors.ENDC.value}"
+            )
+        else:
+            print(
+                f"{UnixColors.WARNING.value}VaRA is outdated!\n"
+                f"Upgrade to major release version "
+                f"{UnixColors.BOLD.value}{UnixColors.OKBLUE.value}"
+                f"{highest_vara_llvm_version}{UnixColors.ENDC.value}"
+            )
```
vulder: This method should only to the checking an return `true`/`false` when a miss match is detected. User output/handling what happens when this is not the case should be done in the driver or another appropriate location.

varats/varats/tools/research_tools/vara.py
```diff
@@ -170,6 +207,77 @@ def setup(self, source_folder: tp.Optional[Path], **kwargs: tp.Any) -> None:
...
+                    if match_version > highest_version:
+                        highest_version = match_version
+
+            if highest_version == -1:
+                LOG.warning(
+                    "No vara-llvm-project release branch name was "
+                    "found."
+                )
+
+            return highest_version
```
vulder: These two methods look very similar, I think we could merge this into on generic method.

vulder (changes requested):
Could you add a few tests maybe for smaller methods that we at least can check the regex matches etc.?

Otherwise, in general your approach should work :+1:

varats/varats/tools/driver_build_setup.py
```diff
@@ -58,6 +59,27 @@ def update_term(text: str, enable_inline: bool = False) -> None:
...
+
+    Args:
+        research_tool: The loaded research tool
+    """
+    highest_release_version = research_tool.find_highest_sub_prj_version(
+        "vara-llvm-project"
+    )
+    if research_tool.is_up_to_date():
+        print(f"{colors.green}VaRA major release version is up to date!")
+    else:
```
vulder: I think it would be enough to only output something if this is not the case but I'm not sure.
@boehmseb What do you think? Should we always output the version check or just when it fails?

varats/varats/tools/driver_build_setup.py
```diff
@@ -183,15 +205,15 @@ def main() -> None:
...
+        print_up_to_date_message(tool)
     if args.build:
         build_type = parse_string_to_build_type(args.buildtype)
         tool.build(build_type, __get_install_prefix(tool, args.installprefix))
         if tool.verify_install(__get_install_prefix(tool, args.installprefix)):
             print(f"{tool.name} was correctly installed.")
         else:
             print(f"Could not install {tool.name} correctly.")
-        tool.is_up_to_date()
+        print_up_to_date_message(tool)
```
vulder: Hmm, shouldn't we run this check before building the "possible" outdated version? Maybe it would be useful to ask the user if he wants to continue building should the version be out of date.

varats/varats/tools/research_tools/vara.py
```diff
@@ -207,76 +193,66 @@ def setup(self, source_folder: tp.Optional[Path], **kwargs: tp.Any) -> None:
...
+            if sub_prj_name == "VaRA":
+                LOG.warning(
+                    "No VaRA tag matching the release pattern was found."
+                )
+            if sub_prj_name == "vara-llvm-project":
                 LOG.warning(
-                    "No vara-llvm-project release branch name was "
-                    "found."
+                    "No vara-llvm-project release branch name was found."
                 )
```
vulder: Just create the error string with an f-string
```
warning_str = f"No {sub_prj_name} ..."
LOG.warning(warning_str)
```

vulder (changes requested):
Overall this looks ok, I plan a new release for the next comming weeks to I'll wait to merge this until the release is done :)

varats/varats/tools/driver_build_setup.py
boehmseb: I would expect such a message only if no update and no upgrade was done.

varats/varats/tools/driver_build_setup.py
boehmseb: Same goes for update btw.

cormensratio: Initially i did not intend to interrupt the build by asking the user if he wants to update and placed the update check at the end to minimize the risk for the user to skip it unintentionally. If interrupting the build process at the beginning for a release check prompt is okay, I think it would be the better solution.

vulder: Yes, we should do the check before.

Update it self than, does not need a release check.

varats/varats/tools/driver_build_setup.py
vulder: I thought so, then let's just print something in the case we have an outdated version.

varats/varats/tools/driver_build_setup.py
```diff
@@ -63,24 +63,55 @@ def update_term(text: str, enable_inline: bool = False) -> None:
...
+
+    if not research_tool.is_up_to_date():
+        print_up_to_date_message(research_tool)
+        print(
+            "Press:\n"
+            f"  {colors.bold}1{colors.bold.reset} to upgrade and continue.\n"
+            f"  {colors.bold}2{colors.bold.reset} to continue without "
+            f"upgrading.\n "
+        )
+        user_choice = input("> ")
```
vulder: Just use the choice util we already implemented in utils.

vulder (approved):

PR #432, opened 2021-01-28
Adds more c projects

cormensratio:
Adds GNU grep to c_projects.


varats/varats/projects/c_projects/grep.py
```diff
@@ -0,0 +1,75 @@
...
+    ]
+
+    @property
+    def binaries(self) -> tp.List[ProjectBinaryWrapper]:
+        """Return a list of binaries generated by the project."""
+
+        grep_source = local.path(self.source_of_primary)
+        return wrap_paths_to_binaries([
+            (grep_source / "src/grep", BinaryType.executable),
+        ])
```
vulder:
```suggestion
        return wrap_paths_to_binaries([
            ("src/grep", BinaryType.executable),
        ])
```
Paths for binaries are given relative to the project root.

varats/varats/projects/c_projects/grep.py
```diff
@@ -0,0 +1,75 @@
...
+        compiler = bb.compiler.cc(self)
+        with local.cwd(grep_source):
+            with local.env(CC=str(compiler)):
+                bb.watch(local["./bootstrap"])()
+                bb.watch(local["./configure"])("--disable-gcc-warnings")
+
+            bb.watch(make)("-j", get_number_of_jobs(bb_cfg()))
+            for binary in self.binaries:
+                if not Path("{binary}".format(binary=binary)).exists():
+                    print("Could not find {binary}")
```
vulder: This does not work. `binary` is of type `ProjectBinaryWrapper`, which is a path and a name. Hence, turning it into a string will never result in a correct Path.
If you want to do some checking use the new `verify_binaries` API that I build to prevent these kind of issues.

see: https://github.com/se-sic/VaRA-Tool-Suite/pull/436

vulder (changes requested):
Otherwise, looks good grep works on my setup.

vulder (approved):

PR #433, opened 2021-01-28
Introducing Commit Detection via Simple SZZ Algorithm

SimonFedick:
Adds PyDriller as dependency and uses PyDriller version of the SZZ algorithm to identify fix inducing/introducing changes


vulder (commented):
overall this looks quite simple with just using the pydriller API, however, we should carefully test this to not miss use their approach and check what results we get

tests/provider/test_bug_provider.py
```diff
@@ -122,12 +123,21 @@ def test_pygit_bug_creation(self):
             pygit2.Repository.revparse_single, return_value=issue_commit
         )
 
-        pybug = _create_corresponding_pygit_bug(
-            issue_event.commit_id, mock_repo, issue_event.issue.number
+        mock_pydrill_repo = create_autospec(pydriller.GitRepository)
+        mock_pydrill_repo.get_commits_last_modified_lines = create_autospec(
+            pydriller.GitRepository.get_commits_last_modified_lines,
+            return_value={}
```
boehmseb: Since you use this in multiple tests, it could make sense to make this a member (`self.mock_pydrill_repo`) and initialize it in a setup function.

tests/provider/test_bug_provider.py
```diff
@@ -332,11 +360,81 @@ def test_basic_repo(self):
...
+        expected_first = {"343bea18b421cfa2eb5945b2672f62f171abcc83"}
+        expected_second = {"343bea18b421cfa2eb5945b2672f62f171abcc83"}
+        expected_third = {"8fae311154a28b7928fe667b6aad09319259b1aa"}
+        expected_fourth = {
+            "c4b7bd9a2cedf1eb67d13be3cf4e826273cfe17b",
+            "8fae311154a28b7928fe667b6aad09319259b1aa"
+        }
+
+        self.assertEqual(pybug_first_intro_ids, expected_first)
+        self.assertEqual(rawbug_first_intro_ids, expected_first)
```
boehmseb: I think it  would make sense to test raw bugs and pygit bugs in separate tests, since two different API functions are tested here. But make sure to not introduce any duplicated code when splitting.

boehmseb (changes requested):
Implementation looks good, tests could use a little work.

boehmseb (approved):

varats-core/varats/provider/bug/bug.py
```diff
@@ -103,6 +108,116 @@ def __hash__(self) -> int:
...
+
+    def is_cleared(self) -> bool:
+        """Returns whether all suspects inside this tuple have been cleared."""
+        return len(self.__uncleared_suspects) == 0
+
+    def extract_next_uncleared_suspect(self) -> pygit2.Commit:
+        """
+        Extracts one uncleared suspect from the set and returns it.
+
+        Returns None if the set is empty.
```
boehmseb: If so, then the return type should be `Optional[pygit2.Commit]`.
But if I read this correctly, an exception is raised if the set is empty.

Also, please use the correct docstring syntax for returns:
```
Returns:
    ...
```

varats-core/varats/provider/bug/bug.py
```diff
@@ -248,27 +479,50 @@ def _filter_all_issue_pygit_bugs(
...
+
+    # CLASSIFY SUSPECTS
+    for suspect_tuple in suspect_tuples:
+        while not suspect_tuple.is_cleared():  # iterate over uncleared suspects
+            suspect = suspect_tuple.extract_next_uncleared_suspect()
+
+            # partial fix?
+            for other_tuple in suspect_tuples:
+                if suspect == other_tuple.fixing_commit:
+                    suspect_tuple.clear_suspect(suspect)
```
boehmseb: You may want to terminate the for loop here.

varats-core/varats/provider/bug/bug.py
```diff
@@ -248,27 +479,50 @@ def _filter_all_issue_pygit_bugs(
...
+
+            # weak suspect?
+            for other_tuple in suspect_tuples:
+                if suspect in other_tuple.non_suspects:
+                    suspect_tuple.clear_suspect(suspect)
```
boehmseb: You may want to terminate the for loop here.

varats-core/varats/provider/bug/bug.py
```diff
@@ -277,20 +531,43 @@ def _filter_all_issue_raw_bugs(
...
+
+    # CLASSIFY SUSPECTS
+    for suspect_tuple in suspect_tuples:
+        while not suspect_tuple.is_cleared():  # iterate over uncleared suspects
+            suspect = suspect_tuple.extract_next_uncleared_suspect()
+
+            # partial fix?
+            for other_tuple in suspect_tuples:
+                if suspect == other_tuple.fixing_commit:
+                    suspect_tuple.clear_suspect(suspect)
```
boehmseb: You may want to terminate the for loop here.

varats-core/varats/provider/bug/bug.py
```diff
@@ -277,20 +531,43 @@ def _filter_all_issue_raw_bugs(
...
+
+            # weak suspect?
+            for other_tuple in suspect_tuples:
+                if suspect in other_tuple.non_suspects:
+                    suspect_tuple.clear_suspect(suspect)
```
boehmseb: You may want to terminate the for loop here.

boehmseb (changes requested):
Looks good overall, but I found a couple of smaller things.

varats-core/varats/provider/bug/bug.py
SimonFedick: This is a remaining piece of the old interface where the return type was an Optional type, but I figured an `is_cleared()` function to check whether there are still uncleared suspects and then just allowing to call this function if the tuple isn't cleared yet was more convenient for me than needing to check if the return type of this function is `None` or not every time.
Do you think I should adjust the docstring or go back to the old interface of this?

boehmseb: Just fix the docstring.

boehmseb (approved):

tests/provider/test_bug_provider.py
```diff
@@ -27,6 +31,28 @@ class TestBugDetectionStrategies(unittest.TestCase):
     """Test several parts of the bug detection strategies used by the
     BugProvider."""
 
+    def setUp(self) -> None:
+        """Set up repo dummy objects needed for multiple tests."""
+
+        # Method that creates simple pygit2 Commit mocks for given ID
+        def mock_revparse(commit_id: str):
```
vulder: Just make the comment to a function doc

varats-core/varats/provider/bug/bug.py
SimonFedick: Great! That's what I did.

varats-core/varats/provider/bug/bug.py
```diff
@@ -187,59 +294,175 @@ def load_issue_events(github: Github) -> 'PaginatedList[IssueEvent]':
...
+    )
+
+    for _, introducing_set in blame_dict.items():
+        introducing_ids.update(introducing_set)
+
+    return RawBug(closing_commit, introducing_ids, issue_id)
+
 
-    return RawBug(closing_commit, introducing_ids, issue_number)
+def _find_corresponding_pygit_suspect_tuple(
```
vulder, line 329: This function has no test coverage but is central to functionality. I think we should try to come up with a way write basic tests for this.
@boehmseb What do you think?

varats-core/varats/provider/bug/bug.py
```diff
@@ -187,59 +294,175 @@ def load_issue_events(github: Github) -> 'PaginatedList[IssueEvent]':
...
+                    )
+
+        return PygitSuspectTuple(
+            fixing_commit, non_suspect_commits, suspect_commits,
+            issue_event.issue.number
+        )
+    return None
+
+
+def _find_corresponding_raw_suspect_tuple(
```
vulder: same here

varats-core/varats/provider/bug/bug.py
```diff
@@ -187,59 +294,175 @@ def load_issue_events(github: Github) -> 'PaginatedList[IssueEvent]':
...
+                    non_suspect_commits.add(introducing_id)
+
+        return RawSuspectTuple(
+            fixing_commit, non_suspect_commits, suspect_commits,
+            issue_event.issue.number
+        )
+    return None
 
 
 def _filter_all_issue_pygit_bugs(
```
vulder, line 382: same here

varats-core/varats/provider/bug/bug.py
```diff
@@ -248,27 +471,58 @@ def _filter_all_issue_pygit_bugs(
...
+                suspect_tuple.clear_suspect(suspect)
+
+        pygit_bug = suspect_filter_function(suspect_tuple)
+        if pygit_bug:
+            resulting_pygit_bugs.add(pygit_bug)
 
     return frozenset(resulting_pygit_bugs)
 
 
 def _filter_all_issue_raw_bugs(
```
vulder: this implementation is the same as the above, expect for small details, can we merge the two?

vulder (changes requested):
Code looks ok to me. Am I seeing this right that the core parts are not covered by the tests or is this a codecov issue?

boehmseb:
> Am I seeing this right that the core parts are not covered by the tests or is this a codecov issue?

There are parts that are not covered, but definitely not as many as codecov claims. The coverage report seems very fishy to me and does not agree with the coverage report I just created on my local machine. My local report tells me, that `bug.py` has a total coverage of 63% and at least some of the methods you marked are definitely covered.

varats-core/varats/provider/bug/bug.py
boehmseb: This function is covered except for the for-loops. I guess it would be quite easy to extend existing tests to completely cover this.

varats-core/varats/provider/bug/bug.py
boehmseb: Also covered, except for the while-loop.

vulder:
I updated to the current vara-dev version, there we fixed the coverage tracking issue.

vulder (approved):

tests/provider/test_bug_provider.py
```diff
@@ -217,13 +275,24 @@ def accept_pybugs(sus_tuple: PygitSuspectTuple):
...
+            pybug_fix_ids = set(pybug.fixing_commit.hex for pybug in pybugs)
+            expected_fix_ids = {"1239", "1240"}
+
+            intro_first_bug = set
+            intro_second_bug = set
+            for pybug in pybugs:
+                if pybug.fixing_commit.hex == "1239":
+                    intro_second_bug = pybug.introducing_commits
+                if pybug.fixing_commit.hex == "1240":
+                    intro_first_bug = pybug.introducing_commits
```
boehmseb:
```suggestion
            intro_first_bug = set()
            intro_second_bug = set()
            for pybug in pybugs:
                if pybug.fixing_commit.hex == event_close_second.commit_id:
                    intro_second_bug = pybug.introducing_commits
                if pybug.fixing_commit.hex == event_close_first.commit_id:
                    intro_first_bug = pybug.introducing_commits
```

vulder (approved):

PR #434, opened 2021-01-29
Add an understandable error message to vara-cs when no paper-config exits

Sinerum:
Add an error message explaining how to create a paper-config when no paper-config was found by vara-cs
closes se-passau/VaRA#719


boehmseb (changes requested):
Can you please make this a decorator so that we can use this like this:
```python
@configuration_lookup_error_handler
def __casestudy_exec_command(...):
    ...
```

varats/varats/tools/tool_util.py
```diff
@@ -64,3 +65,19 @@ def get_research_tool(
 def get_supported_research_tool_names() -> tp.List[str]:
     """Returns a list of all supported research tools."""
     return ["phasar", "vara"]
+
+
+def configuration_lookup_error_handler(func) -> tp.Any:
+    """Wrapper for drivers to catche internal Exceptions and provide a helpful
+    message to the user."""
+
+    def wrapper_configuration_lookup_error_handler(*args, **kwargs):
```
boehmseb:
```suggestion
    @wraps(func)
    def wrapper_configuration_lookup_error_handler(*args, **kwargs):
```

See here for an explanation: https://stackoverflow.com/questions/308999/what-does-functools-wraps-do

boehmseb (changes requested):
Looks good now apart from the two missing type annotations and the small detail I annotated.

vulder (approved):
Looks good from my side, please clean up the open issues (mypy).

varats/varats/tools/tool_util.py
```diff
@@ -64,3 +66,22 @@ def get_research_tool(
 def get_supported_research_tool_names() -> tp.List[str]:
     """Returns a list of all supported research tools."""
     return ["phasar", "vara"]
+
+
+def configuration_lookup_error_handler(func: tp.Any) -> tp.Any:
+    """Wrapper for drivers to catche internal Exceptions and provide a helpful
```
boehmseb:
```suggestion
    """Wrapper for drivers to catch internal Exceptions and provide a helpful
```

varats/varats/tools/tool_util.py
```diff
@@ -64,3 +66,22 @@ def get_research_tool(
...
+    message to the user."""
+
+    @wraps(func)
+    def wrapper_configuration_lookup_error_handler(
+        *args: tp.Any, **kwargs: tp.Any
+    ) -> tp.Any:
```
boehmseb: I think, we should restrict the return type to be a function:
```suggestion
    ) -> tp.Callable[..., tp.Any]:
```

boehmseb (changes requested):

varats/varats/tools/tool_util.py
```diff
@@ -64,3 +66,22 @@ def get_research_tool(
...
+
+
+def configuration_lookup_error_handler(func: tp.Any) -> tp.Any:
+    """Wrapper for drivers to catch internal Exceptions and provide a helpful
+    message to the user."""
+
+    @wraps(func)
+    def wrapper_configuration_lookup_error_handler(
+        *args: tp.Any, **kwargs: tp.Any
+    ) -> tp.Callable[..., tp.Any]:
```
boehmseb, line 78: My mistake. Either, we allow functions with return types to be wrapped, then we need to add a `return` in line 80, or we do not allow returns, then we need to do this change:
```suggestion
    ) -> tp.Callable[..., None]:
```

boehmseb (approved):

PR #436, opened 2021-02-08
Implements utility function to verify binaries

vulder:
Adds new utility function that checks if all the specified binaries
where correctly produced after compilation.

### DIFF
--- a/docs/source/vara-ts-api/projects.rst
+++ b/docs/source/vara-ts-api/projects.rst
@@ -95,7 +95,7 @@ an analysis, a list of
 :class:`~varats.project.project_util.ProjectBinaryWrapper`\ s is returned,
 where every wrapper specifies the name, location of the binary, and the
 :class:`~varats.project.project_util.BinaryType`.
-To guarantee that all specified binaries where produced during compilation, one
+To guarantee that all specified binaries were produced during compilation, one
 should add a :func:`~varats.project.project_util.verify_binaries` call after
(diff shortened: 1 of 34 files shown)


varats/varats/projects/c_projects/gravity.py
```diff
@@ -103,13 +104,17 @@ def __compile_cmake(self) -> None:
                 bb.watch(cmake)("-G", "Unix Makefiles", ".")
             bb.watch(make)("-j", get_number_of_jobs(bb_cfg()))
 
+            verify_binaries(self)
```
boehmseb, line 107: I think we should add this to the example in the docs (vara-ts-api/projects.rst) maybe with an explaining comment.

boehmseb (changes requested):

docs/source/vara-ts-api/projects.rst
```diff
@@ -92,6 +95,13 @@ an analysis, a list of
 :class:`~varats.project.project_util.ProjectBinaryWrapper`\ s is returned,
 where every wrapper specifies the name, location of the binary, and the
 :class:`~varats.project.project_util.BinaryType`.
+To guarantee that all specified binaries where produced during compilation, one
```
boehmseb:
```suggestion
To guarantee that all specified binaries were produced during compilation, one
```

boehmseb (approved):
Just one more typo :)

PR #441, opened 2021-02-10
Enable vara container install

boehmseb:
Resolves se-passau/VaRA#718
The dependency declarations in this PR will also be useful for se-passau/VaRA#626


varats/varats/tools/research_tools/vara.py
```diff
@@ -256,14 +264,27 @@ def verify_install(self, install_location: Path) -> bool:
...
-        raise NotImplementedError("See se-passau/VaRA#718")
+        if not self.verify_install(self.install_location()):
+            raise AssertionError(
+                "VaRA is not correctly installed on your system."
+            )
+
+        container_vara_dir = containers.CONTAINER_VARATS_ROOT / "tools/VaRA"
+        layers.run(*(self.DEPENDENCIES.get_install_command(distro).split(" ")))
+        layers.copy_([str(self.install_location())], str(container_vara_dir))
+        layers.env(BB_ENV=f"{{PATH: [{str(container_vara_dir / 'bin')}]}}")
```
boehmseb: If we do it like this, we can set `BB_ENV` only at one place, since every call to `layers.env()` overwrites existing env vars in the container. What's your opinion on this @vulder

varats/varats/tools/research_tools/research_tool.py
```diff
@@ -29,6 +30,38 @@
...
+        Given a distro, return a command how the dependencies can be installed.
+        Args:
+            distro: the distro to use
+
+        Returns:
+            the command how the dependencies can be installed
+        """
+        return _install_commands[distro] + " " + " ".join(
+            self.__dependencies[distro]
+        )
```
vulder: Maybe use a f string here

varats/varats/tools/research_tools/research_tool.py
```diff
@@ -29,6 +30,38 @@
...
+    Distro.ARCH: "pacman -S --noconfirm"
+}
+
+
+class Dependencies:
+    """Models the dependencies for a research tool."""
+
+    def __init__(self, dependencies: tp.Dict[Distro, tp.List[str]]):
+        self.__dependencies = dependencies
+
```
vulder, line 51: Does it make sense to add a `add` method for a new dependency?

varats/varats/tools/research_tools/research_tool.py
```diff
@@ -29,6 +30,38 @@
...
+    ARCH = "arch"
+
+
+_install_commands = {
+    Distro.DEBIAN: "apt install -y",
+    Distro.ARCH: "pacman -S --noconfirm"
+}
+
+
+class Dependencies:
```
vulder, line 46: We should include 1-2 tests to check that the install strings are correctly generated.

varats/varats/tools/research_tools/vara.py
```diff
@@ -110,6 +113,11 @@ class VaRA(ResearchTool[VaRACodeBase]):
     Find the main repo online on github: https://github.com/se-passau/VaRA
     """
 
+    DEPENDENCIES = Dependencies({
+        Distro.DEBIAN: ["libboost-all-dev"],
+        Distro.ARCH: ["boost"]
+    })
```
vulder:
```suggestion
    __DEPENDENCIES = Dependencies({
        Distro.DEBIAN: ["libboost-all-dev"],
        Distro.ARCH: ["boost"]
    })
```
Should we make this private? Otherwise, users think they can change this.

varats/varats/tools/research_tools/vara.py
vulder: Hmm, one location to define it seems fine to me. But maybe we should add code that checks what is defined in the current BB env not that we miss something. Not related to the path by maybe something like this.
BB_ENV = { 'enable_instrumentation': "FooOn"}

... inside the container ... a binary could look for the env var to enable/disable instrumentation code

varats/varats/tools/research_tools/vara.py
```diff
@@ -256,14 +264,27 @@ def verify_install(self, install_location: Path) -> bool:
...
         """
-        raise NotImplementedError("See se-passau/VaRA#718")
+        if not self.verify_install(self.install_location()):
+            raise AssertionError(
+                "VaRA is not correctly installed on your system."
+            )
+
+        container_vara_dir = containers.CONTAINER_VARATS_ROOT / "tools/VaRA"
+        layers.run(*(self.DEPENDENCIES.get_install_command(distro).split(" ")))
+        layers.copy_([str(self.install_location())], str(container_vara_dir))
```
vulder: I think, we still need to handle the copying of shared libraries that are provided via the LD_LIBRARY_PATH

varats/varats/tools/research_tools/vara.py
```diff
@@ -110,6 +113,11 @@ class VaRA(ResearchTool[VaRACodeBase]):
     Find the main repo online on github: https://github.com/se-passau/VaRA
     """
 
+    DEPENDENCIES = Dependencies({
+        Distro.DEBIAN: ["libboost-all-dev"],
```
vulder: Syntax looks good to me :)

vulder (changes requested):

varats/varats/tools/research_tools/research_tool.py
boehmseb: I don't see any use case for this right now. Currently, we only want to use them for research tools, and there all dependencies are known statically.
But this class will get additional functions anyway at some point (at least when we fix se-passau/VaRA#626).

varats/varats/tools/research_tools/vara.py
boehmseb: When we use older containers, we might. But for a debian 10 container (the one I tested with) I can simply install boost as a dependency and I don't need to copy anything.
That's why I committed it that way.
Also, I'm not quite sure what's the best way to do this. Maybe look at the BB config and if there is a entry for `LD_LIBRARY_PATH` in the env copy the contents to the container? Or add a mountpoint and always let the `LD_LIBRARY_PATH` point there?

varats/varats/tools/research_tools/vara.py
boehmseb: Well, that's actually a second point you are talking about here.
But I guess this will become a longer discussion we should do in person.

varats/varats/tools/research_tools/research_tool.py
vulder: kk, then no

varats/varats/tools/research_tools/vara.py
boehmseb: Not sure. Could be useful to others to query the dependencies of a tool though...

varats/varats/tools/research_tools/vara.py
boehmseb: Should be fixed once and for all by the new `BaseImageCreationContext` class.

varats/varats/containers/containers.py
```diff
@@ -36,34 +39,146 @@ def prepare_buildah() -> ConcreteCommand:
...
+            'apt', 'install', '-y', 'python3', 'python3-dev', 'python3-pip',
+            'musl-dev', 'git', 'gcc', 'libgit2-dev', 'libffi-dev',
+            'libyaml-dev', 'clang'
+        )
 }
 
-CONTAINER_VARATS_ROOT: Path = Path("/varats_root/")
-CONTAINER_BB_ROOT: Path = Path("/app/")
 
+class BaseImageCreationContext():
```
boehmseb, line 73: If we feel the need to, we could hide the complete BB-buildah interface behind this context and only allow a couple of "safe" operations. Although only during base image creation. Projects are still allowed to mess everything up if they want to...
Anyway, this should really simplify the research tool integration.

varats/varats/tools/research_tools/vara.py
boehmseb: Just copying the thing for now

varats/varats/tools/research_tools/vara.py
vulder: That's true but then we should provide a get that copies the list. From my current understanding I only see read_only access making sense here.

vulder (approved):
Regarding the context -> sounds good

varats/varats/tools/research_tools/vara.py
boehmseb: Would make sense to include that as an abstract class-level function (yes, that is a thing with python's abc) in the base class then.

vulder (approved):

PR #443, opened 2021-02-22
Adds blame_diff_library_interaction_database

cormensratio:
Adds a blame_diff_library_interaction_database to store blame library interactions of blame report diffs.


varats/varats/data/databases/blame_diff_library_interaction_database.py
```diff
@@ -0,0 +1,118 @@
...
+from pathlib import Path
+
+import pandas as pd
+
+from varats.data.cache_helper import build_cached_report_table
+from varats.data.databases.blame_diff_metrics_database import (
+    _build_report_pairs_tuple,
+    _id_from_paths,
+    _timestamp_from_paths,
+    _compare_timestamps,
```
vulder: The current state of how the functions are used across modules is a bit mixed up.

@boehmseb How should we handle this case?
Options:
a) Keep them protected on a module level (original module)
b) Define them in `the blame_diff_metrics_database` module, similar for others we treat the most basic module as the one that has the utility functions for this group of databases. But make the functions public so it's clear they can be reused.
c) Put both database impls (old + BDLibIterDB) into one module, and keep them protected.
d) Create a utility module for all databases, could be a bit over generic.
e) Create a utility module for the specific blame_diff databases that both import then?

varats/varats/data/databases/blame_diff_metrics_database.py
```diff
@@ -37,6 +37,127 @@
...
+
+ReportPairTupleList = tp.List[tp.Tuple[Path, Path]]
+
+
+def _build_report_pairs_tuple(
+    project_name: str, commit_map: CommitMap, case_study: tp.Optional[CaseStudy]
+) -> tp.Tuple[ReportPairTupleList, ReportPairTupleList]:
+    report_files, failed_report_files = _build_report_files_tuple(
+        project_name, case_study
+    )
```
vulder, line 148: We should add some docs here that describe what the function does.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -543,8 +545,8 @@ def _save_figure(
         file_name = f"{file_name}_{padded_idx_str}"
 
         figure.render(
-            filename=str(plot_dir) + "/" + plot_subdir + "/" + file_name,
-            directory=plot_dir,
+            filename=file_name,
+            directory=str(plot_dir) + "/" + plot_subdir,
```
vulder: assuming `plot_dir` is a `Path`
It would be better to first concat parts of the path and then convert it into a string.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -1337,7 +1352,7 @@ def plot(self, view_mode: bool) -> None:
                 "ignored."
             )
         self.__graph = self._graphviz_plot(
-            view_mode=view_mode, show_edge_weight=True
+            view_mode=view_mode, show_edge_weight=True, show_blame_diff=True
```
vulder: Did you want to make this configurable from the outside?

vulder (changes requested):
Could you also please put the tweaked command line string for the graph generation into the docs.

varats/varats/data/databases/blame_diff_metrics_database.py
cormensratio: I'll add the docs when we have decided how we handle the function scope.

varats/varats/plots/blame_interaction_degree.py
cormensratio: Yes, my intention is to allow the user to pass params that show only the normal interaction plot or only the blame diff plot or both together, where the latter one results in a blame diff overlay.

varats/varats/data/databases/blame_diff_library_interaction_database.py
boehmseb: Either b) or e). I think the amount of code we reuse would justify a utility module, but we could also leave it in the `blame_diff_metrics_database` module.

varats/varats/data/databases/blame_diff_metrics_database.py
boehmseb: That goes for all these functions btw., they will become public regardless of the place we put them.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -744,35 +791,73 @@ class BlameLibraryInteraction(Plot):
...
+
+            if blame_interactions and blame_diff:
+                inter_df = self._get_interaction_data(False)
+                diff_df = self._get_interaction_data(True)
+                return _add_diff_amount_col_to_df(inter_df, diff_df)
+
+            if blame_interactions:
+                return self._get_interaction_data(False)
+
+            return self._get_interaction_data(True)
```
```suggestion
            if blame_interactions:
                inter_df = self._get_interaction_data(False)
            else:
                inter_df = self._get_interaction_data(True)
                
            if  blame_diff:
                diff_df = self._get_interaction_data(True)
                return _add_diff_amount_col_to_df(inter_df, diff_df)

            return inter_df
```
I think this could make it a bit more readable by separating the two cases

tests/data/databases/test_blame_diff_metrics_database.py
```diff
@@ -0,0 +1,117 @@
...
+                "_9e238675-ee7c-4325-8e9f-8ccf6fd3f05c_success.yaml"
+            ),
+            Path(__file__).parents[2] / Path(
+                "TEST_INPUTS/results/xz/BR-xz-xz-c5c7ceb08a"
+                "_77a6c5bc-e5c7-4532-8814-70dbcc6b5dda_success.yaml"
+            ),
+            Path(__file__).parents[2] / Path(
+                "TEST_INPUTS/results/xz/BR-xz-xz-ef364d3abc"
+                "_feeeecb2-1826-49e5-a188-d4d883f06d00_success.yaml"
+            )
```
vulder: Why do you need such a hacky way to access the parent folder? If you want to access the test data, there should be some helper function in utils.

vulder (changes requested):

varats/varats/plots/blame_interaction_degree.py
cormensratio: Should be resolved after https://github.com/se-sic/VaRA/issues/545 is resolved.

vulder (approved):

PR #444, opened 2021-02-24
Adds a graphviz plot feature to show interactions of the specified commit only

cormensratio:
Adds a graphviz plot feature to show interactions of the specified commit only.

### DIFF
--- a/varats/varats/plots/blame_interaction_degree.py
+++ b/varats/varats/plots/blame_interaction_degree.py
@@ -652,6 +652,24 @@ def _build_sankey_figure(
     return fig
 
 
+def _get_verified_commit_hash(
+    df: pd.DataFrame, show_interactions_of_commit: str
+) -> str:
+    unique_base_hashes = list(
+        np.unique([str(base_hash) for base_hash in df.base_hash])
+    )
+    unique_inter_hashes = list(
+        np.unique([str(inter_hash) for inter_hash in df.inter_hash])
+    )
+    all_unique_hashes = list(
+        np.unique(unique_base_hashes + unique_inter_hashes)
+    )
+
+    return _get_verified_revision(
+        show_interactions_of_commit, all_unique_hashes
+    )
+
+
 LibraryToHashesMapping = tp.Dict[str, tp.List[str]]
 
 
@@ -659,8 +677,14 @@ def _build_graphviz_edges(
     df: pd.DataFrame,
     graph: Digraph,
     show_edge_weight: bool,
-    edge_weight_threshold: tp.Optional[EdgeWeightThreshold] = None
+    edge_weight_threshold: tp.Optional[EdgeWeightThreshold] = None,
+    show_interactions_of_commit: tp.Optional[str] = None
 ) -> LibraryToHashesMapping:
+
+    if show_interactions_of_commit is not None:
+        show_interactions_of_commit = _get_verified_commit_hash(
+            df, show_interactions_of_commit
+        )
 
     base_lib_names = _get_distinct_base_lib_names(df)
     inter_lib_names = _get_distinct_inter_lib_names(df)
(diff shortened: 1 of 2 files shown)


varats/varats/plots/blame_interaction_degree.py
```diff
@@ -652,16 +652,40 @@ def _build_sankey_figure(
     return fig
 
 
+def _get_verified_commit_hash(
```
vulder: What do you mean with "verified" we don't differentiate between verified and non verified commit hashes here (+ this is a specific github feautre). We should not confuse the terms here.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -652,16 +652,40 @@ def _build_sankey_figure(
...
+        np.unique([str(base_hash) for base_hash in df.base_hash])
+    )
+    unique_inter_hashes = list(
+        np.unique([str(inter_hash) for inter_hash in df.inter_hash])
+    )
+    all_unique_hashes = list(
+        np.unique(unique_base_hashes + unique_inter_hashes)
+    )
+
+    return _get_verified_revision(
```
vulder: Maybe a quick rename here would also help :D

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -652,16 +652,40 @@ def _build_sankey_figure(
...
 LibraryToHashesMapping = tp.Dict[str, tp.List[str]]
 
 
 def _build_graphviz_edges(
     df: pd.DataFrame,
     graph: Digraph,
     show_edge_weight: bool,
-    edge_weight_threshold: tp.Optional[EdgeWeightThreshold] = None
+    edge_weight_threshold: tp.Optional[EdgeWeightThreshold] = None,
+    show_interactions_of_commit: tp.Optional[str] = None
```
```suggestion
    show_only_interactions_of_commit: tp.Optional[str] = None
```
Otherwise, it's not clear from the param that we remove all interactions that don't belong to the commit.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -701,13 +734,15 @@ def _build_graphviz_fig(
     show_edge_weight: bool,
     shown_revision_length: int,
     edge_weight_threshold: tp.Optional[EdgeWeightThreshold] = None,
+    show_interactions_of_commit: tp.Optional[str] = None,
```
vulder: same

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -768,6 +803,7 @@ def _graphviz_plot(
         show_edge_weight: bool = False,
         shown_revision_length: int = 10,
         edge_weight_threshold: tp.Optional[EdgeWeightThreshold] = None,
+        show_interactions_of_commit: tp.Optional[str] = None,
```
vulder: same

vulder (changes requested):

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -676,6 +710,15 @@ def _build_graphviz_edges(
         inter_hash = row['inter_hash']
         inter_lib = row['inter_lib']
 
+        base_inter_hash_tuple = (base_hash, inter_hash)
+
+        # Skip edges that do not connect with the specified
+        # ``show_only_interactions_of_commit`` node.
+        if show_only_interactions_of_commit is not None and (
+            show_only_interactions_of_commit not in base_inter_hash_tuple
+        ):
```
vulder, line 735: But does this not also skip interactions where the "only_commit" is involved as an interacting hash?

cormensratio: No i don't think so, because we only skip an interaction if the "only_commit" is neither the base commit nor the interacting commit.

vulder (approved):

PR #447, opened 2021-03-04
Add SZZUnleashed research tool, experiment, and report

boehmseb:
This PR adds SZZUnleashed as a research tool.
It also comes with an experiment to run SZZUnleashed and a SZZReport for the results.

The implementation also anticipates to add more SZZ tools in the future.


varats-core/varats/provider/bug/bug.py
```diff
@@ -187,7 +220,9 @@ def load_issue_events(github: Github) -> 'PaginatedList[IssueEvent]':
 def _create_corresponding_pygit_bug(
     closing_commit: str,
     project_repo: pygit2.Repository,
-    issue_number: tp.Optional[int] = None
+    issue_number: tp.Optional[int] = None,
+    creationdate: tp.Optional[datetime] = None,
+    resolutiondate: tp.Optional[datetime] = None
```
vulder, line 225: I assume you found a case where we cannot get this information, or why do you use optional here? At least from my point of view, at least creation date should always be there.

varats/varats/data/reports/szz_report.py
```diff
@@ -0,0 +1,116 @@
...
+            version_header.raise_if_not_type("SZZReport")
+            version_header.raise_if_version_is_less_than(1)
+            raw_report = next(documents)
+            if not raw_report["szz_tool"] == szz_tool:
+                raise AssertionError(
+                    "Report was not created with the correct tool."
+                )
+            self.__bugs: tp.Dict[str, RawBug] = {}
+            for fix, introducers in raw_report["bugs"].items():
+                self.__bugs[fix] = RawBug(fix, set(introducers), None)
```
vulder, line 36: It's probably better to move this into a factory function and pass the values to init. This makes it easier testable.

varats/varats/experiments/szz/szz_unleashed_experiment.py
```diff
@@ -0,0 +1,183 @@
...
+    DESCRIPTION = "Prepares data needed for running SZZUnleashed."
+
+    def __init__(self, project: Project):
+        super().__init__(obj=project, action_fn=self.prepare_szz_data)
+
+    def prepare_szz_data(self) -> actions.StepResult:
+        """Prepare data needed for running SZZUnleashed."""
+        results_dir = Path(self.obj.source_of_primary).parent
+
+        bug_provider = BugProvider.get_provider_for_project(self.obj)
```
vulder: I like this API :smile:

varats/varats/experiments/szz/szz_unleashed_experiment.py
```diff
@@ -0,0 +1,183 @@
...
+                datetime.fromtimestamp(bug.fixing_commit.commit_time)
+            )
+            creationdate = fix_date(
+                bug.creationdate
+            ) if bug.creationdate else commitdate
+            resolutiondate = fix_date(
+                bug.resolutiondate
+            ) if bug.resolutiondate else commitdate
+            fixers_dict[str(bug.fixing_commit.id)] = {
+                "hash": str(bug.fixing_commit.id),
```
vulder, line 65: Is there a specific reason why you put the hash also into the value? Theoretically, it should be enough if we have it as key to the map.

varats/varats/experiments/szz/szz_unleashed_experiment.py
```diff
@@ -0,0 +1,183 @@
...
+
+    NAME = "PrepareSZZUnleashedData"
+    DESCRIPTION = "Prepares data needed for running SZZUnleashed."
+
+    def __init__(self, project: Project):
+        super().__init__(obj=project, action_fn=self.prepare_szz_data)
+
+    def prepare_szz_data(self) -> actions.StepResult:
+        """Prepare data needed for running SZZUnleashed."""
+        results_dir = Path(self.obj.source_of_primary).parent
```
vulder: Not sure if `results_dir` is a good name, maybe `bb_build_dir`? or run_dir?
My first thought here was the vara results dir.

varats/varats/experiments/szz/szz_unleashed_experiment.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def __init__(self, project: Project):
+        super().__init__(obj=project, action_fn=self.run_szz)
+
+    def run_szz(self) -> actions.StepResult:
+        """Prepare data needed for running SZZUnleashed."""
+        results_dir = Path(self.obj.source_of_primary).parent
+        szzunleashed_jar = SZZUnleashed.install_location(
+        ) / SZZUnleashed.get_jar_name()
+        with local.cwd(results_dir):
```
vulder:
```suggestion
        ) / SZZUnleashed.get_jar_name()

        with local.cwd(results_dir):
```

varats/varats/experiments/szz/szz_unleashed_experiment.py
```diff
@@ -0,0 +1,183 @@
...
+            result_dir=str(bb_cfg()["varats"]["outfile"]),
+            project_dir=str(project.name)
+        )
+        mkdir("-p", varats_result_folder)
+
+        bb_results_folder = Path(project.source_of_primary).parent
+        with (bb_results_folder / "results" /
+              "fix_and_introducers_pairs.json").open("r") as result_json:
+            szz_result = json.load(result_json)
+        bugs: tp.Dict[str, tp.Set[str]] = {}
```
vulder:
```suggestion
            szz_result = json.load(result_json)

        bugs: tp.Dict[str, tp.Set[str]] = {}
```

varats/varats/experiments/szz/szz_unleashed_experiment.py
```diff
@@ -0,0 +1,183 @@
...
+        # entries are lists of the form [<fix>, <introducing>]
+        for result_entry in szz_result:
+            bugs.setdefault(result_entry[0], set())
+            bugs[result_entry[0]].add(result_entry[1])
+        bugs = {k: set(v) for k, v in bugs.items()}
+        raw_szz_report = {"szz_tool": "SZZUnleashed", "bugs": bugs}
+
+        result_file = SZZUnleashedReport.get_file_name(
+            project_name=str(project.name),
+            binary_name="",
```
vulder: Does this work when loading reports? Not sure if our regex allows for this.

varats/varats/experiments/szz/szz_unleashed_experiment.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+
+    NAME = "SZZUnleashed"
+
+    REPORT_TYPE = SZZReport
+
+    @classmethod
+    def sample(cls, prj_cls: ProjectT) -> tp.List[source.VariantContext]:
+        variants = list(source.product(*prj_cls.SOURCE))
+        return [source.context(*variants[0])]
```
vulder, line 185: Hmm, how do you plan on running this experiment for a case study with multiple version? Or do you just want to run this manually?

varats/varats/experiments/szz/szz_unleashed_experiment.py
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, project: Project):
+        super().__init__(obj=project, action_fn=self.run_szz)
+
+    def run_szz(self) -> actions.StepResult:
+        """Prepare data needed for running SZZUnleashed."""
+        results_dir = Path(self.obj.source_of_primary).parent
+        szzunleashed_jar = SZZUnleashed.install_location(
+        ) / SZZUnleashed.get_jar_name()
+        with local.cwd(results_dir):
+            bb.watch(java)(
```
vulder: Does it make sense to wrap an error handler around this?

varats/varats/tools/research_tools/szz_unleashed.py
```diff
@@ -0,0 +1,144 @@
...
+
+class SZZUnleashedCodeBase(CodeBase):
+    """Layout of the SZZUnleashed code base."""
+
+    def __init__(self, base_dir: Path):
+        super().__init__(
+            base_dir, [
+                SubProject(
+                    self, "SZZUnleashed",
+                    "https://github.com/wogscpar/SZZUnleashed.git", "origin",
```
vulder: Maybe we should point this to your fork?

vulder (changes requested):

varats-core/varats/provider/bug/bug.py
boehmseb: This is only there if we have an issue for it. If it was identified by commit message, this information is not available.

varats/varats/experiments/szz/szz_unleashed_experiment.py
boehmseb: Well, the same name is used in other experiments...

varats/varats/experiments/szz/szz_unleashed_experiment.py
boehmseb: That's just the format the tool needs. Technically, the keys are the issue ids/names, but since we don not necessarily have an issue for each fix, I just use the commit hash of the fixing commit.

varats/varats/experiments/szz/szz_unleashed_experiment.py
boehmseb: Could be. I mainly left it out while developing and never remembered to reconsider this.

varats/varats/experiments/szz/szz_unleashed_experiment.py
boehmseb: Yes, that's fine. Only "ugly" thing is that you have two consecutive `--` in the name. But since I don't have any binary name and there is no way to change the name template, I don't know what to do else.

varats/varats/experiments/szz/szz_unleashed_experiment.py
boehmseb: It does not make sense to run this for multiple revisions, because once a bug fixing commit is there, it cannot change. So using the newest possible revision is the best choice because it gets you the most data.

varats/varats/tools/research_tools/szz_unleashed.py
boehmseb: Already done locally, but not pushed yet.

varats/varats/experiments/szz/szz_unleashed_experiment.py
vulder: Yeah, maybe we should change this. Seems like an artefact of "historical growth"

varats/varats/experiments/szz/szz_unleashed_experiment.py
vulder: If it works, fine with me. An alternative could also be `none`, I used for example `all` for the compiler errors where we don't have a binary yet.

varats-core/varats/provider/bug/bug.py
```diff
@@ -25,12 +26,18 @@ class PygitBug:
...
 
     def __init__(
-        self, fixing_commit: pygit2.Commit,
-        introducing_commits: tp.Set[pygit2.Commit], issue_id: tp.Optional[int]
+        self,
+        fixing_commit: pygit2.Commit,
+        introducing_commits: tp.Set[pygit2.Commit],
+        issue_id: tp.Optional[int],
+        creationdate: tp.Optional[datetime] = None,
+        resolutiondate: tp.Optional[datetime] = None
```
SimonFedick, line 34: That's an oversight on my side, wouldn't we also want to default `issue_id` to `None`?

SimonFedick (approved):
Looks good! I'm beginning to understand the research tool interface a little more. One quick question regarding the changes to the bug module:

varats/varats/data/reports/szz_report.py
boehmseb: In principle yes, but then we cannot create this with our DataManager, as it relies on a constructor that takes the path. Same goes for all our reports btw.

vulder (approved):

PR #448, opened 2021-03-05
Add Bug Fixing Relation Plot

SimonFedick:
Adds a chord diagram to plots that displays fixing relations between commits, i.e. which commit has fixed a bug caused by which commit.


mypy.ini
```diff
@@ -35,6 +38,9 @@ ignore_missing_imports = True
 [mypy-pygit2.*]
 ignore_missing_imports = True
 
+[mypy-pyperclip.*]
+ignore_missing_imports = True
+
```
vulder:
```suggestion
```
Can be remove

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -0,0 +1,340 @@
...
+    get_project_cls_by_name,
+    get_local_project_git,
+)
+from varats.provider.bug.bug import RawBug
+from varats.provider.bug.bug_provider import BugProvider
+
+LOG = logging.getLogger(__name__)
+
+
+def _plot_chord_diagram_for_raw_bugs(
```
vulder, line 73: Yeah, this function probably needs some splitting up into smaller helper functions, so we can encapsulate the functionality and give the functions a meaningful name. Plotting functions are always a bit tricky here because they invite one to put everything into one function :laughing: happens to me all the time^^

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -0,0 +1,340 @@
...
+        'fixing head': 'rgba(142, 68, 173, 1)',
+        'default': 'rgba(232, 236, 241, 1)'
+    }
+
+    # draw relations and preprocess commit types
+    nodes = []
+    lines = []
+    for bug in bug_set:
+        bug_fix = bug.fixing_commit
+        fix_ind = map_commit_to_id[bug_fix]
```
vulder:
```suggestion
        fix_id = map_commit_to_id[bug_fix]
```
Looks like a typo? Did you mean `fix_id`?

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -0,0 +1,340 @@
...
+        fix_coordinates = commit_coordinates[fix_ind]
+
+        commit_type[bug_fix] = 'introducing fix' if commit_type[
+            bug_fix] == 'introduction' else 'fix'
+
+        for bug_introduction in bug.introducing_commits:
+            intro_ind = map_commit_to_id[bug_introduction]
```
vulder, line 202: same?

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -0,0 +1,340 @@
...
+
+    control_points = [
+        start,
+        np.true_divide(start, (__cp_parameters[interval])),
+        np.true_divide(end, (__cp_parameters[interval])), end
+    ]
+    curve_points = _get_bezier_curve(control_points)
+
+    return gob.Scatter(
+        x=curve_points[:, 0],
```
vulder, line 271: @boehmseb Did you look into the issue of numpy type errors? I think you mentioned something about this.

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -0,0 +1,340 @@
...
+    interval indices are in [0,3] for 5 thresholds."""
+    k = 0
+    while __distance_thresholds[k] < distance:
+        k += 1
+    return k - 1
+
+
+#defining some constants for diagram generation
+__cp_parameters = [1.2, 1.5, 1.8, 2.1]
+__distance_thresholds = [
```
vulder: Please try to write constant names in all caps

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -0,0 +1,340 @@
...
+]
+
+
+def _get_bezier_curve(ctrl_points: np.array, num_points: int = 5) -> np.array:
+    """Implements bezier edges to display between commit nodes."""
+    n = len(ctrl_points)
+
+    def get_coordinate_on_curve(factor: float) -> np.array:
+        points_cp = np.copy(ctrl_points)
+        for r in range(1, n):
```
vulder: What is `r` here? Can we find a better/clearer name?

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -0,0 +1,340 @@
...
+
+        if self.__szz_tool == 'provider':
+            bug_provider = BugProvider.get_provider_for_project(
+                get_project_cls_by_name(project_name)
+            )
+            raw_bugs = bug_provider.find_all_raw_bugs()
+        elif self.__szz_tool == 'szz_unleashed':
+            pass
+        else:
+            raise PlotDataEmpty
```
vulder:
```suggestion
        elif self.__szz_tool != 'szz_unleashed':
            raise PlotDataEmpty
```
Would this also work? Would make it a bit clearer when the expection is raised.

Furthermore, I'm not sure if `PlotDataEmpty` is the correct exception here, maybe we should build a custom one like: `MissingSZZTool`?
@boehmseb Do you have an opinion about the exception/exception name?

vulder (changes requested):
Expect for some minor comments, this looks ok :)

Please also clean up the mypy/pylint warnings.

varats/varats/plots/bug_fixing_relations.py
boehmseb: The type should be `np.ndarray`

varats/varats/plots/bug_fixing_relations.py
boehmseb: This code does make sense as it is. This section retrieves the data depending on the selected `szz_tool`. The validation as you suggest should be already handled in the driver.
The only thing missing here is to replace the `pass` with the actual method to retrieve data from an `SZZReport`.

varats/varats/plots/bug_fixing_relations.py
SimonFedick: I called it `i` now since it is just a loop variable that iterates over the number of control points, I'm not really sure what I should call it otherwise

varats/varats/plots/bug_fixing_relations.py
SimonFedick: This is implemented now, the `szz_tool` can either be `pydriller`, `szz_unleashed` or `szz_diff`, which creates a different plot each.

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -267,20 +423,31 @@ def plot(self, view_mode: bool) -> None:
...
-            raw_bugs = bug_provider.find_all_raw_bugs()
         elif self.__szz_tool == 'szz_unleashed':
-            pass
+            self.__figure = _plot_chord_diagram_for_raw_bugs(
+                project_name, szzunleashed_bugs, self.__szz_tool
+            )
+        elif self.__szz_tool == 'szz_diff':
+            self.__figure = _bug_data_diff_plot(
+                project_name, pydriller_bugs, szzunleashed_bugs
+            )
```
vulder: @boehmseb Does it make sense to consider an enum here? Just passing strings around looks bug introducing to me :)

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -7,28 +9,32 @@
...
     commits for a given set of RawBugs."""
     project_repo = get_local_project_git(project_name)
 
     # maps commit hex -> node id
     map_commit_to_id: tp.Dict[str, int] = _map_commits_to_nodes(project_repo)
-    commit_count = len(map_commit_to_id.keys())
     commit_type: tp.Dict[str, str] = {}
+    commit_count = len(map_commit_to_id.keys())
+
+    edge_colors = ['#d4daff', '#84a9dd', '#5588c8', '#6d8acf']
```
vulder, line 86: Is it fix for this plot that we will only need 4 colors? Otherwise, a color map would be helpful here.

varats/varats/plots/bug_fixing_relations.py
```diff
@@ -41,39 +47,104 @@ def _plot_chord_diagram_for_raw_bugs(
...
+    return gob.Figure(data=data, layout=layout)
+
+
+KeyT = tp.TypeVar("KeyT")
+ValueT = tp.TypeVar("ValueT")
+
+
+def _generate_diff_line_data(
+    diff_raw_bugs: tp.Generator[tp.Tuple[str, tp.Optional[tp.FrozenSet[str]],
+                                         tp.Optional[tp.FrozenSet[str]], str],
```
vulder: Could you move this into a separate type/s so it's easier to understand. I'm not sure what this type represents.

vulder (approved):

varats/varats/plots/bug_fixing_relations.py
boehmseb: I just introduced such an enum in #477. That does not cover the diff case however, but one could use a separate parameter for that.

vulder (approved):

PR #451, opened 2021-04-13
Adds test to check if project names are well formed

cormensratio:
Adds a test to check if the project names are well formed, e.g., they must not contain a dash character.

closes https://github.com/se-sic/VaRA/issues/721

### DIFF
--- a/tests/utils/test_project_util.py
+++ b/tests/utils/test_project_util.py
@@ -1,7 +1,9 @@
 """Test VaRA project utilities."""
 import tempfile
+import typing as tp
 import unittest
 import unittest.mock as mock
+from copy import deepcopy
 from os.path import isdir
 from pathlib import Path
 
@@ -13,6 +15,8 @@ from varats.project.project_util import (
     VaraTestRepoSource,
     VaraTestRepoSubmodule,
 )
+from varats.tools.bb_config import generate_benchbuild_config
+from varats.utils.settings import vara_cfg, bb_cfg
 
 
 class TestVaraTestRepoSource(unittest.TestCase):
@@ -68,6 +72,12 @@ class TestVaraTestRepoSource(unittest.TestCase):
             limit=None,
             shallow=False,
         )
+
+        # Setup and generate the benchbuild config file
+        cls.tmp_file = tempfile.NamedTemporaryFile()
+        generate_benchbuild_config(vara_cfg(), cls.tmp_file.name)
+        cls.bb_cfg = deepcopy(bb_cfg())
+        cls.bb_cfg.load(cls.tmp_file.name)
 
     @mock.patch('benchbuild.source.base.target_prefix')
     @mock.patch('varats.project.project_util.target_prefix')
@@ -219,3 +229,20 @@ class TestVaraTestRepoSource(unittest.TestCase):
                             "1db6fbe",
                             git('rev-parse', '--short', 'HEAD').rstrip()
                         )
+
+    def test_if_project_names_are_well_formed(self) -> None:
+        """Tests if project names are well formed, e.g., they must not contain a
+        dash."""
+
+        loaded_project_paths: tp.List[str] = self.bb_cfg["plugins"]["projects"
+                                                                   ].value
+        loaded_project_names = [
+            project_path.rsplit(sep='.', maxsplit=1)[1]
+            for project_path in loaded_project_paths
+        ]
+        for project_name in loaded_project_names:
+            if '-' in project_name:
+                self.fail(
+                    f"The project name {project_name} must not contain the "
+                    f"dash character."
+                )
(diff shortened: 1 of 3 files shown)


tests/utils/test_project_util.py
```diff
@@ -69,6 +73,12 @@ def setUp(cls) -> None:
             shallow=False,
         )
 
+        # Setup and generate the benchbuild config file
+        cls.tmp_file = tempfile.NamedTemporaryFile()
+        generate_benchbuild_config(vara_cfg(), cls.tmp_file.name)
+        cls.bb_cfg = deepcopy(bb_cfg())
+        cls.bb_cfg.load(cls.tmp_file.name)
```
vulder: @boehmseb Shouldn't we be able to do this automatically with your helper methods?

boehmseb: I'm not sure why the config is loaded/copied here anyway, since the tests only ever read from the config and never modify it in any way. Just read from the real thing, i.e., `bb_cfg()`.

Otherwise, this this would be a job for `replace_config()` from the `test_utils` module. If you only need a copy of the BB config and don't need to replace the actual config, there is also a function `get_bb_test_config` in that same module. @cormensratio, you can look at the other tests in this module to see how this works.

cormensratio: @boehmseb I don't see how `replace_config` can help me there. To use the `replace_bb_config=True` param I already need a benchbuild config, otherwise the tuple of the context manager only contains the replaced vara config. Is this intended behaviour or should I modify the `replace_config()` method (respectively `class _ReplaceConfig`) by adding a call to `bb_cfg()` if `settings._BB_CFG` is None?

As I understand I need a temp file in which I generate the benchbuild config (containing also the vara projects) and a variable of type `Configuration` to load the written benchbuild file from. Instead of the `deepcopy()` I could use `get_bb_test_config()` function but this does internally the same. So I am not sure if I got you right.

But as you already mentioned, this code part should not be in the setUp() method. I will change that when I know how to get the benchbuild config properly.

boehmseb: > @boehmseb I don't see how `replace_config` can help me there. To use the `replace_bb_config=True` param I already need a benchbuild config, otherwise the tuple of the context manager only contains the replaced vara config. 

That's not true. In case no BB config is given, a copy of the current one is created (see lines 69-72).
The context manager should definitely return both configs if used like this:
```python
with replace_config(replace_bb_config=True) as (vara_cfg, bb_cfg):
```
Note that you have to use tuple notation to get python to do the unpacking right.

> As I understand I need a temp file in which I generate the benchbuild config (containing also the vara projects) and a variable of type `Configuration` to load the written benchbuild file from.

You only need a temp file to store/load the config you generate with `generate_benchbuild_config(vara_config, <tmp_file>)`. Note that `vara_config` is the one returned by the context manager. Everything else is handled by `replace_config()`.

>Instead of the `deepcopy()` I could use `get_bb_test_config()` function but this does internally the same. So I am not sure if I got you right.

It still would be better to use `get_bb_test_config()` since that allows the way how the test config is created to change without your test braking. However, with the `replace_config()` context manager, this is not necessary anyway.

cormensratio: > That's not true. In case no BB config is given, a copy of the current one is created (see lines 69-72).
The context manager should definitely return both configs if used like this:

I have to admit that I forgot the tuple notation `(vara_cfg, bb_cfg)`, but even with the correct unpacking, `bb_cfg` is `None`. Looking at line 68 in `test_utils.py` shows `self.old_bb_config = settings._BB_CFG`. But `_BB_CFG` is per default `None` (line 176 in settings.py), only if the function `bb_cfg()` is called it gets assigned the `CFG` as `BB_CFG`. Since bb_cfg() is never called before we start with `None` as value for the `old_bb_config`, which I think you meant with the lines (69-72).
So I think we either have to call `bb_cfg()` in the test before `replace_config()` or we add a `bb_cfg()` call to `replace_config()`.

boehmseb: That looks like a bug. Replacing the `_BB_CFG` in line 68 with `bb_cfg()` should fix it. And if you are at it, you can also replace the `_CFG` in line 81 with `vara_cfg()`.

tests/utils/test_project_util.py
```diff
@@ -219,3 +221,24 @@ def test_vara_test_repo_lib_checkout(
...
+    def test_if_project_names_are_well_formed(self) -> None:
+        """Tests if project names are well formed, e.g., they must not contain a
+        dash."""
+
+        with replace_config(replace_bb_config=True) as (vara_cfg, bb_cfg):
+            tmp_file = tempfile.NamedTemporaryFile()
+            generate_benchbuild_config(vara_cfg, tmp_file.name)
+            bb_cfg.load(tmp_file.name)
+
+        loaded_project_paths: tp.List[str] = bb_cfg["plugins"]["projects"].value
```
boehmseb: At this point, the replaced config is gone again because you left the context manager. You have to match the indent of the context manager here.

boehmseb (changes requested):

vulder (approved):

boehmseb (approved):

PR #452, opened 2021-04-16
Adds blame interaction graphs

boehmseb:
This PR adds a graph abstraction for blame interaction data with different views on the blame data.
Using graphs allows us to analyze blame interactions from a graph-theoretic perspective.


varats/varats/plots/blame_interaction_graph_plots.py
```diff
@@ -0,0 +1,428 @@
...
+
+class CommitInteractionGraphNodeDegreePlot(Plot):
+    """
+    Plot node degrees of a commit interaction graph.
+
+    Additional arguments:
+      - sort: criteria to sort the revisions [degree, time]
+    """
+
+    NAME = 'cig_node_degrees'
```
cormensratio: Regarding the consistent naming of the prior ``NAME``s of the plots, shouldn't this be 'cig_node_degree_plot'? Or do you want to keep the NAME with a maximum of two underscores?

varats/varats/plots/blame_interaction_graph_plots.py
```diff
@@ -0,0 +1,428 @@
...
+
+class CommitInteractionGraphNodeDegreeScatterPlot(Plot):
+    """
+    Plot node degrees of a commit interaction graph.
+
+    Additional arguments:
+      - sort: criteria to sort the revisions [degree, time]
+    """
+
+    NAME = 'cig_node_degrees_scatter'
```
cormensratio: same as above

varats/varats/tables/blame_interaction_graph_table.py
```diff
@@ -0,0 +1,111 @@
...
+    CommitRepoPair,
+    create_commit_lookup_helper,
+    DUMMY_COMMIT,
+)
+
+
+class BlameInteractionGraphMetricsTable(Table):
+    """Visualizes bug metrics of a project."""
+
+    NAME = "cig_metrics_table"
```
cormensratio, line 104: Shouldn't it be called ``big_metrics_table``?

cormensratio (approved):
Besides my comments regarding the naming and after the passing of the pipeline checks, lgtm.

varats-core/varats/project/project_util.py
```diff
@@ -66,10 +66,7 @@ def get_local_project_git_path(
     if is_git_source(source):
         source.fetch()
 
-    return tp.cast(
-        Path,
-        Path(target_prefix()) / source.local.replace(os.sep, '-')
-    )
+    return tp.cast(Path, Path(target_prefix()) / source.local)
```
vulder: Just out of curiosity, why do we need a cast here? Should operator / on a Path not always return a Path?

varats/varats/data/reports/blame_interaction_graph.py
```diff
@@ -163,7 +164,10 @@ def author_interaction_graph(self) -> nx.Graph:
         commit_lookup = create_commit_lookup_helper(self.__project_name)
 
         def partition(u: CommitRepoPair, v: CommitRepoPair) -> bool:
-            return str(commit_lookup(u).author) == str(commit_lookup(v).author)
+            if u.commit_hash == DUMMY_COMMIT or v.commit_hash == DUMMY_COMMIT:
+                return u.commit_hash == v.commit_hash
+            return str(commit_lookup(u).author.name
+                      ) == str(commit_lookup(v).author.name)
```
vulder: I bet you checked this, but is `.name` not a string?

varats/varats/data/reports/blame_interaction_graph.py
```diff
@@ -179,10 +183,14 @@ def edge_data(
             }
 
         def node_data(b: tp.Set[CommitRepoPair]) -> AIGNodeAttrs:
```
vulder: What is `b` here?

varats/varats/data/reports/blame_interaction_graph.py
```diff
@@ -224,7 +231,9 @@ def commit_author_interaction_graph(
         commit_lookup = create_commit_lookup_helper(self.__project_name)
 
         commit_author_mapping = {
-            commit: commit_lookup(commit).author for commit in (list(ig.nodes))
+            commit: commit_lookup(commit).author.name
```
vulder, line 240: Here it seems to work without the extra `str()`

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -16,8 +16,8 @@
 from benchbuild.utils.cmd import mkdir
 from graphviz import Digraph  # type: ignore
 from matplotlib import cm
-from plotly import graph_objs as go
-from plotly import io as pio
+from plotly import graph_objs as go  # type: ignore
+from plotly import io as pio  # type: ignore
```
vulder: Shouldn't this be ignored by the plotly.* ignore in the config?

vulder (changes requested):
Just found a few small things and 2 questions.
Otherwise, this looks fine to me.

vulder:
A one thing I missed, try to update to vara-dev and check how the TC coverage is :wink: 
I like that it works now :laughing:

tests/utils/test_git_util.py
```diff
@@ -80,39 +86,68 @@ def test_get_initial_commit_with_specified_path(self) -> None:
...
+        )
+
+    def test_get_submodule_head(self):
+        """Check if correct submodule commit is retrieved."""
+        repo_path = get_local_project_git_path("grep")
+        old_head = get_head_commit(repo_path)
+        repo_head = FullCommitHash("cb15dfa4b2d7fba0d50e87b49f979c7f996b8ebc")
+        git_checkout(repo_head, repo_path)
+
+        try:
```
vulder, line 131: Why do you need a try here?

tests/utils/test_git_util.py
```diff
@@ -80,39 +86,68 @@ def test_get_initial_commit_with_specified_path(self) -> None:
...
+            git_checkout(old_head, repo_path)
+
+    def test_get_submodule_head_main_repo(self):
+        """Check if correct main repo commit is retrieved."""
+        repo_path = get_local_project_git_path("grep")
+        old_head = get_head_commit(repo_path)
+        repo_head = FullCommitHash("cb15dfa4b2d7fba0d50e87b49f979c7f996b8ebc")
+        git_checkout(repo_head, repo_path)
+
+        try:
```
vulder, line 147: same here

tests/utils/test_git_util.py
```diff
@@ -376,10 +411,10 @@ class TestCodeChurnCalculation(unittest.TestCase):
     def test_one_commit_diff(self):
         """Check if we get the correct code churn for a single commit."""
 
-        repo = get_local_project_git("brotli")
+        repo = get_local_project_git_path("brotli")
```
vulder: above `repo` was rename to `repo_path` because you switched the func, we should keep this consistent

varats-core/varats/project/project_util.py
```diff
@@ -87,10 +87,10 @@ def get_local_project_git_path(
     else:
         source = get_primary_project_source(project_name)
 
-    if is_git_source(source):
-        source.fetch()
+    if not is_git_source(source):
+        raise AssertionError(f"Project {project_name} does not use git.")
 
-    return Path(target_prefix()) / Path(source.local)
+    return Path(source.fetch())
```
vulder, line 94: Fetch does return a Path tot the git? This is a strange API

varats-core/varats/project/project_util.py
```diff
@@ -136,6 +136,28 @@ def get_local_project_git(
...
+
+    Returns:
+        dict with the git repositories for the project's sources
+    """
+    repos: tp.Dict[str, pygit2.Repository] = {}
+    project_cls = get_project_cls_by_name(project_name)
+
+    for source in project_cls.SOURCE:
+        source_name = os.path.basename(source.local)
+        repos[source_name] = get_local_project_git(project_name, source_name)
```
vulder: We should add a check here that the source is actually a git/gitsubmodule source and skip the others.

We'll have a lot of non git sources in the future.

varats-core/varats/utils/git_util.py
```diff
@@ -172,6 +186,19 @@ def get_initial_commit(repo_folder: tp.Optional[Path] = None) -> FullCommitHash:
     )
 
 
+def git_checkout(
```
vulder: There should be a checkout method in the vara_manager.py, which uses the ProcessManager.

varats-core/varats/utils/git_util.py
```diff
@@ -542,6 +485,77 @@ def __hash__(self) -> int:
     def __str__(self) -> str:
         return f"{self.repository_name}[{self.commit_hash}]"
 
+    def __repr__(self) -> str:
+        return self.__str__()
```
```suggestion
        return str(self)
```

varats-core/varats/utils/git_util.py
```diff
@@ -554,7 +568,7 @@ def map_commits(
     """Maps a function over a range of commits."""
     # Skip 0000 hashes that we added to mark uncommitted files
     return [
-        func(commit_lookup(cr_pair.commit_hash, cr_pair.repository_name))
+        func(commit_lookup(cr_pair))
```
vulder, line 558: How does this work, without knowing the corresponding repository?

varats-core/varats/utils/git_util.py
```diff
@@ -672,24 +686,42 @@ def calc_code_churn_range(
...
+        repo_path: path to git repository
+        commit_hash: commit hash to get churn for
         churn_config: churn config to customize churn generation
 
     Returns:
         dict of churn triples, where the commit hash points to
         (files changed, insertions, deletions)
     """
     churn_config = ChurnConfig.init_as_default_if_none(churn_config)
-    return calc_code_churn_range(repo, churn_config, commit, commit)[
```
vulder, line 691: Why do/can we no longer reuse range here?

varats/varats/data/metrics.py
```diff
@@ -59,3 +61,68 @@ def normalized_gini_coefficient(distribution: pd.Series) -> float:
         return gini_coefficient(distribution)
 
     return gini_coefficient(distribution) * (n / (n - 1.0))
+
+
+def apply_tukeys_fence(
```
vulder, line 66: Just checking, was there no tukeys_fence in scipy or some other lib ?

varats/varats/paper_mgmt/case_study.py
```diff
@@ -73,6 +73,30 @@ class ExtenderStrategy(Enum):
     ADD_BUGS = 6
 
 
+def newest_processed_revision_for_case_study(
```
vulder, line 76: Do I understand this correctly: This function retrieves the newest revision (newest commit) that actually has results and, e.g., ignores newer revs that have no results up to now?

varats/varats/plots/blame_interaction_graph_plots.py
```diff
@@ -0,0 +1,532 @@
...
+    cig = create_blame_interaction_graph(project_name,
+                                         revision).commit_interaction_graph()
+
+    def filter_nodes(node: CommitRepoPair) -> bool:
+        if node.commit_hash == UNCOMMITTED_COMMIT_HASH:
+            return False
+        commit = commit_lookup(node)
+        if not commit:
+            return False
+        # TODO: make filter configurable
```
vulder: If we don't need this, we should remove the TODO or?
Maybe just leave a comment with a hint: # Filter could be made configurable

tests/utils/test_git_util.py
boehmseb: Currently, we do not have a good way to isolate git repos in tests (using our test environment does a full clone each time).
Therefore, I need to ensure that I leave the repo in the state it was before the test. Hence the `finally` block with the checkout.

varats-core/varats/project/project_util.py
boehmseb: Yes it does.

varats-core/varats/utils/git_util.py
boehmseb: There is, but it is framed to only be used for branches and needs to be adapted to also be used for commits.

On a side note: `vara_manager.py` probably isn't the best module for these kind of functions anymore.

varats-core/varats/utils/git_util.py
boehmseb: It does know. All I do here is avoid unnecessarily unpacking the commit-repo-pair. The commit lookup function was refactored to do the unpacking by itself.

varats-core/varats/utils/git_util.py
boehmseb: The range implementation does not work for merge commits as it displays diffs for multiple commits in that case but not the one you want. This implementation does not have that problem.
It  could be that you can merge these implementations if you find the right options but I do not know them.

varats/varats/data/metrics.py
boehmseb: Nope, just IQR calculations.

varats/varats/paper_mgmt/case_study.py
boehmseb: Exactly

varats/varats/plots/blame_interaction_graph_plots.py
boehmseb: We didn't really use this plot yet except for some early visualizations, but it will only be useful if the filter is adjustable. However, I would postpone this to when we actually need the plot because then we will have a better understanding of what filters we need, hence, the TODO.

vulder (approved):

tests/table/test_blame_interaction_graph_table.py
```diff
@@ -0,0 +1,130 @@
...
+        # latex booktabs is default format
+        table_str = CommitInteractionGraphMetricsTable().tabulate()
+
+        self.assertEqual(
+            r"""\begin{tabular}{lrrrrrrrrrrrrrr}
+\toprule
+{} & commits & authors & nodes & edges & \multicolumn{4}{c}{node degree} & \multicolumn{3}{c}{node out degree} & \multicolumn{3}{c}{node in degree} \\
+{} &        mean & median & min &  max &          median & min & max &         median & min & max \\
+\midrule
+\textbf{xz} &    1325 &      28 &   124 &   928 &       14.97 &    8.0 &   1 &  154 &             4.0 &   0 &  64 &            3.0 &   0 &  92 \\
```
boehmseb:
```suggestion
\textbf{xz} &    1143 &      28 &   124 &   928 &       14.97 &    8.0 &   1 &  154 &             4.0 &   0 &  64 &            3.0 &   0 &  92 \\
```

tests/table/test_blame_interaction_graph_table.py
```diff
@@ -0,0 +1,130 @@
...
+        # latex booktabs is default format
+        table_str = AuthorInteractionGraphMetricsTable().tabulate()
+
+        self.assertEqual(
+            r"""\begin{tabular}{lrrrrrrrrrrrrrr}
+\toprule
+{} & commits & authors & nodes & edges & \multicolumn{4}{c}{node degree} & \multicolumn{3}{c}{node out degree} & \multicolumn{3}{c}{node in degree} \\
+{} &        mean & median & min & max &          median & min & max &         median & min & max \\
+\midrule
+\textbf{xz} &    1325 &      28 &     1 &     0 &         0.0 &    0.0 &   0 &   0 &             0.0 &   0 &   0 &            0.0 &   0 &   0 \\
```
boehmseb:
```suggestion
\textbf{xz} &    1143 &      28 &     1 &     0 &         0.0 &    0.0 &   0 &   0 &             0.0 &   0 &   0 &            0.0 &   0 &   0 \\
```

tests/table/test_blame_interaction_graph_table.py
```diff
@@ -0,0 +1,130 @@
...
+        # latex booktabs is default format
+        table_str = CommitAuthorInteractionGraphMetricsTable().tabulate()
+
+        self.assertEqual(
+            r"""\begin{tabular}{lrrrrrrrrrrrrrr}
+\toprule
+{} & commits & authors & nodes & edges & \multicolumn{4}{c}{node degree} & \multicolumn{3}{c}{node out degree} & \multicolumn{3}{c}{node in degree} \\
+{} &        mean & median & min & max &          median & min & max &         median & min & max \\
+\midrule
+\textbf{xz} &    1325 &      28 &   125 &    92 &        1.47 &    1.0 &   0 &  92 &             1.0 &   0 &   1 &            0.0 &   0 &  92 \\
```
boehmseb:
```suggestion
\textbf{xz} &    1143 &      28 &   125 &    92 &        1.47 &    1.0 &   0 &  92 &             1.0 &   0 &   1 &            0.0 &   0 &  92 \\
```

varats/varats/tables/blame_interaction_graph_table.py
```diff
@@ -0,0 +1,260 @@
...
+                "node_out_degree": graph.out_degree(node),
+                "node_in_degree": graph.in_degree(node),
+            }))
+
+        data = pd.DataFrame(nodes)
+        degree_data.append(
+            pd.DataFrame.from_dict({
+                project_name: {
+                    ("commits", ""):
+                        int(project_git("rev-list", "--count", "HEAD")),
```
boehmseb:
```suggestion
                        int(project_git("rev-list", "--count", revision.hash)),
```

PR #453, opened 2021-04-23
Rework plot architecture

boehmseb:
This PR reworks how plots can be generated with the tool suite.

## Plot Generators

`PlotGenerator`s are now responsible for creating one or more `Plots`.
They get instantiated with a general options, a plot config (as intended by se-sic/VaRA#545) and Plot(Generator) specific options.
Generators declare these options and the CLI for each generator is automatically generated at load-time.

## CLI

The plot driver discovers all plot generators and constructs CLI a command for each generator from its option declarations.
When executed, the created command instantiates the generator class with the general plot config object and the generator-specific options.
It then invokes the generator's `__call__()` function passing the remaining arguments that are common to all plots (e.g., the 'view' option).
The idea is that subclasses of PlotGenerator only instantiate the concrete plots with their specific options and, in general, should not need the common options or override the `__call__()` function.

As an example, the paper config overview plot is already compatible with the new architecture.

## Other stuff

- From now on, we use `click` as CLI library
- Artefacts have adapted their format to support plot generators
- Specific artefact types (plots/tables) are now managed in modules of their types (e.g., `plots.py` for plot artefacts) 

Resolves se-sic/VaRA#545
Related: se-sic/VaRA#628, se-sic/VaRA#605, se-sic/VaRA#574


varats-core/varats/plot/plots.py
```diff
@@ -180,3 +182,111 @@ def prepare_plots(**args: tp.Any) -> tp.Iterable['varats.plot.plot.Plot']:
...
+    )
+
+    GENERATORS: tp.Dict[str, tp.Type['PlotGenerator']] = {}
+    NAME: str
+    PLOT: tp.Type['varats.plot.plot.Plot']
+    OPTIONS: tp.List[CLIOptionTy]
+
+    def __init__(self, plot_config: PlotConfig, **plot_kwargs: tp.Any):
+        self.__plot_config = plot_config
+        self.__plot_kwargs = plot_kwargs
```
boehmseb: I would like to not store the kwargs here at all. Instead, each concrete generator should know what it needs (a similar thing could be done now for plots, too). The only reason I need this here is so that the constructor signatures match.
Any ideas/opinions on how to do this the best way?

varats-core/varats/plot/plots.py
```diff
@@ -180,3 +182,111 @@ def prepare_plots(**args: tp.Any) -> tp.Iterable['varats.plot.plot.Plot']:
...
+    @property
+    def plot_config(self) -> PlotConfig:
+        """Option with options that influence a plot's appearance."""
+        return self.__plot_config
+
+    @abc.abstractmethod
+    def generate(self) -> tp.List['varats.plot.plot.Plot']:
+        """This function is called to generate the plot instance(s)."""
+
+    def __call__(self, common_options: CommonPlotOptions):
```
boehmseb: The idea is that this function handles everything that is common for all plots, i.e., the parameters for the `vara-plot` command and _not_ its subcommands (aka the concrete plot generators). The reason for this is, that those options should to the same thing every time. Otherwise, they should be plot-specific.

varats-core/varats/utils/cli_util.py
```diff
@@ -80,3 +81,22 @@ def initialize_logger_config() -> None:
     to pass the warning level via an environment variable ``LOG_LEVEL``."""
     log_level = os.environ.get('LOG_LEVEL', "WARNING").upper()
     logging.basicConfig(level=log_level)
+
+
+CLIOptionTy = tp.Callable[..., tp.Any]
```
boehmseb: Unfortunately, click is completely untyped :frowning_face:

varats-core/varats/utils/cli_util.py
```diff
@@ -80,3 +81,22 @@ def initialize_logger_config() -> None:
...
+
+
+# TODO: make this have a typed version of the click.core.Option constructor?
+def make_cli_option(*param_decls: str, **attrs: tp.Any) -> CLIOptionTy:
```
boehmseb: We could use this function to hide the concrete CLI library we are using, or at least provide a typed version. Opinions?

varats/varats/plot/plot.py
```diff
@@ -17,14 +17,63 @@ class PlotDataEmpty(Exception):
...
 
+    PLOTS: tp.Dict[str, tp.Type['Plot']] = {}
+
     def __init__(self, name: str, **kwargs: tp.Any) -> None:
         self.__name = name
         self.__style = "classic"
         self.__saved_extra_args = kwargs
 
+    @classmethod
+    def __init_subclass__(cls, plot_name: str, **kwargs: tp.Any) -> None:
```
cormensratio: I think that the plot discovery currently fails because the signature doesn't match. At least my IDE highlights this ;).

varats/varats/paper_mgmt/paper_config.py
```diff
@@ -91,6 +91,27 @@ def get_all_case_studies(self) -> tp.List[CaseStudy]:
...
+    def get_all_case_study_paths(self) -> tp.List[Path]:
+        """
+        Generate a list of all paths to case studies in the paper config.
+
+        Returns:
+            list of all case study paths in the paper config
+        """
+        filenames: tp.List[str] = self.get_all_case_study_filenames()
+        return [self.__path / filename for filename in filenames]
+
```
boehmseb: We can simply use `PaperConfig.get_all_case_studies()` instead. No need to manually collect those files.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -844,22 +954,48 @@ def _build_graphviz_fig(
...
+    paper_conf: PC.PaperConfig = PC.get_paper_config()
+    if cs_file_name not in paper_conf.get_all_case_study_filenames():
+        all_cs_string = [
+            cs + " " for cs in paper_conf.get_all_case_study_filenames()
+        ]
+        raise FileNotFoundError(
+            f"The selected case study filename could not be found in the "
+            f"current paper config. Did you mean?\n {all_cs_string}"
+        )
+    return PC.load_case_study_from_file(paper_conf.path / cs_file_name)
```
boehmseb: This should be in the `case_study` module. 
Should be obsolete anyway (see my comment in the `paper_config` module.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -1474,7 +1533,126 @@ def calc_missing_revisions(self, boundary_gradient: float) -> tp.Set[str]:
...
+                revision=self.__revision,
+                fig_title=self.__fig_title,
+                width=self.__width,
+                height=self.__height,
+                font_size=self.__font_size,
+            )
+        ]
+
+
+class SankeyLibraryInteractionsGeneratorCS(
```
boehmseb, line 1559: We have multiple options on how to handle the a situation "generate a plot for each case study in my paper config":

- do it as @cormensratio implemented it with multiple different generators
- we could merge this generator and the PC version simply by making the case study optional and selecting the behavior depending on whether a case study is selected
- the PC case is a common pattern that matches many plots and until now is handled by the artefacts mechanism:
  you can currently specify an artefact with an option `paper_config=True` and the artefact toolchain will call the plot (or the generator in the new architecture) once for each case study in the current paper config and no plot(generator) should implement such a paper config behavior. This means that such a behavior can only be achieved using artefacts.

I personally like the last option since it "just works" for all plots that require a (mandatory or optional) case study (this can also be checked with our new architecture). This also means less or less cluttered plot generators and one can argue that if you are at the point that you need to generate the same plot you should use artefacts anyway.

@vulder what's your view on this?

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -50,6 +56,120 @@ class EdgeWeightThreshold(Enum):
...
+    PUBU = 'PuBu'
+    YLGNBU = 'YlGnBu'
+    PUBUGN = 'PuBuGn'
+    BUGN = 'BuGn'
+    YLGN = 'YlGn'
+
+
+# TODO: Remove param default values in plot generation if already set in cli
+
+OPTIONAL_FIG_TITLE: CLIOptionTy = make_cli_option(
```
boehmseb: Some of these options will move to the plot config object at some point, so we should make sure they are easy to migrate.

varats/varats/plots/blame_interaction_degree.py
vulder: Having everything unified with good reuse sounds not to bad, however, I'm not 100% sure about always requiring artefacts. From what I found, listing to users, people already have problems with the paper config/case study model when they start working with varats. For example, they struggle why BB is not executing things or why suddenly so many different revisions are executed, how to get an overview of everything, and how to create/manipulate case studies.
I'm currently a bit torn about this. From one point of view, things could become easier with "How to" guides in the documentation, from another point of view, a lot of our tooling needs a bit of polishing to make things like: "I just want on revision of `foo` as a case study" easier. (Currently, one has to use the case study generator, pic a distribution, set the sample size to 0, and add extra revs, which is the rev one wants... also having to look up the rev in the git.)
Now, we want to further hardly tie artefacts together with plots, which on it self does not have to be a bad thing, but forces users to build up/understand artefacts and artefact managment "just" to get their 3 examples plots.

TDL: In general, I agree that making things reusable is probably the best route design wise, I just want to remark that we need to work and improve on the usability side, which currently is lacking and could get even worse.

boehmseb: I see your point. 
However, solution would not force a user to always use artefacts. Only in the case that a user already has multiple case studies and she wants to generate the same plot for all of them.

I also just had an idea for a fourth possibility:
We currently plan to enhance the case study command line option to properly check whether the user gives a valid case study. We could also add a special value to that option that means "use all case studies". This would force the implementer to handle this case and thus, give the same guarantees as the artefacts solution. This would also simplify the artefacts code a little bit since this case needs no special handling anymore but it would be a breaking change to the artefacts file format (however, this could happen during this PR anyway).

varats/varats/paper_mgmt/paper_config.py
cormensratio: Right, we can gather all case studies of the current PC with `PaperConfig.get_all_case_studies()`, but I thought about a way to let the user specify which case study they want to select for the current plot by choosing a unique name of the case study in the cli. Since the user cannot pass a CaseStudy object to the cli, I used the case study file name as identifier. Furthermore, we could use this to show the user a list of available case study files, if an incorrect one is selected. Even better would be a `click.Choice` that already provides available case studies, but this is currently not possible because of a cyclic dependency that we already discussed last time.
What other unique 'key' of a case study besides its file name do we have that a user can pass to the cli in order to specify the case study?

boehmseb: The only unique identifier is name + version. Maybe I manage to finish my Choice implementation of this in the evening, than this problem would be gone for good.

vulder:
related se-sic/VaRA#628

varats/varats/plots/bug_annotation.py
```diff
@@ -22,21 +20,8 @@ def draw_bugs(
         project: the project to add bugs for
         revisions: a list of revisions included in the plot in the order they
                    appear on the x-axis
-        extra_plot_cfg:
-            * linewidth
-            * color
-            * vertical_alignment
-            * label_size
+        plot_kwargs: the arguments that specify a plots style
```
boehmseb: I don't like the idea of relying on kwargs here since this function could be called from any  plot.
Since it always takes the same couple of arguments, just make them explicit.
I also suggest to pass a `PlotConfig` object as an additional argument since some of the required arguments will definitely be migrated to the plot config.

Same goes for `draw_cves()` btw.

varats/varats/plots/cve_annotation.py
```diff
@@ -22,21 +20,8 @@ def draw_cves(
         project: the project to add CVEs for
         revisions: a list of revisions included in the plot in the order they
                    appear on the x-axis
-        extra_plot_cfg:
-            * linewidth
-            * color
-            * vertical_alignment
-            * label_size
+        plot_kwargs: the arguments that specify a plots style
```
boehmseb: See comment on `draw_bugs()`

varats/varats/ts_utils/cli_util.py
```diff
@@ -84,19 +86,114 @@ def initialize_logger_config() -> None:
...
+    name = "typed multi choice"
+
+    def __init__(
+        self,
+        choices: tp.Dict[str, tp.List[ChoiceTy]],
+        case_sensitive: bool = True
+    ):
+        # Relates to: https://thingspython.wordpress.com/2010/09/27/another
+        # -super-wrinkle-raising-typeerror/
+        self.as_super = super(TypedMultiChoice, self)
```
boehmseb: I'm not sure if that is really what happens. The linked article is for Python 2.4/2.6.

Could you give me a MWE to reproduce the issue so I can confirm this?

varats/varats/plots/paper_config_overview.py
```diff
@@ -286,6 +289,8 @@ def calc_missing_revisions(
     ) -> tp.Set[FullCommitHash]:
         revisions = _gen_overview_plot_for_project(**self.plot_kwargs)
         revisions.sort_values(by=['revision'], inplace=True)
+
+        # TODO: Adapt to click cli.
```
cormensratio: Which `case_study`(ies) do we use to get the correct `commit_map`s ?

varats/varats/plot/plots.py
```diff
@@ -0,0 +1,559 @@
...
+        "-rev",
+        "--revision",
+        type=str,
+        required=True,
+        metavar="revision",
+        help="The revision to use for the plot."
+    )
+    REQUIRE_REPORT_TYPE: CLIOptionTy = make_cli_option(
+        "--report-type",
+        type=click.Choice([
```
cormensratio: Should we directly convert the report `SHORTHAND` strings to their corresponding `report_type` when saving them in the generators as attributes? 
E.g.:
`self.__report_type: BlameReport = BaseReport.lookup_report_type_by_shorthand(plot_kwargs["report_type"])`

docs/source/tutorials/visualizing_data.rst
```diff
@@ -0,0 +1,63 @@
...
+    In the following, we will only write about plots since plots and tables work essentially in the same way.
+    For tables, just replace *plot* with *table*.
+
+You can check what plots are available by running ``vara-plot -h``.
+The available plot names are listed under the section ``Commands`` in the help output.
+
+.. program-output:: vara-plot -h
+    :nostderr:
+
+To see what plot-specific a plot has, you need to look at the help output for that specific plot, e.g.:
```
cormensratio:
```suggestion
To see what plot-specific options a plot has, you need to look at the help output for that specific plot, e.g.:
```

docs/source/tutorials/visualizing_data.rst
```diff
@@ -0,0 +1,63 @@
...
+
+VaRA-TS allows you to visualize the data generated by experiments in various plots and tables.
+This guide explains how plots and tables can be generated via our command line tools.
+If you want to write your own plots check :ref:`here <Plots>`.
+
+
+Generating Plots and Tables
+---------------------------
+
+Plots and tables can be generated with the tools :ref:`vara-plot` and :ref:`vara-table`.
```
cormensratio, line 12: I think that the image from `:ref: here <Plots>` under `Plot Architecture` would be (also) relevant to see on this page, since it gives a nice overview of how we can call a plot and is not only relevant `If you want to write your own plots ...`.

docs/source/tutorials/visualizing_data.rst
```diff
@@ -0,0 +1,63 @@
...
+.. program-output:: vara-plot -h
+    :nostderr:
+
+To see what plot-specific a plot has, you need to look at the help output for that specific plot, e.g.:
+
+.. program-output:: vara-plot pc-overview-plot -h
+    :nostderr:
+
+The plot files are stored in ``$VARATS_ROOT/plots``.
+You can group plot files into subdirectories using the option ``--plot-dir``, e.g. ``vara-plot --plot-dir=foo pc-overview-plot --report-type=EmptyReport`` will create a file in the directory ``$VARATS_ROOT/plots/foo``.
```
cormensratio, line 39: Has to be changed if we decide to use the reports' `SHORTHAND` as click.Choices.
I don't think we need the `=`?
```suggestion
You can group plot files into subdirectories using the option ``--plot-dir``, e.g. ``vara-plot --plot-dir foo pc-overview-plot --report-type EMPTY`` will create a file in the directory ``$VARATS_ROOT/plots/foo``.
```

docs/source/tutorials/visualizing_data.rst
```diff
@@ -0,0 +1,63 @@
...
+
+:ref:`Artefacts` are a way to persist plot and table configurations (i.e., ``vara-plot`` and ``vara-table`` calls) so that they can be easily (re-) generated.
+The artefact specifications are a part of the paper config and therefore, can be easily shared and are useful to reproduce results.
+
+To store a plot or table configuration you simply need to provide a name for the artefact via the option ``--save-artefact``.
+The ``vara-plot`` and ``vara-table`` tools will then store the configuration in the current paper config instead of generating the plot or table.
+For example, an artefact that will generate a ``pc-overview-plot`` for the current paper config can be added
+via::
+
+    vara-plot --save-artefact="PC Overview" pc-overview-plot --report-type=EmptyReport
```
cormensratio, line 54: s.o.
```suggestion
    vara-plot --save-artefact "PC Overview" pc-overview-plot --report-type EMPTY
```

docs/source/vara-ts-api/plot_architecture.svg
```diff
@@ -0,0 +1,492 @@
...
+       xml:space="preserve"
+       style="font-style:normal;font-weight:normal;font-size:3.52778px;line-height:1.25;font-family:sans-serif;letter-spacing:0px;word-spacing:0px;fill:#000000;fill-opacity:1;stroke:none;stroke-width:0.264583"
+       x="2.6458333"
+       y="5.2916665"
+       id="text850"><tspan
+         sodipodi:role="line"
+         id="tspan848"
+         x="2.6458333"
+         y="5.2916665"
+         style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:3.52778px;font-family:'DejaVu Sans Mono';-inkscape-font-specification:'DejaVu Sans Mono';stroke-width:0.264583">vara-plot --plot-dir=foo --title=Overview pc-overview-plot --report-type=EmptyReport</tspan></text>
```
cormensratio, line 267: If the colors do not decrease the text's readability, maybe color the parts of the command line string according to the underlying parentheses' color?

docs/source/vara-ts-api/plot_architecture.svg
```diff
@@ -0,0 +1,492 @@
...
+       xml:space="preserve"
+       style="font-style:normal;font-weight:normal;font-size:3.52778px;line-height:1.25;font-family:sans-serif;letter-spacing:0px;word-spacing:0px;fill:#000000;fill-opacity:1;stroke:none;stroke-width:0.264583"
+       x="134.9375"
+       y="58.208336"
+       id="text926-2-9-8"><tspan
+         sodipodi:role="line"
+         id="tspan924-47-3-9"
+         x="134.9375"
+         y="58.208336"
+         style="font-size:3.52778px;fill:#000000;fill-opacity:1;stroke-width:0.264583">Plot</tspan></text>
```
cormensratio, line 409: Maybe this is too captious, but are the three dots under the last Plot box horizontally centered? ;)

docs/source/vara-ts-api/plots.rst
```diff
@@ -3,50 +3,81 @@ Plots
...
+
+The command line interface for ``vara-plot`` consists of four parts:
+
+1. *Common options:* handle functionality common to all plots, like file format
+2. *Plot config:* for tuning visual appearance; specific plots may or may not respect them
+3. *Plot generator name:* name of the concrete plot generator
+4. *Plot- or table-specific options*: options specific to a certain plot
+
+Plot generation is handeled by the :class:`~varats.plot.plots.PlotGenerator` classes.
+The CLI selects a concrete plot generater via the passed **plot generator name** and immeadiately calls its ``__call__()`` function (which is implemented by the :class:`~varats.plot.plots.PlotGenerator` itself and must not be overridden by subclasses).
```
cormensratio:
```suggestion
The CLI selects a concrete plot generator via the passed **plot generator name** and immediately calls its ``__call__()`` function (which is implemented by the :class:`~varats.plot.plots.PlotGenerator` itself and must not be overridden by subclasses).
```

docs/source/tutorials/visualizing_data.rst
cormensratio, line 52: Do we need the `=`?
```suggestion
    vara-plot --save-artefact "PC Overview" pc-overview-plot --report-type=EmptyReport
```

docs/source/tutorials/visualizing_data.rst
boehmseb: The `=` is optional but typically used with long-style options.
As for the reports, I'm not sure whether it is good to use the shorthands, since they are less self-documenting than the full report names. @vulder wha's your opinion on this?

docs/source/vara-ts-api/plot_architecture.svg
boehmseb: I mainly didn't do it because inkscape didn't let me use different colors in the same text box and I was too lazy to split it into multiple text boxes so that they still align properly.

docs/source/tutorials/visualizing_data.rst
vulder: I agree, the `SHORTHANDS` are often not very readable (e.g., `BlameReport` vs `BR`). If people want it, we would add a lookup mechanic to allow both, but from my point of view, the default style we advertise should be the report name.

varats/varats/plot/plots.py
boehmseb: I changed that to use the full report name. I think that's better readable for users.

docs/source/vara-ts-api/plot_architecture.svg
boehmseb: I tried this but in my opinion, the black text is better readable.

boehmseb:
Ready for a first round of review.
@vulder: focus on `driver_plot.py`, `plots.py` and `plot.py`.
There is also some [documentation](https://vara.readthedocs.io/en/f-plotrework/tutorials/visualizing_data.html) available.
For a simple example plot take a look at `paper_config_overview.py`.

varats/varats/plot/plots.py
```diff
@@ -0,0 +1,686 @@
...
+            )
+        else:
+            return make_cli_option(
+                f"--{self.__name}",
+                type=self.__type,
+                required=False,
+                help=self.__help
+            )
+
+    def value_or_default(self, default: tp.Optional[OptionType]) -> OptionType:
```
cormensratio:
```suggestion
    def value_or_default(self, default: tp.Optional[OptionType] = None) -> OptionType:
```

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -929,42 +1047,23 @@ def _get_degree_data(self) -> pd.DataFrame:
...
-                       f'Project {self.plot_kwargs["project"]}'
-        plot_cfg["fig_suptitle"] = fig_suptitle
+        project_name = self.plot_kwargs['case_study'].project_name
+        fig_suptitle = f'{str(self.plot_kwargs["fig_title"])} - ' \
+                       f'Project {project_name}'
+        self.plot_kwargs["fig_suptitle"] = fig_suptitle
 
-        style.use(self.style)
-        commit_map: CommitMap = self.plot_kwargs['get_cmap']()
+        style.use(self.plot_config.style())
```
cormensratio, line 1089: Calling `self.plot_config.style()` returns the unset `value`, which is `None` instead of the wanted `default`.

boehmseb: How so?

boehmseb: Nevermind. That happens when you forget a return :man_facepalming:

cormensratio: :D

varats/varats/plot/plots.py
```diff
@@ -0,0 +1,690 @@
...
+                    help_str="The title of the legend."
+                ),
+                PlotConfigOption(
+                    "legend_size",
+                    default=2,
+                    help_str="The size of the legend."
+                ),
+                PlotConfigOption(
+                    "show_legend",
+                    default=False,
```
cormensratio, line 334: Maybe change this to True?
```suggestion
                    default=True,
```

cormensratio (approved):
Looks good. ~~Maybe consider setting the `show_legend` option `default` to `True`.~~

varats/varats/plot/plots.py
boehmseb: Hmm, I'd say that it is better if a plot needs to opt in to having a legend (i.e., by using `PlotConfig.show_legend(True)`)

cormensratio: Right, this would be better.

boehmseb:
mypy complains because of numpy. It doesn't do so locally despite having the same mypy version. I would ignore this for this PR.

docs/source/tutorials/visualizing_data.rst
```diff
@@ -0,0 +1,65 @@
...
+.. figure:: ../vara-ts-api/plot_architecture.svg
+
+1. *Common options:* handle functionality common to all plots or tables, like file format
+2. *Plot or table config:* for tuning visual appearance; specific plots/tables may or may not respect them
+3. *Plot or table name:* name of the concrete plot or table
+4. *Plot- or table-specific options*: options specific to a certain plot or table
+
+.. note::
+
+    In the following, we will only write about plots since plots and tables work essentially in the same way.
```
vulder:
```suggestion
    In the following, we only write about plots since plots and tables work essentially in the same way.
```

docs/source/vara-ts-api/plots.rst
```diff
@@ -3,50 +3,81 @@ Plots
...
+
+The command line interface for ``vara-plot`` consists of four parts:
+
+1. *Common options:* handle functionality common to all plots, like file format
+2. *Plot config:* for tuning visual appearance; specific plots may or may not respect them
+3. *Plot generator name:* name of the concrete plot generator
+4. *Plot- or table-specific options*: options specific to a certain plot
+
+Plot generation is handeled by the :class:`~varats.plot.plots.PlotGenerator` classes.
+The CLI selects a concrete plot generator via the passed **plot generator name** and immediately calls its ``__call__()`` function (which is implemented by the :class:`~varats.plot.plots.PlotGenerator` itself and must not be overridden by subclasses).
```
vulder, line 26: Should we decorate the method with `@final`?

vulder (approved):
Looks good to me.
The only thing I did not fully understand, regarding the configs, do we automatically provide different defaults depending on whether we generate a plot or present it (view) to the user?

docs/source/vara-ts-api/plots.rst
boehmseb: I wasn't aware that such a decorator exists. And yes, for now, we could definitely add that decorator.

I can imagine that in the future there maybe might be a case where we would need to override this function (e.g., for some obscure plotting framework), but if so, we can still remove the annotation.

vulder (approved):

PR #463, opened 2021-05-13
Adds test_suite project

J03D03:
### Description

This PR adds the [Test-Suite](https://github.com/SVF-tools/Test-Suite) project "a micro-benchmark suite designed for validating various static analysis algorithms".


varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,81 @@
...
+    GROUP = 'test_projects'
+    DOMAIN = 'testing'
+
+    SOURCE = [
+        bb.source.Git(
+            remote="https://github.com/SVF-tools/Test-Suite.git",
+            local="svf-test-suite",
+            refspec="HEAD",
+            limit=1,  # None,
+            shallow=False,
```
vulder, line 33: If you only need the last commit you can enable shallow here.

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,81 @@
...
+    @property
+    def binaries(self) -> tp.List[ProjectBinaryWrapper]:
+        """Return a list of binaries generated by the project."""
+        source = os.getcwd() + '/tmp/' + self.source[0].local
+        for directory in self.DIRS:
+            for filename in os.listdir(source + '/' + directory):
+                absolute_filepath = source + '/' + directory + '/' + filename
+                ext = os.path.splitext(absolute_filepath)[1]
+                if('.cpp' == ext or '.c' == ext):
+                    self.LOCAL_PATHS.append(directory + '/' + filename)
```
vulder: I don't think you should change the class variable in this method. This would add new PATHS to the variable every time someone wants to look up which binaries are generated by this project.

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,81 @@
...
+        pass
+
+    def compile(self) -> None:
+        source = local.path(self.source_of_primary)
+        with local.cwd(source):
+            for file in self.LOCAL_PATHS:
+                clang = bb.compiler.cxx(self)
+                if('.c' == os.path.splitext(file)[1]):
+                    clang = bb.compiler.cc(self)
+                with local.env(CC=str(clang)):
```
vulder: If you also have cpp files you need to also override CXX with something like this `CXX=str(clang)`

Also, if you don't use a build system, meaning you call clang you self for the compilation you don't need to set the env vars at all.

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,81 @@
...
+        return wrap_paths_to_binaries([(file_name.replace('.cpp', '').replace('.c', ''), BinaryType.executable) for file_name in self.LOCAL_PATHS])
+
+    def run_tests(self) -> None:
+        pass
+
+    def compile(self) -> None:
+        source = local.path(self.source_of_primary)
+        with local.cwd(source):
+            for file in self.LOCAL_PATHS:
+                clang = bb.compiler.cxx(self)
```
vulder: Just to point this out, you are also compiling your `c` files with clang++ in this case because you don't differentiate between `bb.compiler.cxx` and `bb.compiler.cc`

vulder (changes requested):

varats/varats/projects/test_projects/test_suite.py
J03D03:
```
==== ERROR ====
Execution of a binary failed in step: * Project environment for: test_suite @ (Variant(version='ab0923aabe'),)
Unexpected exit code: 128
Command line: | /usr/bin/git clone --dissociate --recurse-submodules --reference /home/user/Git/PhASAR/VARA/vara-root/benchbuild/tmp/svf-test-suite https://github.com/SVF-tools/Test-Suite.git .
Stderr:       | Cloning into '.'...
              | fatal: reference repository '/home/user/Git/PhASAR/VARA/vara-root/benchbuild/tmp/svf-test-suite' is shallow
==== ERROR ====
```

Seems like shallow should stay disabled.

varats/varats/projects/test_projects/test_suite.py
J03D03: Agreed. It was moved into an `init_file_paths` function which is called on `__init__`.

varats/varats/projects/test_projects/test_suite.py
J03D03: Missed that. Thank you!

Omitted the env part.

varats/varats/projects/test_projects/test_suite.py
J03D03: Missed that. Thanks

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+class MicroBenchmark(bb.Project):  # type: ignore
+    """
+    Get examples from the Test-Suite micro-benchmark suite.
+
+    Which can be found at https://github.com/SVF-tools/Test-Suite.
+    """
+
+    NAME = 'test_suite'
+    GROUP = 'test_projects'
+    DOMAIN = 'testing'
```
vulder: We could pick better names for the class and the project. How does one of these sound like?

- Class: SVFPointsToAnalysisBenchmark | Name: SVFPointsToBench
- Class: SVFMemoryAnalysisBenchmark | Name: SVFMemBench
- Class: ??? | Name: ???

any other ideas?

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+                    file.replace('.cpp', '').replace('.c', '')
+                ]
+                if os.path.splitext(file)[1] == '.c':
+                    bb.watch(bb.compiler.cc(self))(arguments)
+                else:
+                    bb.watch(bb.compiler.cxx(self))(arguments)
+
+    def __init__(self, var) -> None:
+        super().__init__()
+        self.init_file_paths()
```
vulder: Please move the `__init__` method to the top of the class so it's the first method (just a convention we have)

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+        # "mta",
+        "non_annotated_tests",
+        "path_tests",
+    ]
+
+    FILE_PATHS = []  # type: tp.List[str]
+
+    def init_file_paths(self) -> None:
+        """Initialize FILE_PATHS with .c and .cpp files found in DIRS."""
+        source = os.getcwd() + '/tmp/' + self.source[0].local
```
vulder: hmm, I'm not sure where this path points to in the end. Try to use `self.source_of_primary` to access the current primaries source location (the primary source is the first one specified in `SOURCE`). Take a look at gravity.py s `compile_cmake` method as an example.

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+        for directory in self.DIRS:
+            for filename in os.listdir(source + '/' + directory):
+                absolute_filepath = source + '/' + directory + '/' + filename
+                ext = os.path.splitext(absolute_filepath)[1]
+                if ext in ('.cpp', '.c'):
+                    self.FILE_PATHS.append(directory + '/' + filename)
```
vulder: After changing to the source access function you'll get a python pathlib Path back, with this type you can simplify parts of this code here, e.g.,
```
p = Path('some/location/foo')
p / "subfolder"
```
Will automatically do the correct path extension, picking the correct platform specific separator etc.

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def run_tests(self) -> None:
+        pass
+
+    def compile(self) -> None:
+        source = local.path(self.source_of_primary)
+        with local.cwd(source):
+            for file in self.FILE_PATHS:
+                arguments = [
+                    "{source}/{file}".format(source=source, file=file),
```
vulder:
```suggestion
                    f"{source}/{file}",
```
Try to use python f-strings, they are a very handy way to insert data into a string.
```
f"Some text {code} other text"
```
With f-strings you can just insert curly braces in the string and inside the braces write code, e.g., access variables and even write more complex statements like this:
```
f"Trying to execute {command if command else 'empty command'}, may the emperor protect us."
```

varats/varats/projects/test_projects/test_suite.py
vulder: Create a `pathib.Path` here, makes it easier to work with later on.

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+        # "fstbhc_tests",
+        "path_tests",
+        "graphtxt",
+        # "mem_leak",
+        # "mta",
+        "non_annotated_tests",
+        "path_tests",
+    ]
+
+    FILE_PATHS = []  # type: tp.List[str]
```
vulder:
```suggestion
    FILE_PATHS: tp.List[Path] = []
```
easier to work with later on

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
+"""Project to run tests from the micro-benchmark Test-Suite."""
+import os
+import typing as tp
```
vulder, line 2:
```suggestion
import typing as tp
from pathlib import Path
```

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+        with local.cwd(source):
+            for file in self.FILE_PATHS:
+                arguments = [
+                    "{source}/{file}".format(source=source, file=file),
+                    "-I{source}".format(source=source),
+                    "-g",  # Generate source-level debug information
+                    "-o",
+                    file.replace('.cpp', '').replace('.c', '')
+                ]
+                if os.path.splitext(file)[1] == '.c':
```
vulder:
```suggestion
                if file.suffix == 'c':
```
After switching to path, is becomes easier to write and more readable :smiley:

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+                ext = os.path.splitext(absolute_filepath)[1]
+                if ext in ('.cpp', '.c'):
+                    self.FILE_PATHS.append(directory + '/' + filename)
+
+    @property
+    def binaries(self) -> tp.List[ProjectBinaryWrapper]:
+        """Return a list of binaries generated by the project."""
+        return wrap_paths_to_binaries([(
+            file_name.replace('.cpp', '').replace('.c',
+                                                  ''), BinaryType.executable
```
vulder:
```suggestion
            file_name.with_suffix(''), BinaryType.executable
```

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+    def compile(self) -> None:
+        source = local.path(self.source_of_primary)
+        with local.cwd(source):
+            for file in self.FILE_PATHS:
+                arguments = [
+                    "{source}/{file}".format(source=source, file=file),
+                    "-I{source}".format(source=source),
+                    "-g",  # Generate source-level debug information
+                    "-o",
+                    file.replace('.cpp', '').replace('.c', '')
```
vulder:
```suggestion
                    file.with_suffix('')
```

varats/varats/projects/test_projects/test_suite.py
```diff
@@ -0,0 +1,93 @@
...
+from varats.project.project_util import (
+    ProjectBinaryWrapper,
+    wrap_paths_to_binaries,
+    BinaryType,
+)
+
+
+class MicroBenchmark(bb.Project):  # type: ignore
+    """
+    Get examples from the Test-Suite micro-benchmark suite.
```
vulder:
```suggestion
    (InsertProjectName) provides and easy way to execute the examples from the Test-Suite micro-benchmark suite for testing points to analyses.
```

vulder (changes requested):

varats/varats/projects/test_projects/test_suite.py
J03D03: SVFPointsToAnalysisBenchmark sounds good to me

varats/varats/projects/test_projects/test_suite.py
J03D03: The description has been updated

varats/varats/projects/test_projects/test_suite.py
J03D03: Sure! Done

varats/varats/projects/test_projects/test_suite.py
J03D03: Done

varats/varats/projects/test_projects/test_suite.py
J03D03: True that :)

varats/varats/projects/test_projects/test_suite.py
J03D03: Hi @vulder, I have a little problem with your suggestion.
I can't use `self.source_of_primary` in the `__init__` method since it isn't set yet.
Since the `binaries` method is called before `compile` i can't call `init_file_paths` from there :thinking:

vulder: hmm, then we need to specify the paths manually or do you have other ideas?
The problem with `os.getcwd()` is it depends on the exec location/current location, which is not portable to other systems, machines or containers.
I currently see two options here:
1) find a way to safely checkout the repo before have/find the location with the repository
2) manually add the file paths to the list

J03D03: I've no better idea, that's why I picked option two and added the file paths manually with https://github.com/se-sic/VaRA-Tool-Suite/pull/463/commits/c81be36bddc593109307261c5be73d27d9052101.

vulder (approved):

PR #475, opened 2021-06-06
Fixes Issue Bug Datetime Format

SimonFedick:
Accounts for the fact that pydriller commits and GitHub issues use different datetime formats and converts both to UTC format before comparing them. This caused an issue during the suspect filtering of issue event bugs.


boehmseb:
Is there a specific reason why you need an external time library? Python's `datetime` should be able to handle timezones.

vulder (changes requested):
I think @boehmseb is right, we should try to use the standard lib as much as possible and not draw in new libs for everything.

vulder (approved):

PR #477, opened 2021-06-10
SZZ quality metric database + table

boehmseb:


varats/varats/data/reports/szz_report.py
```diff
@@ -9,6 +10,17 @@
 from varats.report.report import BaseReport, FileStatusExtension, MetaReport
 
 
+class SZZTool(Enum):
+    """Enum for available SZZ tools."""
+
+    def __init__(self, tool_name: str, shorthand: str):
+        self.tool_name = tool_name
+        self.shorthand = shorthand
+
```
vulder, line 27: You need to add properties here or use a type decl for the enum, so pylint understands it.

varats/varats/tables/szz_quality_metrics_table.py
```diff
@@ -0,0 +1,65 @@
...
+                project_name, columns, commit_map
+            )
+        else:
+            raise ValueError(f"Unknown SZZ tool '{szz_tool_name}'")
+
+        data.rename(
+            columns={
+                "revision": "fix",
+                "introducer": "introducer",
+                "score": "score"
```
vulder: You could define this mapping above where you specify the columns and then use `.keys()` when calling `get_data_for_project`. This would ensure that the column values are always in sync, when someone wants to change them later on.

vulder (approved):
The enum needs a bit of typing love, otherwise, this looks fine to me.

varats/varats/tables/szz_quality_metrics_table.py
vulder: after this we can merge the PR

vulder (changes requested):

vulder (approved):
Did you add szz etc. to the docs? Could be useful for others to have more examples.

vulder (approved):

PR #480, opened 2021-06-22
Remove duplicate `save_bb_config()` and fix issue with custom BB config options

boehmseb:
One open issue left:
We cannot import stuff from any other varats modules in `settings` as we would always have some cyclic imports. That's why some BB settings stuff is in `settings.py`. If we want to separate this better, everything BB settings related must move to `bb_config.py` including `bb_cfg()`. That however, would lead to quite a big change because every import of  `bb_cfg` would change. Technically not a problem, but a large diff.


vulder (approved):
Refactoring looks fine to me :+1:

vulder (approved):

PR #482, opened 2021-06-22
Fixes wrong integer comparison

vulder:


cormensratio:
Thanks for pointing that out, I'll keep in mind to use the `==` operator to check for value/object equality.

cormensratio (approved):

PR #484, opened 2021-06-25
FTA experiment added

Keller-1998:
added experiment for FTA


varats/varats/experiments/vara/phasar_fta.py
```diff
@@ -0,0 +1,158 @@
...
+        # Add own error handler to compile step.
+        project.compile = get_default_compile_error_wrapped(
+            project, self.REPORT_TYPE, PhASARFTACheck.RESULT_FOLDER_TEMPLATE
+        )
+
+        project.cflags += [
+            "-O1", "-Xclang", "-disable-llvm-optzns", "-fvara-IFA"
+        ]
+
+        bc_file_extensions = [BCFileExtensions.NO_OPT, BCFileExtensions.TBAA]
```
vulder: We need a new BC file extension that marks BC files with feature information in them. Just create a new entry in the enum `FEATURE`.

varats/varats/experiments/vara/phasar_fta.py
```diff
@@ -0,0 +1,158 @@
...
+            << RunWLLVM() \
+            << run.WithTimeout()
+
+        # Add own error handler to compile step.
+        project.compile = get_default_compile_error_wrapped(
+            project, self.REPORT_TYPE, PhASARFTACheck.RESULT_FOLDER_TEMPLATE
+        )
+
+        project.cflags += [
+            "-O1", "-Xclang", "-disable-llvm-optzns", "-fvara-IFA"
```
vulder: For the analysis there is now an easier way, you can just use the "-fvara-feature" flag. I refactored out the feature part a few weeks ago so we don't need to generate the if annotations.

varats/varats/experiments/vara/phasar_fta.py
```diff
@@ -0,0 +1,158 @@
...
+        project = self.obj
+
+        # Define the output directory.
+        vara_result_folder = self.RESULT_FOLDER_TEMPLATE.format(
+            result_dir=str(bb_cfg()["varats"]["outfile"]),
+            project_dir=str(project.name)
+        )
+        mkdir("-p", vara_result_folder)
+
+        timeout_duration = '24h'
```
vulder: I would just remove the timeout, it should no longer be needed, as our analysis run on the cluster with a c group timeout anyhow.

varats/varats/experiments/vara/phasar_fta.py
```diff
@@ -0,0 +1,158 @@
...
+            ]
+
+            run_cmd = opt[opt_params]
+
+            run_cmd = wrap_unlimit_stack_size(run_cmd)
+
+            # Run the command with custom error handler and timeout
+            exec_func_with_pe_error_handler(
+                timeout[timeout_duration, run_cmd] >
+                f"{vara_result_folder}/{result_file}",
```
vulder: I would move the `>` operator to create the final run_cmd before passing it to the exec function, makes it easier readable.
```
run_cmd = run_cmd > f"foo/bar"

exec_func(
    run_cmd, ...
)

File: varats/varats/experiments/vara/phasar_fta.py [hunk 5]
```diff
@@ -0,0 +1,158 @@
...
+            result_dir=str(bb_cfg()["varats"]["outfile"]),
+            project_dir=str(project.name)
+        )
+        mkdir("-p", vara_result_folder)
+
+        timeout_duration = '24h'
+
+        for binary in project.binaries:
+            # Define empty success file
+            result_file = EMPTY.get_file_name(
```
vulder: I know the GenerateBlameReport experiment is not a good example for this but I would move report and the BC_file_extensions list into the class and pass it via __init__ to the Step, that ensures that there is only one location where they are defined. So when someone changes the report type below, you automatically use the right one here.


```
    __init__(self, project, report_type,... ):
       self.__report_type  = report_type


... later on ....
       result_file = self.__report_type.get_file_name(...)
```

If you like you can also define report_type as a property

varats/varats/experiments/vara/phasar_fta.py
```diff
@@ -0,0 +1,158 @@
...
+        self,
+        project: Project,
+    ):
+        super().__init__(obj=project, action_fn=self.analyze)
+
+    def analyze(self) -> actions.StepResult:
+        """This step performs the actual analysis with the correct flags."""
+
+        if not self.obj:
+            return
```
vulder:
```suggestion
```
This is no longer needed as we make sure to always pass a project. This pattern was necessary before BB make sure to always inject correct projects.

vulder (changes requested):
Looks quite good for you first experiment :+1: 
I only jound a few things how we could improve things.

vulder (approved):

PR #485, opened 2021-06-25
Fixes broken docu generation due to numpy typing issue

vulder:


docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
+numpy<=1.20
```
boehmseb: Maybe add a note why we pinned it and when we can unpin it again.

boehmseb (approved):

PR #486, opened 2021-06-30
Extends the BlameReport meta data with phasar specific infos

vulder:


boehmseb (approved):
I assume you made the new meta data optional since you didn't want to break the current BR file format version and you are not sure whether in the future, all blame reports will be generated with this meta data?

vulder:
> I assume you made the new meta data optional since you didn't want to break the current BR file format version and you are not sure whether in the future, all blame reports will be generated with this meta data?

Yes, with optional we don't have to break the version and most importantly the "just vara" version does not support these two values. After removing this at some point we can make them non optional.

PR #502, opened 2021-07-19
Introduces type-safe commit hash abstractions.

boehmseb:
This PR introduces classes that encapsulate commit hashes in a type-safe way:

`CommitHash`: base class for more specific types
`ShortCommitHash`: abbreviated commit hash (10 characters)
`FullCommitHash`: full-length commit hash
`CommitHashTy`: type variable that is bound to `CommitHash`

All commit hash classes are constructible from strings and pygit2 commit objects. In addition, `FullCommitHash` can be converted to `ShortCommitHash` (but not the other way round).


varats-core/varats/utils/git_util.py
```diff
@@ -15,6 +15,82 @@
...
+        self.__commit_hash = short_commit_hash[:self.hash_length()]
+
+    @property
+    def hash(self) -> str:
+        return self.__commit_hash
+
+    @staticmethod
+    @abc.abstractmethod
+    def hash_length() -> int:
+        pass
```
```suggestion
        """Required length of the CommitHash"""
```

varats-core/varats/utils/git_util.py
```diff
@@ -15,6 +15,82 @@
...
+    def __str__(self) -> str:
+        return self.hash
+
+    def __repr__(self) -> str:
+        return self.hash
+
+    def __lt__(self, other: tp.Any) -> bool:
+        if isinstance(other, CommitHash):
+            return self.hash < other.hash
+        return False
```
vulder: Not sure if this is a good idea, this would define an ordering just based on the string, however, someone might not expect this as an ordering. Another ordering we use is regarding the `git commit graph` ala commit ID. Hence, now by default someone can get the wrong ordering and the problem only surfaces very late when inspecting the results.

I would suggest to provide an external function for sorting commits with a standalone comparator.
```
def sort_commits_lexicographically(foo: tp.List[Commits]):
    pass

def compare_lexicographically(c1, c2):
    pass
```

varats-core/varats/utils/git_util.py
```diff
@@ -15,6 +15,82 @@
     get_primary_project_source,
 )
 
+FULL_COMMIT_HASH_LENGTH = 40
+SHORT_COMMIT_HASH_LENGTH = 10
```
vulder: Does it not make more sense to define them on the class for which they are just, compared to on `git_util` module level?

varats-core/varats/utils/git_util.py
```diff
@@ -15,6 +15,82 @@
...
+        return self.hash[:SHORT_COMMIT_HASH_LENGTH]
+
+    def to_short_commit_hash(self) -> ShortCommitHash:
+        return ShortCommitHash(self.hash)
+
+    def startswith(self, short_hash: CommitHash) -> bool:
+        return self.hash.startswith(short_hash.hash)
+
+
+DUMMY_COMMIT_HASH = FullCommitHash("0000000000000000000000000000000000000000")
```
```suggestion
UNCOMMITED_COMMIT_HASH = FullCommitHash("0000000000000000000000000000000000000000")
```

vulder: Or `NOT_COMMITED_COMMIT_HASH` something like this

vulder (changes requested):
Just an idea, should be introduce short cuts for the names?
FullCH
ShortCH

or 

FCH
SCH

vulder (approved):

PR #503, opened 2021-07-22
Add a table showing general info about case studies

boehmseb:


tests/table/test_case_study_metrics_table.py
```diff
@@ -0,0 +1,33 @@
...
+
+        # latex booktabs is default format
+        table_str = CaseStudyMetricsTable().tabulate()
+
+        self.assertEqual(
+            r"""\begin{tabular}{llrrr}
+\toprule
+{} &       Domain &    LOC &  Commits &  Authors \\
+\midrule
+\textbf{brotli} &  compression &  37033 &     1030 &       87 \\
```
vulder: Will tests not be brittle? To calculate the commits/locs/authors we clone the brotli repo and run our code (which is good) but it the brotli repo state changes, the tests afterwards will fail.

Can we pin the table calc to a specific brotli repo version?

vulder (changes requested):

tests/table/test_case_study_metrics_table.py
boehmseb: You are right. I did not think about that. I will try to fix that

vulder (approved):

PR #504, opened 2021-07-22
Fix import cycles breaking the docs.

boehmseb:


docs/source/vara-ts-api/utils.rst
```diff
@@ -21,18 +21,14 @@ Module: cli_util
     :undoc-members:
     :show-inheritance:
 
------
```
vulder, line 24: Why did you remove these? If everything works, they should generate divider lines between the different modules, so it becomes easier to see where a new one starts.

boehmseb: Because they cause warnings.

You can read the reason why that raises warnings here: https://sourceforge.net/p/docutils/mailman/message/6975902/

vulder: What a bullshit... put a picture of a line there... yeah, surely not going to add that

I actually don't care if it "should be there" or not for text as this is documentation an separates automatically generated class docs. Either we ignore the warning or replace the dashes with inserted html or so (`<hr>` could do the trick).

vulder (approved):

PR #508, opened 2021-08-03
Implements vara-run command line tool

boehmseb:
This tool allows running benchbuild experiments directly from the tool suite.
It provides the (to us) relevant functionalities of benchbuild's `run`, `slurm`, and `container` commands in a unified interface and enables a simpler handling of paper configs. For example it checks whether selected projects and versions are contained in the current paper config and if no projects are specified, it runs the experiment on all case studies in the current paper config.

resolves se-sic/VaRA#777


boehmseb:
We probably should also write a short tutorial/documentation before merging this.

varats/varats/tools/driver_run.py
```diff
@@ -0,0 +1,175 @@
...
+@click.option(
+    "-E", "--experiment", required=True, help="The experiment to run."
+)
+@click.option(
+    "-P",
+    "--project",
+    "projects",
+    multiple=True,
+    callback=__validate_project_parameters,
+    help="Only run experiments for the given project."
```
vulder:
```suggestion
    help="Only run experiments for the given project. "
```

docs/source/tutorials/getting_started.rst
```diff
@@ -0,0 +1,57 @@
...
+    pip3 install varats
+
+
+Post Install Steps
+------------------
+
+After a successful install, we need to create a proper environment to run experiments in.
+First, we need to select a directory where we create our setup.
+From now on, we refer to this directory as the `VaRA-TS root`.
+**Unless noted otherwise, ``vara-*`` commands should always be run from this directory.**
```
vulder:
```suggestion
**Unless noted otherwise,** ``vara-*`` **commands should always be run from this directory.**
```

vulder (changes requested):
- [x] Pass extra -vvvv flags
- [x] Maybe provide project list as last paramters (make projects arguments instead of options)
- [x] Stream output from BB to stdout
- [x] Add option for auto submitting (by default no submission, with option submission)

vulder (approved):
Looks good to me. Let's wait for some feedback from our other testers and then merge it :grin:

Sinerum (approved):
Looks good to me. Test went smooth.

MMory:
The output of vara-run -vvvv is not consistent with benchbuild -vvvv - when compiling xz, vara-run does not print the output of the configure script and furthermore the coloring does not work.

boehmseb:
> The output of vara-run -vvvv is not consistent with benchbuild -vvvv - when compiling xz, vara-run does not print the output of the configure script and furthermore the coloring does not work.

As for the color, I think this is stripped because I'm using `tee` to forward the output. Not sure how I can circumvent this. The plumbum documentation doesn't say anything about that issue.
I have no idea however, why the output of configure is missing. Both, `stdout` and `stderr` should be forwarded.
I might have to ask the BenchBuild  dev what other tricks he uses there :thinking:

cormensratio (approved):
Test runs worked flawlessly. Lgtm

J03D03 (approved):
Tested the `JustCompile` experiment with the `vara-run` command line tool against several C projects.
Lgtm

boehmseb:
We have 2 smaller issues left:

se-sic/VaRA#797
se-sic/VaRA#798

vulder:
resolves se-sic/VaRA#779

PR #510, opened 2021-08-12
Plot rework for the interaction-degree-multi-lib-plot

cormensratio:
Adapts the `b_interaction_degree_multi_lib` plot to the click CLI.


varats/varats/plots/blame_interaction_degree.py
```diff
@@ -1099,39 +1117,21 @@ def _degree_plot(self, degree_type: DegreeType) -> None:
...
-        self,
-        view_mode: bool,
-        degree_type: DegreeType,
-        extra_plot_cfg: tp.Optional[tp.Dict[str, tp.Any]] = None,
-        with_churn: bool = True
-    ) -> None:
-        plot_cfg = {
-            'linewidth': 1 if view_mode else 0.25,
-            'legend_size': 8 if view_mode else 2,
-            'xtick_size': 10 if view_mode else 2,
```
boehmseb, line 1112: Plot config options can have separate defaults for view mode. If these values here are consistent between mutliple plots, you may put these as defaults in the PlotConfig class.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -1451,6 +1435,41 @@ def calc_missing_revisions(
...
+                show_churn=self.plot_kwargs["show_churn"],
+                edge_color=self.plot_kwargs["edge_color"],
+                colormap=self.plot_kwargs["colormap"],
+                show_cve=self.plot_kwargs["show_cve"],
+                show_bugs=self.plot_kwargs["show_bugs"],
+                cve_line_width=self.plot_kwargs["cve_line_width"],
+                bug_line_width=self.plot_kwargs["bug_line_width"],
+                cve_color=self.plot_kwargs["cve_color"],
+                bug_color=self.plot_kwargs["bug_color"],
+                vertical_alignment=self.plot_kwargs["vertical_alignment"]
```
boehmseb: Didn't we agree to just pass the kwargs through to the plot as-is?

boehmseb (changes requested):

varats/varats/plots/blame_interaction_degree.py
cormensratio: Some of them are indeed consistent. I'll change these defaults appropriately.

varats/varats/plots/blame_interaction_degree.py
cormensratio: You are right, I only passed the `PlotConfig` as a whole and forgot the `kwargs`. This is also the case for the reworked plots of the `PlotRework` branch. Therefore I am going to open another PR to fix these too.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -1451,6 +1435,29 @@ def calc_missing_revisions(
...
+):
+    """Generates multi-lib degree plot(s) for the selected case study(ies)."""
+
+    def generate(self) -> tp.List[Plot]:
+        return [
+            BlameInteractionDegreeMultiLib(
+                self.plot_config, **{
+                    **self.plot_kwargs, "case_study": cs
+                }
+            ) for cs in self.plot_kwargs["case_study"]
```
boehmseb: I think this can be simplified:

```suggestion
            BlameInteractionDegreeMultiLib(
                self.plot_config, case_study=cs, **plot_kwargs
            ) for cs in self.plot_kwargs["case_study"]
```

boehmseb (changes requested):
Almost done ;)

varats/varats/plots/blame_interaction_degree.py
cormensratio: Initially I tried the same, but this fails with:
`TypeError:  varats.plots.blame_interaction_degree.BlameInteractionDegreeMultiLib() got multiple values for keyword argument 'case_study'`

boehmseb: Then I suggest deleting the entry from the kwargs before. This makes it also more obvious that it is overridden.

boehmseb (approved):

varats/varats/plot/plots.py
```diff
@@ -326,7 +326,7 @@ def __init__(self, view: bool, *options: PlotConfigOption[tp.Any]) -> None:
                 PlotConfigOption(
                     "legend_size",
                     default=2,
-                    view_default=2,
+                    view_default=8,
```
vulder, line 329: Just to double check, the new default view numbers are the numbers that are common between the different plots?

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -1451,6 +1435,29 @@ def calc_missing_revisions(
...
 
+class BlameInteractionDegreeMultiLibGenerator(
+    PlotGenerator,
+    generator_name="interaction-degree-multi-lib-plot",
+    options=[
+        REQUIRE_REPORT_TYPE, REQUIRE_MULTI_CASE_STUDY, REQUIRE_BASE_LIB,
+        REQUIRE_INTER_LIB, OPTIONAL_SHOW_CHURN, OPTIONAL_EDGE_COLOR,
+        OPTIONAL_COLORMAP, OPTIONAL_SHOW_CVE, OPTIONAL_SHOW_BUGS,
+        OPTIONAL_CVE_LINE_WIDTH, OPTIONAL_BUG_LINE_WIDTH, OPTIONAL_CVE_COLOR,
+        OPTIONAL_BUG_COLOR, OPTIONAL_VERTICAL_ALIGNMENT
```
vulder, line 1446: @boehmseb Maybe it could make sense to group some of those conceptually together? Like, `get_all_X()`, so one does not have so specify everything by hand. *not sure* if this makes it easier or less clear

vulder (approved):

varats/varats/plots/blame_interaction_degree.py
boehmseb: Hmm, the problem here is that those options occur in many different combinations. It could make sense for things like `OPTIONAL_SHOW_CVE`, `OPTIONAL_CVE_COLOR`, and `OPTIONAL_CVE_LINE_WIDTH`, but for many others I don't see that much benefit. We have to find a balance between increasing readability here, and having too many groups of options.

vulder: Yes, that's group is in the direction of my idea, however, I'm still not sure about the trade offs.
@cormensratio What do you think?

varats/varats/plot/plots.py
cormensratio: Exactly. Before the plot rework, there were some plots that explicitly set a view default for these options in a `extra_plot_cfg` or `plot_cfg` `dict`. The values of these options were the same in each `dict` and therefore the origin of our `view_default` values here.

varats/varats/plots/blame_interaction_degree.py
cormensratio: I think that we currently would not really benefit from such groups, mainly because of 2 reasons:
1. These `OPTIONAL_...`s are currently specific for one or two plots. If they were more commonly used among other plots they would already be in the `PlotConfig` or `CommonPlotOptions`. Groups out of these options would therefore hardly find another application as in the few plots that already use them.
2. As @boehmseb already stated, I think that the introduction of such groups would result in too many combinations, too. As long as we do not have more plots that rely on the same set of options I think we have a better readability by sticking to the current state and pass them one by one.

vulder: Ok, then we leave them out for now. Should a pattern emerge, we can always adapt things :)

PR #514, opened 2021-08-12
Plot rework for the fraction-overview-plot

cormensratio:
Adapts the `b_interaction_fraction_overview` plot to the click CLI.


varats/varats/plots/blame_interaction_degree.py
```diff
@@ -440,102 +440,97 @@ def _generate_degree_stackplot(
...
         x_axis = churn_axis
     else:
         x_axis = in_axis
 
+    project_name: str = plot_kwargs["case_study"].project_name
     fig.subplots_adjust(top=0.95, hspace=0.05, right=0.95, left=0.07)
     fig.suptitle(
-        str(plot_cfg['fig_title']) + f' - Project {plot_kwargs["project"]}',
-        fontsize=8
+        f'{plot_config.fig_title("Fraction overview")} - Project {project_name}'
```
boehmseb: I'm not sure I like the idea that the user can only customize a part of the title. I believe the most common use case to override the title via command line is to customize figures for use in a paper or presentation. And in that case I would like to have full control over the title. I see the point that one might still like to include the project name but having both would mean to implement some kind of templating solution (which I think is overkill here).

What's your opinion on this?

cormensratio: I agree. With the the new default value handler for `PlotConfig` values one could leave the title untouched and get, e.g., the project name as a default, so this information is not lost. Therefore allowing the user to modify the complete title makes sense.
As this is also the case for the title in, e.g., the `degree_plot()` I'll open another PR to fix this for plots that already got "reworked". For plots that are going to be reworked I am going apply this fix directly.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -440,102 +440,96 @@ def _generate_degree_stackplot(
 
 
 def _calc_fractions(
-    unique_revisions: tp.List[FullCommitHash], all_base_lib_names: tp.List[str],
-    all_inter_lib_names: tp.List[str],
-    revision_to_base_names_mapping: tp.Dict[FullCommitHash, tp.List[str]],
-    revision_to_inter_names_mapping: tp.Dict[FullCommitHash, tp.List[str]],
-    revision_to_dataframes_mapping: tp.Dict[FullCommitHash, pd.DataFrame],
-    revision_to_total_amount_mapping: tp.Dict[FullCommitHash, int]
+    unique_short_revisions: tp.List[ShortCommitHash],
```
vulder: Was this initially typed incorrectly? Or what did change that we now only have `ShortCommitHashes`?

boehmseb: I can see that this was changed because of the change in line 1194. But I have no idea why since the commits just got completed to their long form in the line before.
@cormensratio what's the background here?

boehmseb (changes requested):

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -1200,25 +1186,35 @@ def _fraction_overview_plot(
...
+        grouped_df = grouped_df.groupby(['revision'])
+
+        revision_to_dataframes_mapping: tp.Dict[ShortCommitHash,
                                                 pd.DataFrame] = {}
 
-        for revision in unique_revisions:
+        for revision in unique_short_revisions:
             revision_to_dataframes_mapping[revision] = grouped_df.get_group(
                 revision.hash
             )
```
cormensratio: I initially thought that it is better to work with `ShortCommitHashes` rather than `FullCommitHashes` as the dataframe already provides `ShortCommitHashes` and the final plot image only shows the `ShortCommitHashes` anyway. My main problem there was the grouping on `ShortCommitHashes` in line 1200 and the required `FullCommitHashes` in line 1206.
As I now looked over the code again I realized that it should be rather simple to work with `FullCommitHashes` up to the point where we reduce the revision hashes to their short form for better readability in the final plot image.
I`ll push my fix in one sec. .

varats/varats/plots/blame_interaction_degree.py
cormensratio: See my comment below.

varats/varats/plots/blame_interaction_degree.py
boehmseb: Sounds good

boehmseb (approved):

vulder (approved):

PR #517, opened 2021-08-16
Plot rework for the distribution-comparison-plot

cormensratio:
Adapts the `b_distribution_comparison` plot to the click CLI.


cormensratio:
~~Problems:~~
~~1. Currently fails with: `AttributeError: 'DataFrame' object has no attribute 'churn'`. Did this plot work before the click CLI adaption?~~
~~2. Correct naming of plot arguments `var_x` and `var_y` are unclear. Files of previous plots might clear up this confusion for me.~~

varats/varats/data/databases/blame_diff_metrics_database.py
```diff
@@ -254,6 +255,20 @@ def get_successor_report_file(
     return report_files.get(succ_report_commit_hash, None)
 
 
+class BlameDiffMetrics(Enum):
```
cormensratio, line 258: Is there a way to use this `enum` directly as value for the `columns` parameter of the `BlameDiffMetricsDatabase`?

vulder: What do you mean by that?

boehmseb: I think he wants to replace the `columns` parameter in the `BlameDiffMetricsDatabase` class with something generated from this enum.

Something like `[metric.name for metric in BlameDiffMetrics]` could work.

varats/varats/plots/blame_diff_correlation_matrix.py
```diff
@@ -18,16 +18,39 @@
...
+from varats.ts_utils.click_param_types import EnumChoice
 from varats.utils.git_util import FullCommitHash
 
 LOG = logging.getLogger(__name__)
 
+REQUIRE_X_METRIC: CLIOptionTy = make_cli_option(
+    "--var-x",
+    type=EnumChoice(BlameDiffMetrics, case_sensitive=False),
+    required=True,
+    help="The metric shown on the x-axis of the distribution-comparison plot."
```
boehmseb:
```suggestion
    help="The metric shown on the x-axis of the distribution comparison plot."
```

varats/varats/plots/blame_diff_correlation_matrix.py
```diff
@@ -18,16 +18,39 @@
...
+)
+
+REQUIRE_Y_METRIC: CLIOptionTy = make_cli_option(
+    "--var-y",
+    type=EnumChoice(BlameDiffMetrics, case_sensitive=False),
+    required=True,
+    help="The metric shown on the y-axis of the distribution-comparison plot."
```
boehmseb:
```suggestion
    help="The metric shown on the y-axis of the distribution comparison plot."
```

varats/varats/data/databases/blame_diff_metrics_database.py
cormensratio: Exactly, because I think that this is kind of redundant. Unfortunately, I think that we have to live with that, as something like a list comprehension is not allowed as default value. If nobody wants to change anything here, I would leave it like that.

boehmseb: You could also try to implement the list comprehension as a class-level function of the enum class. Using that could maybe be allowed.

cormensratio: The class level function works 👍🏽

boehmseb (approved):

vulder (approved):

PR #518, opened 2021-08-16
Plot rework for the repo-churn-plot

cormensratio:
Adapts the `repo_churn` plot to the click CLI.


cormensratio:
~~Problems:~~

~~1.Currently fails with:~~
~~`ImportError: cannot import name 'OPTIONAL_X_TICK_SIZE' from partially initialized module 'varats.plots.blame_interaction_degree' (most likely due to a circular import) (/vara/VaRA-Tool-Suite/varats/varats/plots/blame_interaction_degree.py)`~~

~~This should be resolved when the `PlotConfig` parameter handles plot styling arguments.~~

varats/varats/plots/repository_churn.py
```diff
@@ -170,14 +174,19 @@ def draw_code_churn(
     ]
 
     axis.set_ylim(-CODE_CHURN_DELETION_LIMIT, CODE_CHURN_INSERTION_LIMIT)
-    axis.fill_between(revisions, clipped_insertions, 0, facecolor='green')
+    axis.fill_between(revision_strs, clipped_insertions, 0, facecolor='green')
     axis.fill_between(
-        revisions,
+        revision_strs,
```
vulder, line 180: If I understand this correctly, it does not matter if we add the time_id before/after filling the axes?

boehmseb: I'm not exactly sure why this was changed. Probably so that the values on the x-axis match with those from other plots that use the churn plot as a subplot.
Let's see what @cormensratio has to say about this.

cormensratio: I assumed that this is the same pattern as in the `draw_code_churn_for_revisions()` function, but as it seems this is not the case. I`ll revert it closer to the original one.

vulder (approved):

boehmseb (approved):

PR #526, opened 2021-08-17
Plot rework for the verifier-no-opt-plot

cormensratio:
Adapts the `b_verifier_report_no_opt_plot` plot to the click CLI.


cormensratio:
Currently untested due to an error in the generation of BlameVerifierNoOpt reports: 
`...benchbuild/environments/entrypoints/cli.py", line 15, in <module>
    rich.traceback.install()
AttributeError: module 'rich' has no attribute 'traceback'`

vulder (approved):
lgtm but again, ~~pre-commit is unhappy~~

varats/varats/plots/blame_verifier_report_plots.py
```diff
@@ -10,23 +10,27 @@
...
+)
 from varats.plots.case_study_overview import SUCCESS_COLOR, FAILED_COLOR
 from varats.utils.git_util import FullCommitHash
 
 LOG = logging.getLogger(__name__)
 
 
 def _get_named_df_for_case_study(
-    case_study: CaseStudy, opt_level: OptLevel
+    case_study: CaseStudy, opt_level: OptLevel, plot_kwargs: tp.Any
```
boehmseb: `plot_kwargs` as you use it is of type `Dict[str, Any]`. Only if you use the notation `**plot_kwargs` you only need to annotate the value type of the dict. Same goes for other locations in this file where you pass kwargs.

boehmseb (changes requested):

varats/varats/plots/blame_verifier_report_plots.py
cormensratio: Done

boehmseb (approved):

varats/varats/plots/blame_verifier_report_plots.py
```diff
@@ -43,9 +48,10 @@ def _get_named_df_for_case_study(
     verifier_plot_df = verifier_plot_df.loc[verifier_plot_df['opt_level'] ==
                                             opt_level.value]
     if verifier_plot_df.empty or len(
-        np.unique(verifier_plot_df['revision'])
+        np.unique([rev.hash for rev in verifier_plot_df['revision']]
+                 )  # type: ignore
```
boehmseb: Same as #527. I don't mind where you fix it. The other one will know it after merging.

cormensratio: Fixed in https://github.com/se-sic/VaRA-Tool-Suite/pull/527

PR #538, opened 2021-08-24
Adds blame interaction graphs

boehmseb:
This PR introduces graph/network representations for blame data along with some plots and tables.


vulder:
duplicate

PR #543, opened 2021-09-01
Replace revision strings with CommitHash types

cormensratio:
Starts the replacement of revision strings with `CommitHash` types by adapting the databases and the `interaction_degree_plot` as first plot. Further adaptations will be performed in the PlotRework PR (https://github.com/se-sic/VaRA-Tool-Suite/pull/453) or in one of its sub-PRs.
The current state is that most plots are broken, which should be resolved after the PlotRework PR is merged.

### DIFF
--- a/varats-core/varats/utils/git_util.py
+++ b/varats-core/varats/utils/git_util.py
@@ -43,9 +43,6 @@ class CommitHash(abc.ABC):
     def from_pygit_commit(commit: pygit2.Commit) -> 'FullCommitHash':
         return FullCommitHash(str(commit.id))
 
-    def __len__(self) -> int:
-        return self.hash_length()
-
     def __str__(self) -> str:
         return self.hash
 
@@ -333,7 +330,7 @@ def create_commit_lookup_helper(project_name: str) -> CommitLookupTy:
             commit = primary_project_repo.get(c_hash.hash)
             if commit is None:
                 raise LookupError(
-                    f"Could not find commit {c_hash.hash} in {project_name}"
+                    f"Could not find commit {c_hash} in {project_name}"
                 )
             cache_dict[primary_source_name][c_hash] = commit
             return commit
--- a/varats-core/varats/utils/git_util.py
+++ b/varats-core/varats/utils/git_util.py
@@ -43,6 +43,9 @@ class CommitHash(abc.ABC):
     def from_pygit_commit(commit: pygit2.Commit) -> 'FullCommitHash':
         return FullCommitHash(str(commit.id))
 
+    def __len__(self) -> int:
+        return self.hash_length()
+
     def __str__(self) -> str:
         return self.hash
 
@@ -330,7 +333,7 @@ def create_commit_lookup_helper(project_name: str) -> CommitLookupTy:
             commit = primary_project_repo.get(c_hash.hash)
             if commit is None:
                 raise LookupError(
-                    f"Could not find commit {c_hash} in {project_name}"
+                    f"Could not find commit {c_hash.hash} in {project_name}"
                 )
             cache_dict[primary_source_name][c_hash] = commit
             return commit
@@ -343,7 +346,7 @@ def create_commit_lookup_helper(project_name: str) -> CommitLookupTy:
         if c_hash in cache_dict[git_name]:
             return cache_dict[git_name][c_hash]
 
-        commit = repos[git_name].get(c_hash)
+        commit = repos[git_name].get(c_hash.hash)
         if commit is None:
             raise LookupError(
                 f"Could not find commit {c_hash} in "
--- a/varats/varats/data/databases/evaluationdatabase.py
+++ b/varats/varats/data/databases/evaluationdatabase.py
@@ -94,7 +94,7 @@ class EvaluationDatabase(abc.ABC):
             return data_frame[data_frame["revision"].
                               apply(lambda x: revisions.has_node(x.hash) != 0)]
 
-        # Cast all revisions to ShortCommitHash(es)
+        # Convert all revisions to ShortCommitHash(es)
         data['revision'] = data['revision'].apply(ShortCommitHash)
         data = cs_filter(data)
         return data[columns]
--- a/varats/varats/data/databases/evaluationdatabase.py
+++ b/varats/varats/data/databases/evaluationdatabase.py
@@ -91,10 +91,11 @@ class EvaluationDatabase(abc.ABC):
             revisions = CharTrie()
             for revision in case_study.revisions:
                 revisions[revision.hash] = True
-            return data_frame[data_frame["revision"].apply(
-                lambda x: revisions.has_node(ShortCommitHash(str(x)).hash) != 0
-            )]
+            return data_frame[data_frame["revision"].
+                              apply(lambda x: revisions.has_node(x.hash) != 0)]
 
+        # Cast all revisions to ShortCommitHash(es)
+        data['revision'] = data['revision'].apply(ShortCommitHash)
         data = cs_filter(data)
         return data[columns]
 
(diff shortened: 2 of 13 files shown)


varats-core/varats/utils/git_util.py
```diff
@@ -330,7 +333,7 @@ def get_commit(
             commit = primary_project_repo.get(c_hash.hash)
             if commit is None:
                 raise LookupError(
-                    f"Could not find commit {c_hash} in {project_name}"
+                    f"Could not find commit {c_hash.hash} in {project_name}"
```
boehmseb: Shouldn't this be converted with `str()` automatically by python?

varats/varats/data/databases/evaluationdatabase.py
```diff
@@ -91,10 +91,11 @@ def cs_filter(data_frame: pd.DataFrame) -> pd.DataFrame:
             revisions = CharTrie()
             for revision in case_study.revisions:
                 revisions[revision.hash] = True
-            return data_frame[data_frame["revision"].apply(
-                lambda x: revisions.has_node(ShortCommitHash(str(x)).hash) != 0
-            )]
+            return data_frame[data_frame["revision"].
+                              apply(lambda x: revisions.has_node(x.hash) != 0)]
 
+        # Cast all revisions to ShortCommitHash(es)
```
boehmseb: Technically not a cast, but a conversion

varats/varats/plots/repository_churn.py
```diff
@@ -200,8 +200,15 @@ def draw_code_churn_for_revisions(
...
+
+    # TODO: Fix duplicated revisions on x-axis caused by the prepended
+    #  time_id when the code_churn is activated. Relates to
+    #  https://github.com/se-sic/VaRA-Tool-Suite/pull/453
+
+    # TODO: Move shown revision length to PlotConfig with length=10 as
+    #  default. Relates to https://github.com/se-sic/VaRA-Tool-Suite/pull/453
     revisions = churn_data.time_id.astype(str) + '-' + churn_data.revision.map(
-        lambda x: x[:10]
+        lambda x: x.hash[:10]
```
boehmseb: This case is so common that `FullCommitHash` has its own property for this:

```suggestion
        lambda x: x.short_hash
```

varats-core/varats/utils/git_util.py
```diff
@@ -43,6 +43,9 @@ def hash_length() -> int:
     def from_pygit_commit(commit: pygit2.Commit) -> 'FullCommitHash':
         return FullCommitHash(str(commit.id))
 
+    def __len__(self) -> int:
```
boehmseb: What do you need this for?
I don't really like the idea of allowing `len(commit_hash)` since the length of a commit hash is a static property based on the class and does not depend on the concrete instance.

varats/varats/plots/blame_interaction_degree.py
```diff
@@ -170,8 +175,9 @@ def _generate_stackplot(
     fig.subplots_adjust(top=0.95, hspace=0.05, right=0.95, left=0.07)
     fig.suptitle(plot_cfg["fig_suptitle"], fontsize=8)
 
+    unique_rev_strings: tp.List[str] = [rev.hash for rev in unique_revisions]
```
boehmseb: Maybe using `short_hash` would be more readable in a plot.

varats/varats/plots/repository_churn.py
```diff
@@ -200,8 +200,15 @@ def draw_code_churn_for_revisions(
     churn_data = build_revisions_churn_table(
         project_name, commit_map, revisions
     )
+
+    # TODO: Fix duplicated revisions on x-axis caused by the prepended
+    #  time_id when the code_churn is activated. Relates to
+    #  https://github.com/se-sic/VaRA-Tool-Suite/pull/453
```
boehmseb: What exactly is the problem here and how does it relate to the plot rework?

varats/varats/plots/repository_churn.py
```diff
@@ -200,8 +200,15 @@ def draw_code_churn_for_revisions(
...
+
+    # TODO: Move shown revision length to PlotConfig with length=10 as
+    #  default. Relates to https://github.com/se-sic/VaRA-Tool-Suite/pull/453
```
boehmseb: In my opinion, this should be consistent with the length of short commit hashes and therefore not be configurable by plots.

boehmseb (changes requested):

varats-core/varats/utils/git_util.py
cormensratio: You are right.

varats/varats/data/databases/evaluationdatabase.py
cormensratio: Changed it to `Convert`

varats-core/varats/utils/git_util.py
cormensratio: Some subroutine failed due to the missing `__len__`. Now I tried to build the database again and generated a plot without any issues, so i guess I can remove it again.

varats/varats/plots/blame_interaction_degree.py
cormensratio: Wasn't sure for the cases where the `short_hash` is not unique, but for most cases I agree. Changed it now to `short_hash`es.

varats/varats/plots/repository_churn.py
cormensratio: Perhaps this option would bring more trouble than good. Changed it to the fixed length of a `short_hash`.

varats/varats/plots/repository_churn.py
cormensratio: With the implementation as it is right now, we have all unique revision strings as x-ticks on the x-axis. When we pass the option to additionally plot the `code_churn`, the revisions are prepended with their `time_id` in the corresponding function `draw_code_churn_for_revisions()`. Now we have two versions of the x-ticks, one with and one without the `time_id`s prepended, which leads to a broken plot as we now have duplicates. 
I am a little confused right now how prior implementations did not cause the same wrong behaviour, as it seems that the corresponding code parts did not get changed, or I do not know how my implementation is involved in this.
I attach an example of such a broken plot below.
TwoLibsOneProjectInteractionDiscreteLibsSingleProject_0_b_interaction_degree

varats/varats/plots/blame_interaction_degree.py
boehmseb: Collisions are only expected in larger projects. And for displaying in a plot, it is not that important.

varats/varats/plots/repository_churn.py
boehmseb: I would blame that to changing the function `_get_unique_revisions`. I would assume that at some point, it generated the same format as the churn code. Maybe the best way to fix this is to make the code churn plotting code use the same format as the rest and then replace the axis labels with the version including the time id.

cormensratio: I'll try that, thank you.

varats/varats/plots/repository_churn.py
```diff
@@ -221,6 +221,12 @@ def draw_code_churn_for_revisions(
         0,
         facecolor='red'
     )
+    revisions = churn_data.time_id.astype(str) + '-' + churn_data.revision.map(
+        lambda x: x.short_hash
+    )
+    with warnings.catch_warnings():
```
cormensratio: Catches `'UserWarning: FixedFormatter should only be used together with FixedLocator'` due to a missing, but redundant `set_xticks()` call.

boehmseb (approved):

varats/varats/plots/repository_churn.py
```diff
@@ -158,7 +158,7 @@ def draw_code_churn(
     code_churn = sort_df(code_churn)
 
     revisions = code_churn.time_id.astype(str) + '-' + code_churn.revision.map(
-        lambda x: x[:10]
+        lambda x: x
```
boehmseb: I think that map can safely be removed ;)

cormensratio: True ;)

PR #544, opened 2021-09-02
Introduces enum for project domains

vulder:
Introduces a collection of different project domains, encoded into a enum, so we can guarantee that projects uniformly specify their project domains.

resolves se-sic/VaRA#792


vulder:
Needs benchbuild support (https://github.com/PolyJIT/benchbuild/pull/433) to fully work, otherwise, an exception is triggered in BB when storing the domain into the in memory database.

boehmseb (approved):

PR #546, opened 2021-09-03
Updates max supported python version to 3.9

vulder:


requirements.txt
```diff
@@ -27,9 +27,9 @@ pyzmq>=19.0.0
 requests>=2.23.0
 requests_cache>=0.5.2
 rich>=1.3.1
-scikit-learn~=0.23.1
+scikit-learn~=1.0.2
 seaborn>=0.8.0
-statsmodels~=0.11.1
+statsmodels~=0.13.1
```
boehmseb: @vulder do you remember if there was a reason why these packages are pinned to a specific version instead of only giving a lower bound?

vulder: If I remember correctly, that was produced by some tool at some point. From my point of view, we can just give lower bounds here.

requirements.txt
```diff
@@ -27,9 +27,9 @@ pyzmq>=19.0.0
 requests>=2.23.0
 requests_cache>=0.5.2
 rich>=1.3.1
-scikit-learn~=0.23.1
+scikit-learn~=1.0.2
```
vulder:
```suggestion
scikit-learn>=1.0.2
```

requirements.txt
vulder:
```suggestion
statsmodels>=0.13.1
```

boehmseb (approved):

PR #551, opened 2021-09-22
Cleans up result folder handling

vulder:


varats-core/varats/experiment/experiment_util.py
```diff
@@ -26,12 +26,34 @@
 from varats.utils.settings import vara_cfg, bb_cfg
 
 
+def get_vara_result_folder(project: Project) -> Path:
```
boehmseb: A test for this function would be nice. This should be easy enough to do.

boehmseb: This is maybe a bit picky, but is it good to include `vara` in the function name? Other tools/experiments also put their results there.

boehmseb (changes requested):

cormensratio (approved):
Lgtm

varats-core/varats/experiment/experiment_util.py
vulder: done

vulder: varats is probably better, that's what I aimed for anyhow

boehmseb (approved):

PR #555, opened 2021-09-28
Use benchbuild's --force-tty and --force-watch-unbuffered in vara-run

boehmseb:


vulder (approved):

vulder:
resolves se-sic/VaRA#798
resolves se-sic/VaRA#797

PR #556, opened 2021-10-07
Fix include number spec in JustCompile docs

boehmseb:


docs/source/vara-ts-api/experiments/just_compile.rst
```diff
@@ -2,7 +2,7 @@ Experiment: JustCompile
 =======================
 
 .. literalinclude:: ../../../../varats/varats/experiments/base/just_compile.py
-    :lines: 25-76,79-112
+    :lines: 26-68,70-101
```
vulder, line 5: Do you shorten the just compile part on purpose? because this reduces the code part by ~8 lines

boehmseb: I didn't shorten anything. PR #551 modified `just_compile.py` but did not update the docs accordingly.

vulder: Hmm, why is the docs change not in the PR? I know for 100% that fixed it after you mentioned the docs. Maybe it was not pushed :thinking:

vulder (approved):

PR #557, opened 2021-10-12
Port vara-cs to use click

Sinerum:
Basic  port of the functionality of vara-cs to use the click package. For now ext functionality is removed, will be added again in the acctual rework.

resolves se-sic/VaRA#793

related se-sic/VaRA-Tool-Suite#561
related se-sic/VaRA-Tool-Suite#563
related se-sic/VaRA-Tool-Suite#562


varats/varats/tools/driver_casestudy.py
```diff
@@ -1,682 +1,477 @@
...
+
+
+@click.group()
+@configuration_lookup_error_handler
+def main() -> None:
+    """Allow easier management of case studies."""
+    initialize_cli_tool()
+    initialize_projects()
+    initialize_reports()
+    initialize_plots()  # needed for vara-cs ext smooth_plot
```
boehmseb: If this is only needed for ext, then I suggest moving it to the subcommand.

varats/varats/tools/driver_casestudy.py
```diff
@@ -1,682 +1,477 @@
...
+@main.command("package")
+@click.option("-o", "--output", help="Output file")
+@click.option(
+    "--filter-regex",
+    help="Provide a regex to only include case "
+    "studies that match the filter.",
+    type=str,
+    default=".*"
+)
+@click.argument("report_names", type=_create_report_type_choice(), nargs=-1)
```
boehmseb: Consider moving directly below `@main.command` to be consistent with the other subcommands.

boehmseb (changes requested):
Some commands and arguments have lost their documentation or help strings. Please (re-) add them. Other than that, I found only some small things.

vulder:
@Sinerum Regarding the integration tests. If it helps, try to replace integration tests with unittests. With the click unittesting setup, handling these cases there is much easier.

@boehmseb What do you think, how close are we to replace all of them? Especially, because some in the chain depend on the previous steps, which makes removing them a bit more difficult.

vulder (changes requested):
Overall the port looks good, we just need to fix the interface comparability issues, which fail the integration tests.

boehmseb:
> @Sinerum Regarding the integration tests. If it helps, try to replace integration tests with unittests. With the click unittesting setup, handling these cases there is much easier.
> 
> @boehmseb What do you think, how close are we to replace all of them? Especially, because some in the chain depend on the previous steps, which makes removing them a bit more difficult.

All tests where the CLI is ported to click should be ready to be ported. You can always prepare the environment in the test to your likings, so dependencies between tests should not be necessary anymore. However, until all tests are replaced, it might be necessary to preserve some of them in the old script so that tests that are not yet replaced continue to work.

The failing `gen` test is one of those that needs to be preserved for following tests to work properly. However, it should still be possible to write an equivalent test with unittest.

varats/varats/tools/driver_casestudy.py
```diff
@@ -4,420 +4,222 @@
...
 from varats.utils.git_util import ShortCommitHash
 from varats.utils.settings import vara_cfg
 
 LOG = logging.getLogger(__name__)
 
 
+def _create_report_type_choice() -> TypedChoice[tp.Type[BaseReport]]:
+    """Create a choice parameter type that allows selecting a report type."""
+    initialize_reports()
+    return TypedChoice(BaseReport.REPORT_TYPES)
```
boehmseb: This should now be available from the `click_param_types` module.

boehmseb (changes requested):
The report type choice is now also available from the new `click_param_types` module. Otherwise, looks good.

PR #558, opened 2021-10-21
Remove passing of explicit generator plot_kwargs

cormensratio:
Removes the explicit pass of `plot_kwargs` to plots within their corresponding generator, but passes all `plot_kwargs` instead.
Affected plots/generators: `interaction-degree-plot`, `sankey-plot-rev`, `sankey-plot-cs`, `graphviz-plot-rev`, `graphviz-plot-cs`, `cs-overview-plot`, `pc-overview-plot`

### DIFF
--- a/varats/varats/plots/blame_interaction_degree.py
+++ b/varats/varats/plots/blame_interaction_degree.py
@@ -1542,10 +1542,8 @@ class SankeyLibraryInteractionsGeneratorRev(
         c_map: CommitMap = get_commit_map(
             self.plot_kwargs["case_study"].project_name
         )
-        self.plot_kwargs.update(
-            revision=c_map.convert_to_full_or_warn(
-                ShortCommitHash(self.plot_kwargs["revision"])
-            )
+        self.plot_kwargs["revision"] = c_map.convert_to_full_or_warn(
+            ShortCommitHash(self.plot_kwargs["revision"])
         )
         return [BlameLibraryInteractions(self.plot_config, **self.plot_kwargs)]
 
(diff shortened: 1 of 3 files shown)


varats/varats/plots/blame_interaction_degree.py
```diff
@@ -1652,22 +1637,17 @@ class GraphvizLibraryInteractionsGeneratorRev(
...
 
     def generate(self) -> tp.List[Plot]:
+        c_map: CommitMap = get_commit_map(
+            self.plot_kwargs["case_study"].project_name
+        )
+        self.plot_kwargs.update(
+            revision=c_map.convert_to_full_or_warn(
+                ShortCommitHash(self.plot_kwargs["revision"])
+            )
+        )
```
boehmseb: Why use `update()` here and not simply `self.plot_kwargs[revision] = cmap.convert_to_full_or_warn(...)`?

boehmseb (changes requested):

varats/varats/plots/blame_interaction_degree.py
cormensratio: Does it make a difference semantically or is `update()` just not as readable?

boehmseb: I'd say that `update()` is less readable. It is usually used to update multiple values at once. But here, you only set a single value.

cormensratio: Thanks for the clarification, I'll change it accordingly.

boehmseb (approved):

vulder (approved):

PR #560, opened 2021-10-21
Updates CLion setup docs

boehmseb:


docs/source/research_tool_docs/vara/add_setup_infos/clion_setup.rst
```diff
@@ -1,83 +1,117 @@
...
+   Use the **+** symbol to create a new profile and adjust the settings as follows and confirm with **Ok** once you are done:
+
+   - Debug build:
+      - **Name:** Debug
+      - **Build type:** Debug
+      - **Toolchain:** Default  (make sure that your toolchain is configured to use clang)
+      - **CMake options:**
+        .. code-block::
+
+           -DBUILD_SHARED_LIBS=ON
```
vulder, line 22: Do you think it makes sense/is needed to list all options here? Does it not work otherwise with Clion?

docs/source/research_tool_docs/vara/add_setup_infos/clion_setup.rst
```diff
@@ -1,83 +1,117 @@
...
+           -DLLVM_TARGETS_TO_BUILD=X86
```
vulder: It was not the case before, but could you please sort them, makes reading a long list easier :laughing:

docs/source/research_tool_docs/vara/add_setup_infos/clion_setup.rst
```diff
@@ -1,83 +1,117 @@
...
+           -DLLVM_ENABLE_PROJECTS="clang;lld;compiler-rt;clang-tools-extra;vara;phasar"
+           -DCMAKE_CXX_STANDARD=17
+           -DLLVM_ENABLE_BINDINGS=OFF
+           -DCMAKE_INSTALL_PREFIX=<varats_root>/tools/VaRA
+
+      - **Build directory:** ``<varats_root>/tools_src/vara-llvm-project/build/dev-clion``
+      - **Build options:** leave empty
+      - **Environment:**
+         - ``CFLAGS=-O2 -g -fno-omit-frame-pointer``
+         - ``CXXFLAGS=-O2 -g -fno-omit-frame-pointer``
```
vulder: You don't have to pass a `env` vars if you just add the value directly to the cmake variable `CMAKE_C_FLAGS` etc. For the setup script it was just easier to aggregate and pass them in this way.

docs/source/research_tool_docs/vara/add_setup_infos/clion_setup.rst
```diff
@@ -1,83 +1,117 @@
...
+
+2. Start CLion and from the menu select **File | Open** and point to ``<varats_root>/vara-llvm-project/llvm/CMakeLists.txt``.
+   In the dialog that opens, click **Open as Project**.
+
+3. Go to **Settings/Preferences | Build, Execution, Deployment | CMake** to configure the CMake project.
+   Use the **+** symbol to create a new profile and adjust the settings as follows and confirm with **Ok** once you are done:
+
+   - Debug build:
+      - **Name:** Debug
+      - **Build type:** Debug
```
vulder: Should we call this Debug here? From the setup script there is `Debug` vs `Dev`, from the CFLAGS that I see below this looks more like an `Dev` build. They are quite similar, the big difference is `-O0` vs `-O2`.
From a naming point of view, we should keep things consistent, however, I don't like duplicating the rest. Maybe we write Name: Debug/Dev
And below when setting the cmake variable with the CFLAGS we say something like this: For Debug pass `-O0` and for Dev pass `-O2`

vulder (changes requested):

docs/source/research_tool_docs/vara/add_setup_infos/clion_setup.rst
boehmseb: I just listed all options that would also get set when using `vd` to set up a dev build.

docs/source/research_tool_docs/vara/add_setup_infos/clion_setup.rst
boehmseb: I tried to group the variables the same way they are grouped in your build config scripts.

docs/source/research_tool_docs/vara/add_setup_infos/clion_setup.rst
boehmseb: Yes, that's the dev build. I can include both here.

docs/source/research_tool_docs/vara/add_setup_infos/clion_setup.rst
boehmseb: Both things are equally easy to configure in CLion. As I said, I tried to stay close to the way how the build config scripts configure things. I figured, that way it is easier to stay in sync (and for you to find problems, e.g., when students use CLion)

vulder (approved):

PR #562, opened 2021-11-03
Port vara-config to click and add a unit test equivalent to the smoke test

Sinerum:


boehmseb (changes requested):
Please mark or delete the old smoke test so we can keep track of what is already ported.

vulder (approved):
after marking the old smoke tests

lgtm

boehmseb (approved):

PR #568, opened 2021-12-03
Remove pylint suppression that is no longer needed and improve pylint config.

boehmseb:
Resolves se-sic/VaRA#717


vulder (approved):
If it works it works, I guess.

Let's hope pylint has regression tests^^

boehmseb:
According to the release notes this is fixed since version 2.9. The current version is 2.12, so we are fine.

PR #570, opened 2021-12-11
Refactor vara-cs status to be based on experiments

vulder:
related se-sic/VaRA#791


tests/tools/test_driver_casestudy.py
```diff
@@ -0,0 +1,63 @@
...
+    run_in_test_environment,
+    TEST_INPUTS_DIR,
+    UnitTestInputs,
+)
+from varats.paper_mgmt.paper_config import load_paper_config
+from varats.tools import driver_casestudy
+from varats.utils.settings import vara_cfg, save_config
+
+
+class TestDriverContainer(unittest.TestCase):
```
boehmseb: Copy-Paste error?

varats-core/varats/report/report.py
```diff
@@ -22,9 +22,10 @@ class FileStatusExtension(Enum):
     value: tp.Tuple[str, Color]  # pylint: disable=invalid-name
 
     SUCCESS = ("success", colors.green)
+    PARTIAL = ("partial", colors.orangered1)
```
boehmseb: To me, `INCOMPLETE` sounds a bit better fitting to this use case than `PARTIAL`. What do you think?

varats-core/varats/report/report.py
```diff
@@ -113,6 +114,25 @@ def get_file_status_from_str(status_name: str) -> 'FileStatusExtension':
...
+        """
+        if (
+            lhs == FileStatusExtension.SUCCESS and
+            rhs != FileStatusExtension.SUCCESS
+        ) or (
+            rhs == FileStatusExtension.SUCCESS and
+            lhs != FileStatusExtension.SUCCESS
+        ):
+            return FileStatusExtension.PARTIAL
+        return lhs
```
boehmseb: I could also imagine other reasonable cases (commutative of course):

- `FAILED && <Any> -> FAILED`
- `BLOCKED && <Any> -> BLOCKED`
- ...
Then precedence becomes important. So I guess we should discuss the best implementation sometime.

Also, I'd prefer to call this function `merge`.

tests/tools/test_driver_casestudy.py
vulder: This part is from @Sinerum

varats-core/varats/report/report.py
vulder: Don't care, we can name it `INCOMPLETE` if you like

varats-core/varats/report/report.py
vulder: Sure, I don't know what important combinations but all failed because one failed makes the INCOMPLETE/PARTIAL useless.
Furthermore, BLOCKED is a property of a project revisions that one cannot be different between reports.

I initially thought of merge but did not like `merge` it because it overlays to much with git terminology.

boehmseb: I wouldn't worry about the name clash with git here too much since it should be very clear from the context: `FileStatusExtension.merge(status_a, status_b)`.

boehmseb (approved):

Sinerum (approved):

PR #574, opened 2021-12-15
Implements feature-region precision experiment

Zatho:


varats-core/varats/experiment/experiment_util.py
```diff
@@ -71,7 +71,7 @@ def __call__(self, ex: Exception, func: tp.Callable[..., tp.Any]) -> None:
                 except FileNotFoundError:
                     pass
 
-        error_file = self.__result_folder / self.__error_file_name
+        error_file = f"{self.__result_folder}/{self.__error_file_name}"
```
vulder: The original implementation is correct, why do you need to convert a path back to a string?

varats/varats/experiments/vara/feature_region_verifier_experiment.py
```diff
@@ -0,0 +1,164 @@
...
+            run_cmd = opt[opt_params]
+
+            run_cmd = wrap_unlimit_stack_size(run_cmd)
+
+            run_cmd = run_cmd > f"{vara_result_folder}/{result_file}"
+
+            exec_func_with_pe_error_handler(
+                run_cmd,
+                create_default_analysis_failure_handler(
+                    self.__experiment_handle, project, FRR, Path(vara_result_folder)
```
vulder:
```suggestion
                    self.__experiment_handle, project, FRR, vara_result_folder
```
The result folder is already a path

vulder (commented):
You should probably fix the mypy type errors and pylint warnings.

vulder:
Hoi, at some point we should clean up this PR and merge the parts that should/need to be kept in the codebase.

PR #575, opened 2021-12-16
Port vara-develop to click

boehmseb:


vulder (changes requested):
Code looks good, do we have any some tests or tests what so ever? If not maybe we should at least add one or two to cover the basic user interface from the outside with click.

boehmseb:
There weren't any tests for `vara-develop` since one would have to mock the whole research tool setup stuff (a test cloning llvm isn't quite feasible I'd say). Such things are possible now with click, so we could add some tests.

vulder:
I see, yeah the llvm problem :see_no_evil: 
If you have a bit of time, I'd say we add ~3 basic tests to roughly cover the code and add more should stuff break. Ok 4 y?

boehmseb:
Let's see what I can come up with.

boehmseb:
Not much useful stuff I can test here since everything is very research tool specific and needs the research tools' git repos to work with. From the driver side, all commands are direct API calls. I added a smoke test for `status` as an example, but I doubt it is very useful.

The only way I see that actually tests something useful is to create a dummy research tool for testing. That would also allow us to test the rest of the whole research tool infrastructure but requires some more work.

vulder (approved):

PR #578, opened 2021-12-26
Implements RevisionBinaryMap

vulder:
With the RevisionBinaryMap it's easier for projects to define
RevisionRanges at which a binary is valid, i.e., produced by the project.


varats-core/varats/utils/git_util.py
```diff
@@ -132,7 +136,60 @@ def get_current_branch(repo_folder: tp.Optional[Path] = None) -> str:
...
+    Args:
+        repo_folder: where the git repository is located
+
+    Returns: initial commit hash
+    """
+    if repo_folder is None or repo_folder == Path(''):
+        return __get_initial_commit_impl()
+
+    with local.cwd(repo_folder):
+        return __get_initial_commit_impl()
```
boehmseb: Probably not a huge issue here, but I always prefer using `git -C <path>` instead of `local.cwd` since I found the latter to be quite slow if called often.

varats-core/varats/utils/git_util.py
```diff
@@ -132,7 +136,60 @@ def get_current_branch(repo_folder: tp.Optional[Path] = None) -> str:
...
+        c_start: first commit of the range
+        c_end: last commit of the range
+        short: shorten revision hashes
+        repo_folder: where the git repository is located
+    """
+    if repo_folder is None or repo_folder == Path(''):
+        return __get_all_revisions_between_impl(c_start, c_end, hash_type)
+
+    with local.cwd(repo_folder):
+        return __get_all_revisions_between_impl(c_start, c_end, hash_type)
```
boehmseb: same as above.

varats-core/varats/utils/git_util.py
```diff
@@ -716,3 +773,74 @@ def calc_repo_loc(repo: pygit2.Repository, rev_range: str) -> int:
...
+    valid."""
+
+    def __init__(self, repo_location: Path) -> None:
+        self.__repo_location = repo_location
+        self.__revision_specific_mappings: tp.Dict[RevisionRange,
+                                                   ProjectBinaryWrapper] = {}
+        self.__always_valid_mappings: tp.List[ProjectBinaryWrapper] = []
+
+    def specify_binary(
+        self, location: str, binary_type: BinaryType, **kwargs: tp.Any
```
boehmseb, line 771: Do you expect that we will need more parameters in the future? Otherwise I would prefer having two optional parameters here.

tests/utils/test_git_util.py
```diff
@@ -306,3 +336,73 @@ def test_commit_range(self):
...
+        self.rv_map.specify_binary(
+            "build/bin/SingleLocalSimple",
+            BinaryType.EXECUTABLE,
+            override_binary_name="Overridden"
+        )
+
+        self.assertIn("Overridden", self.rv_map)
+
+    def test_wrong_contains_check(self) -> None:
+        """Check if we wrong values are correctly shows as not in the map."""
```
boehmseb:
```suggestion
        """Check if wrong values are correctly shows as not in the map."""
```

boehmseb (commented):
Some smaller points require discussion.

varats-core/varats/utils/git_util.py
vulder: Works for me

varats-core/varats/utils/git_util.py
vulder: Yes, there could be more. Another reason for this impl was that `binary_name` depends/can depend on the other parameter `location`, which was not easy to express in the arguments list.

boehmseb (approved):

PR #582, opened 2022-01-06
Refactor the vara-cs gen/ext interface 

Sinerum:
Refactors the vara-cs gen/ext interface to make better use of click and allow for a more intuitive use 

new sturcure is:

`vara-cs gen [general Options] select_[strategy] [strategy Options]`

additionaly a new `latest` startegy was added to easily generate the latest revision of a project
resolves se-sic/VaRA#794
resolves se-sic/VaRA#764
resolves se-sic/VaRA#679


tests/tools/test_driver_casestudy.py
```diff
@@ -0,0 +1,64 @@
...
+    run_in_test_environment,
+    TEST_INPUTS_DIR,
+    UnitTestInputs,
+)
+from varats.paper_mgmt.paper_config import load_paper_config
+from varats.tools import driver_casestudy
+from varats.utils.settings import vara_cfg, save_config
+
+
+class TestDriverContainer(unittest.TestCase):
```
boehmseb: Copy-Paste error?

varats/varats/paper_mgmt/case_study.py
```diff
@@ -502,22 +429,20 @@ def extend_with_distrib_sampling(
...
+        sampling_method.sample_n(revision_list, num_rev), merge_stage
     )
 
 
-@check_required_args('plot_type', 'boundary_gradient')
 def extend_with_smooth_revs(
-    case_study: CaseStudy, cmap: CommitMap, **kwargs: tp.Any
+    case_study: CaseStudy, cmap: CommitMap, boundary_gradient_int: int,
+    ignore_blocked: bool, plot_type: tp.Type['Plot'], merge_stage: int,
+    **kwargs: tp.Any
```
boehmseb: rename to `plot_kwargs`

varats/varats/paper_mgmt/case_study.py
```diff
@@ -527,18 +452,21 @@ def extend_with_smooth_revs(
...
+        boundary_gradient: Maximal expected gradient in percent between two revisions
     """
-    plot_type = Plot.get_class_for_plot_type(kwargs['plot_type'])
-
-    kwargs['plot_case_study'] = case_study
+    kwargs['case_study'] = case_study
     kwargs['cmap'] = cmap
-    plot = plot_type(**kwargs)
+    print(kwargs)
+    plot = plot_type(None, **kwargs)
```
boehmseb: Passing `None` here is dangerous because it requires all plots to not access the plot config in the interpolation function.

Also this wouldn't type-check because the config object is not optional even if the type checker wouldn't fail here for a different reason. Suppress the warning for now and reference the issue se-sic/VaRA#830.

As for the config object, we can either pass in the default config (which should suffice in my opinion) or we need to extract this from the CLI arguments (which would require to add all those params to the `vara-cs gen` CLI).

varats/varats/tools/driver_casestudy.py
```diff
@@ -34,390 +42,370 @@
...
+@click.option(
+    "--filter-regex",
+    help="Provide a regex to filter the shown case studies",
+    default=".*"
+)
+@click.option(
+    "--paper-config",
+    help="Use this paper config instead of the configured one",
+    default=None
+)
```
boehmseb: Didn't even know we had this option before. Do we want to keep this? It would require some additional checks whether the paper config is valid (or an appropriate click-choice type)

varats/varats/tools/driver_casestudy.py
```diff
@@ -34,390 +42,370 @@
...
+@click.option(
+    "--paper_config_path",
+    "-pc",
+    type=click.Path(exists=True),
+    default=Path(vara_cfg()["paper_config"]["folder"].value) /
+    vara_cfg()["paper_config"]["current_config"].value
+)
+@click.pass_context
+def __casestudy_gen(
```
boehmseb: We should clean up some options here:

I would kick out the paper config path for good. It comes from a time where we had no `vara-pc` and is now bound to only cause issues. At the very least restrict it to existing paper config names via an appropriate type but I would prefer to only work on the paper config specified in the varats config file.

Next, the git path should go. I would only allow to specify projects.

Then, `ext` should not specify a path, but a case study name + an optional version (which defaults to 0). Again restrict via types to existing case studies of the current paper config.

@vulder do you agree?

varats/varats/tools/driver_casestudy.py
```diff
@@ -34,390 +42,370 @@
...
+    type=int,
+    default=5,
+    help="Maximal expected gradient in percent between " +
+    "two revisions, e.g., 5 for 5%%"
+)
+@click.option(
+    "--result-folder",
+    type=click.Path(exists=True),
+    help="Folder in which to search for result files."
+)
```
boehmseb: @vulder do we really still need this option? I would prefer to use what the varats config says.

varats/varats/tools/driver_casestudy.py
```diff
@@ -34,390 +42,370 @@
...
+    )()
+    extend_with_revs_per_year(
+        ctx.obj['case_study'], cmap, ctx.obj['merge_stage'],
+        ctx.obj['ignore_blocked'], ctx.obj['git_path'], revs_per_year, separate
+    )
+    store_case_study(ctx.obj['case_study'], ctx.obj['path'])
+
+
+@__casestudy_gen.command("select_plot")
+@click.argument("plot_type", type=create_plot_type_choice())
```
boehmseb: I fear this is not as simple as you think. The plots need their plot-specific arguments to work, even for the extend case. Otherwise, most of them don't even know what data to use. We actually need to mimic the functionality of the `vara-plot` tool here, but instead of generating the plot, we need to call the `calc_missing_revisions` function. I guess we best talk about this in our next meeting.

varats/varats/tools/driver_casestudy.py
```diff
@@ -34,390 +42,370 @@
...
+    )()
+    extend_with_release_revs(
+        ctx.obj['case_study'], cmap, release_type, ctx.obj['ignore_blocked'],
+        ctx.obj['merge_stage']
+    )
+    store_case_study(ctx.obj['case_study'], ctx.obj['path'])
+
+
+@__casestudy_gen.command("select_bug")
+@click.argument("report_type", type=create_report_type_choice())
```
boehmseb: Can we restrict this to only reports from SZZ tools?

varats/varats/tools/driver_casestudy.py
```diff
@@ -568,21 +508,79 @@ def open_in_editor(result_file: Path) -> None:
...
+                longest_file_status_extension +
+                commit_status_pair[1].num_color_characters(), " "
+            )
+
+            return f"[{status}] {commit_status_pair[0]}"
+
+        # Ask user which commit we should use
+        try:
+            cli_list_choice(
+                "Please select a hash:",
```
boehmseb:
```suggestion
                "Please select a commit hash:",
```

boehmseb (changes requested):
Quite a lot going on here :smiley: 

I would take the chance to clean up some of the CLI and remove obsoloete options.

There is also one major and a few minor aspects that we should discuss in our next meeting.

tests/tools/test_driver_casestudy.py
Sinerum: yeah same as on the other branch as this comes from the parrent this has already been fixed on that branch

varats/varats/paper_mgmt/case_study.py
Sinerum: I agree passing `None` is a bad Idea. But it was an easy way to make it work and i wasn't able to find a default Config probably should have added a todo.

varats/varats/tools/driver_casestudy.py
Sinerum: yes that was an option but these changes are from the parent pr anyways

varats/varats/paper_mgmt/case_study.py
```diff
@@ -502,22 +429,19 @@ def extend_with_distrib_sampling(
...
-        sampling_method.sample_n(revision_list, kwargs['num_rev']),
-        kwargs['merge_stage']
+        sampling_method.sample_n(revision_list, num_rev), merge_stage
     )
 
 
-@check_required_args('plot_type', 'boundary_gradient')
 def extend_with_smooth_revs(
-    case_study: CaseStudy, cmap: CommitMap, **kwargs: tp.Any
+    case_study: CaseStudy, cmap: CommitMap, boundary_gradient_int: int,
```
vulder:
```suggestion
    case_study: CaseStudy, cmap: CommitMap, boundary_gradient: int,
```

varats/varats/tools/driver_casestudy.py
```diff
@@ -4,25 +4,36 @@
...
     get_revisions_status_for_case_study,
-    ExtenderStrategy,
-    extend_case_study,
-    generate_case_study,
+    extend_with_distrib_sampling,
+    extend_with_revs_per_year,
+    extend_with_smooth_revs,
+    extend_with_release_revs,
+    extend_with_bug_commits,
+    extend_with_extra_revs,
```
vulder, line 31: What's your reasoning to pull all of them directly into the driver? I initially provided the two dispatching functions in the case_study module, so they are colocated with the case study code, group togetter and we don't have to adapt the driver every time we add a new extension method.

Is it because you need to create a CLI function for every strat?

varats/varats/tools/driver_casestudy.py
```diff
@@ -33,391 +44,426 @@
...
-        type=str
-    )
-    cleanup_parser.add_argument(
-        "--silent",
-        action="store_true",
-        default=False,
-        help="Hide the output of the matching filenames."
-    )
-
+@__casestudy_gen.command("select_simple")
```
vulder: Would `select_revs` or `select_specific` not be a more fitting name?

vulder (changes requested):

varats/varats/tools/driver_casestudy.py
vulder: Yes, gen actually does also not have so specify the path, as we can just require from the user to create a paper config beforehand

varats/varats/tools/driver_casestudy.py
vulder: Yes, this one should not really be needed anymore

varats/varats/tools/driver_casestudy.py
Sinerum: Exactly because the split into strategies is now done by the CLI it seemd unecessary to keep the strategy enum and switch over that instead of calling the right function directly.

vulder (approved):

boehmseb (approved):

PR #584, opened 2022-01-12
Extends the ProjectBinaryWrapper by an entry point

vulder:
The new entry point specifies which callable should be called to execute
the wrapped binary. The entry point can be overwritten to allow for
cases where the actual binary is not the entry point to call.

### DIFF
--- a/varats-core/varats/project/project_util.py
+++ b/varats-core/varats/project/project_util.py
@@ -163,11 +163,22 @@ class ProjectBinaryWrapper():
     """
 
     def __init__(
-        self, binary_name: str, path_to_binary: Path, binary_type: BinaryType
+        self,
+        binary_name: str,
+        path_to_binary: Path,
+        binary_type: BinaryType,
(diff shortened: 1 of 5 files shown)


boehmseb (approved):

varats-core/varats/project/project_util.py
```diff
@@ -163,12 +163,23 @@ class ProjectBinaryWrapper():
...
+        entry_point: tp.Optional[Path] = None
     ) -> None:
         self.__binary_name = binary_name
         self.__binary_path = path_to_binary
         self.__type = binary_type
 
+        if binary_type is BinaryType.EXECUTABLE:
+            self.__entry_point = entry_point
+            if not self.entry_point:
+                self.__entry_point = self.path
```
vulder:
```suggestion
        if binary_type is BinaryType.EXECUTABLE:
            self.__entry_point = entry_point
            if not self.__entry_point:
                self.__entry_point = self.path
```

PR #588, opened 2022-01-21
Implements experiments to run and time traced and untraced versions of all binaries in a project

niomate:
Implemented a few experiments to time traced and untraced binaries. RunUntraced and RunTracedTime generate TimeReports and RunTracedTEF generates a TEFReport in case this is needed for the evaluation.


varats/varats/experiments/vara/compare_traced_untraced.py
```diff
@@ -0,0 +1,275 @@
...
+                binary_name=binary.name,
+                project_revision=project.version_of_primary,  # type: ignore
+                project_uuid=str(project.run_uuid),
+                extension_type=FSE.SUCCESS,
+            )
+
+            with local.cwd(local.path(project.source_of_primary)):
+                print(f"Currently at {local.path(project.source_of_primary)}")
+                print(f"Bin path {binary.path}")
+                time_cmd = local["/usr/bin/time"]
```
vulder: You can just use `from benchbuild.utils.cmd import time`
This automatically gives you a wrapped time binary, determined by a PATH lookup, which makes it a bit more flexible compare to hard coding the path.

varats/varats/experiments/vara/compare_traced_untraced.py
```diff
@@ -0,0 +1,275 @@
...
+                run_cmd = time_cmd[
+                    "-o", f"{vara_result_folder}/{result_file}", "-v", binary.path
```
vulder:
```suggestion
                    "-o", f"{vara_result_folder}/{result_file}", "-v", binary["--args", "that", "are", "passed", "to", "the", "executable"]
```
Just as a hint/further information. We are currently working on making the args generic and determined by the project.
But for testing purposes, feel free to hard code here what ever you need for the projects you want to run.

varats/varats/experiments/vara/compare_traced_untraced.py
```diff
@@ -0,0 +1,275 @@
...
+
+        for binary in project.binaries:  # type: ignore[attr-defined]
+            if binary.type != BinaryType.EXECUTABLE:
+                continue
+
+            result_file = self.__experiment_handle.get_file_name(
+                TimeReport.shorthand(),
+                project_name=str(project.name),
+                binary_name=binary.name,
+                project_revision=project.version_of_primary,  # type: ignore
```
vulder: Is this giving you a type error? Maybe BB is missing some type hints here :thinking:

varats/varats/experiments/vara/compare_traced_untraced.py
```diff
@@ -0,0 +1,275 @@
...
+                        project,
+                        TimeReport,
+                        Path(vara_result_folder),
+                    ),
+                )
+
+        return actions.StepResult.OK
+
+
+class ExecAndTraceBinary(actions.Step):  # type: ignore
```
vulder: For the second part, you don't need to create TEFReports, as you are not interested in the measured performance of the program. You need to run the instrumented program also with time. The goal is to determine how much influence the instrumentation itself has on the binary.

Conceptually:
`instr_const = time(binary_with) - time(binary_without)`

vulder (changes requested):

varats/varats/experiments/vara/compare_traced_untraced.py
niomate: It's giving me the error 

```Argument of type 'str' cannot be assigned to parameter 'project_revision' of type 'ShortCommitHash' in function 'get_file_name'. 'str' is incompatible with 'ShortCommitHash'```

I also get a few other errors with `project.compiler_extension` and `FeatureModelProvider.create_provider_for_project`

varats/varats/experiments/vara/compare_traced_untraced.py
niomate: I kept this just in case we needed it, but I guess I can just remove it and use the FeaturePerfRunner

vulder: What do you mean by that? Just use the FeatuerPerfRunner?

niomate: In case I need TEFReports I could just use the experiment implemented on the f-FeaturePerfMeasure branch

vulder: If you pull in the current `vara-dev` this should all be resolved :)

varats/varats/experiments/vara/compare_traced_untraced.py
vulder: Well, in this case mypy is right. `project_revision` needs to be a `ShortCommitHash` and not a `str`, so if you want to pass the primary version, you should make a `ShortCommitHash` out of the string.

`project_revision=ShortCommitHash(project.version_of_primary)`

PR #592, opened 2022-01-29
Fixes type errors and introduces numpy typing

vulder:


varats/varats/plots/bug_fixing_relations.py
```diff
@@ -207,6 +209,7 @@ def _generate_line_data(
                 bug_introduction] = NodeType.INTRODUCING_FIX if commit_type[
                     bug_introduction] == NodeType.FIX else NodeType.INTRODUCTION
 
+            # HERE
```
boehmseb: What does this mean?

boehmseb (changes requested):
There is one comment which I don't understand. The rest looks fine.

varats/varats/plots/bug_fixing_relations.py
vulder: nothing :rofl:

vulder:
```suggestion
```

boehmseb (approved):

PR #594, opened 2022-01-31
Automatically generate documentation for all included projects

vulder:


varats-core/varats/project/project_domain.py
```diff
@@ -30,3 +30,26 @@ class ProjectDomains(Enum):
...
+
+        return False
+
+
+class ProjectGroups(Enum):
+    """Defines a set of project groups."""
+    value: str
+
+    C_PROJECTS = "c_projects"
+    CPP_PROJECTS = "cpp_projects"
```
boehmseb, line 46: What about the `test_projects` group?

vulder: I implemented the basic group but I need a BB release to fully use it, as BB needs to convert the domain to a `str`.
I'll update this after the BB release. We can wait with this PR or make another one. Both work for me.

boehmseb: Then I would go with the separate PR.

boehmseb (approved):

PR #596, opened 2022-02-03
Implements a paper config specific git filter

vulder:


boehmseb (changes requested):
The `VaraTestRepoSource` classes in `project_util` probably need the same treatment.

boehmseb (approved):
lgtm

PR #603, opened 2022-02-18
Implement vara-cs cleanup all

Sinerum:
Implement new sub command `all` for `vara-cs cleanup`  
Add options `--experiment`, `--case-studies` and `--report` to the base command to better filter which reports are deleted
Remove subcommand `error` this functionality is now provided by a `--error` flag for the `all` subcommand.
Close se-sic/VaRA#844


boehmseb (commented):
lgtm

vulder (approved):

PR #604, opened 2022-02-23
Pull project repository before generating CaseStudy

Sinerum:
`vara-cs gen` now pulls before generating a case study, can be disabled with a `--no-update` flag.
Fixes se-sic/VaRA#847


varats-core/varats/utils/git_util.py
```diff
@@ -572,6 +572,11 @@ def map_commits(
 )
 
 
+def update_repo(git_path: Path) -> None:
+    with local.cwd(git_path):
+        git('pull')
```
boehmseb: Wouldn't it be cleaner to call `fetch` on the project's primary source?

boehmseb (changes requested):

varats-core/varats/utils/git_util.py
vulder: Depends, I think select_latest picks from master (or the current branch?)
but, anyhow we probably need to update the branch with pull and not only fetch it

vulder: I think, we should use the `pull_current_branch` implemented for the vara manager to not double the impl here.
The `pull_current_branch` does also work well when working with multiple repos in parallel

varats/varats/tools/driver_casestudy.py
```diff
@@ -156,10 +156,14 @@ def __casestudy_status(
     is_flag=True,
     help="Ignore revisions that are marked as blocked."
 )
+@click.option(
+    "--no-update", is_flag=True, help="Project repository will not be updated."
+)
```
vulder: a) which default do we want? update or no update?
b) we should invert the flag and set it by default to false, makes the code easier to understand

vulder (changes requested):

varats-core/varats/utils/git_util.py
Sinerum: It uses the current branch.

vulder: I thought so. This means pull is better than just fetch.

vulder (approved):

boehmseb (approved):

PR #610, opened 2022-03-03
FeatureModelProvider to FTA experiment

Keller-1998:
Added FeatureModelProvider to phasar_fta experiment


varats/varats/experiments/vara/phasar_fta.py
```diff
@@ -0,0 +1,167 @@
...
+
+    NAME = "PhASARFTACheck"
+    DESCRIPTION = "Generate a full FTA."
+
+    RESULT_FOLDER_TEMPLATE = "{result_dir}/{project_dir}"
+
+    def __init__(
+        self,
+        project: Project,
+        report_type: tp.Type[BaseReport],
```
vulder: It's easier to just hand the experiment handle to the Step, this let's you access a lot of the experiment infos.

varats/varats/experiments/vara/phasar_fta.py
```diff
@@ -0,0 +1,167 @@
...
+        """This step performs the actual analysis with the correct flags."""
+
+        project = self.obj
+
+        # Define the output directory.
+        vara_result_folder = self.RESULT_FOLDER_TEMPLATE.format(
+            result_dir=str(bb_cfg()["varats"]["outfile"]),
+            project_dir=str(project.name)
+        )
+        mkdir("-p", vara_result_folder)
```
vulder: There is an easier helper function for this, use `get_varats_result_folder` from `experiment_util`

varats/varats/experiments/vara/phasar_fta.py
```diff
@@ -0,0 +1,167 @@
...
+
+        for binary in project.binaries:
+            # Define empty success file
+            result_file = self.__report_type.get_file_name(
```
vulder: There was a slight change here, that makes creating these file names a bit easier/different.
+ You should not need an error filename anymore

See blame_report_experiment.py as an example

varats/varats/experiments/vara/phasar_fta.py
```diff
@@ -0,0 +1,167 @@
...
+                run_cmd, PEErrorHandler(vara_result_folder, error_file)
+            )
+
+
+class PhASARTaintAnalysis(VersionExperiment):
+    """Generates a feature taint analysis (FTA) of the project(s) specified in
+    the call."""
+
+    NAME = "PhASARFeatureTaintAnalysis"
+    REPORT_TYPE = EMPTY
```
vulder: Let's talk about adding a specific report here.

vulder (changes requested):
You should also merge the changes from vara-dev into this branch, as there are a few that should make your life a bit easier.

varats/varats/data/reports/feature_analysis_report.py
```diff
@@ -196,3 +200,145 @@ def get_feature_analysis_result_function_entry(
...
         return self.__function_entries[mangled_function_name]
+
+    def get_feature_locations_dict(self) -> tp.Dict[str, tp.Set[str]]:
+        """Returns a dictionary that maps a feature name to a list of all
+        locations of tainted br and switch instructions."""
+        feat_loc_dict: tp.Dict[str, tp.Set[str]] = {}
+        for function_entry in self.__function_entries.values():
+            for tainted_inst in function_entry.feature_tainted_insts:
+                for feature in tainted_inst.feature_taints:
+                    if tainted_inst.is_br_switch():
```
vulder: Maybe we could just call `is_br_switch` is_terminator` as we are, in general, interested in terminator instructions.

varats/varats/data/reports/feature_analysis_report.py
```diff
@@ -196,3 +200,145 @@ def get_feature_analysis_result_function_entry(
...
+
+    Can be given a list of features for providing feature-specific evaluation
+    information.
+    """
+
+    def __init__(
+        self,
+        fa_report: FeatureAnalysisReport,
+        ground_truth: FeatureAnalysisGroundTruth,
+        features: tp.List[str] = None
```
vulder: The type needs to be `tp.Optional[tp.List[str]]` to be `None`

vulder (changes requested):

varats/varats/data/reports/feature_analysis_report.py
Keller-1998: I've called it that way because it only checks for br and switch instructions as these have the crucial influence on feature-related control-flow decisions. Other terminator instructions are not considered. But yeah, probably `is_terminator` looks a bit prettier.

vulder: The checking behavior is ok for now, as currently only the two are relevant for us, I just wanted a more general name.

varats/varats/data/reports/feature_analysis_report.py
```diff
@@ -249,8 +249,10 @@ class FeatureAnalysisReportEval():
...
 
     def __init__(
-        self, fa_report: FeatureAnalysisReport,
-        ground_truth: FeatureAnalysisGroundTruth, features: tp.List[str]
+        self,
+        fa_report: FeatureAnalysisReport,
+        ground_truth: FeatureAnalysisGroundTruth,
+        features: tp.Optional[tp.List[str]] = None
     ) -> None:
         self.__initialize_eval_data(features)
```
vulder, line 255: Now, you moved the type missmatch to here :wink:
You have two choices now:
a) propagate the optional type further down the call chain
b) check if `features` is `None`, and if so, pass a new empty list to `__initialize_eval_data`

vulder (approved):

varats/varats/tables/feature_analysis_eval.py
```diff
@@ -0,0 +1,354 @@
...
+        ]
+
+        features: tp.List[str] = []
+        if self.table_kwargs['features'] is not None:
+            features = re.compile(r',\s*').split(self.table_kwargs['features'])
+
+        insts: int = 0
+        data: tp.List[pd.DataFrame] = []
+        binaries = case_study.project_cls.binaries_for_revision(
+            ShortCommitHash(str(cs_revisions[0]))
```
vulder: This should no longer be necessary with this patch https://github.com/se-sic/VaRA-Tool-Suite/pull/656

Could you please try if it works without the manual ShortCommitHash conversion :)

Keller-1998: Works!

PR #618, opened 2022-03-17
Rework table architecture

cormensratio:
This PR reworks how tables can be generated with the tool suite. 
The rework allows the generation of tables mostly identical to the generation of plots (https://github.com/se-sic/VaRA-Tool-Suite/pull/453), namely with the `click` CLI.

Related: se-sic/VaRA#628


tests/TEST_INPUTS/paper_configs/test_artefacts_driver/artefacts.yaml
```diff
@@ -13,7 +13,12 @@ artefacts:
   report_type: EmptyReport
   view: false
 - artefact_type: table
-  artefact_type_version: 1
+  artefact_type_version: 2
+  dry_run: false
```
vulder, line 17: Why is the `dry_run` flag saved here?

tests/table/test_blame_interaction_graph_table.py
```diff
@@ -31,7 +31,10 @@ def test_cig_metrics_table(self) -> None:
         load_paper_config()
 
         # latex booktabs is default format
-        table_str = CommitInteractionGraphMetricsTable().tabulate()
+        table_str = CommitInteractionGraphMetricsTable(
+            TableConfig.from_kwargs(view=False),
+            case_study=get_loaded_paper_config().get_all_case_studies()
```
vulder, line 36: This seems like a small miss match. `case_study = list_of_case_studies`

tests/table/test_case_study_metrics_table.py
```diff
@@ -21,18 +22,15 @@ def test_one_case_study_latex_booktabs(self) -> None:
 
         # latex booktabs is default format
         table_str = CaseStudyMetricsTable(
-            revisions={
-                "brotli":
-                    FullCommitHash("ce222e317e36aa362e83fc50c7a6226d238e03fd")
-            }
-        ).tabulate()
+            TableConfig.from_kwargs(view=False)
+        ).tabulate(TableFormat.LATEX_BOOKTABS, False)
```
vulder, line 26: Doesn't this make the test brittle? We pinned the brotli version for the table, so the stats would stay the same. In the new setup, I don't see how the results are stable now.

tests/table/test_case_study_metrics_table.py
```diff
@@ -47,11 +45,8 @@ def test_multiple_case_studies_latex_booktabs(self) -> None:
 
         # latex booktabs is default format
         table_str = CaseStudyMetricsTable(
-            revisions={
-                "xz":
-                    FullCommitHash("c5c7ceb08a011b97d261798033e2c39613a69eb7")
-            }
-        ).tabulate()
+            TableConfig.from_kwargs(view=False)
+        ).tabulate(TableFormat.LATEX_BOOKTABS, False)
```
vulder, line 49: same here

varats/varats/table/table.py
```diff
@@ -30,53 +33,79 @@ class Table(metaclass=TableRegistry):
...
-        self.__name = name
-        self.__format = TableFormat.LATEX_BOOKTABS
+    def __init__(self, table_config: TableConfig, **kwargs: tp.Any) -> None:
+        self.__table_config = table_config
         self.__saved_extra_args = kwargs
 
-    @property
-    def name(self) -> str:
+    @classmethod
+    def __init_subclass__(
```
vulder, line 41: Is this a pylint confusion or what am I missing here?

varats/varats/table/table.py
```diff
@@ -30,53 +33,79 @@ class Table(metaclass=TableRegistry):
...
         """
-        return self.__name
+        return "The following tables are available:\n  " + "\n  ".join(
+            list(Table.TABLES)
+        )
 
-    @property
-    def format(self) -> TableFormat:
+    @staticmethod
+    def get_class_for_table_type(table_type: str) -> tp.Type['Table']:
```
vulder, line 64: I always find these types strangely looking, maybe it would be good idea to introduce a type alias `TableTy = tp.Type['Table']` or something like this?

varats/varats/table/tables.py
```diff
@@ -41,251 +61,543 @@ class TableFormat(Enum):
...
+
+    OPTIONS: tp.List[CLIOptionTy]
+    """Table generator CLI Options (set automatically)."""
+
+    def __init__(self, table_config: TableConfig, **table_kwargs: tp.Any):
+        self.__table_config = table_config
+        self.__table_kwargs = table_kwargs
+
+    @classmethod
+    def __init_subclass__(
```
vulder, line 351: same here

varats/varats/tools/driver_plot.py
```diff
@@ -103,6 +104,8 @@ def main(context: click.Context, **kwargs: tp.Any) -> None:
 
     initialize_cli_tool()
     initialize_projects()
+    initialize_tables()
+    initialize_plots()
```
vulder, line 108: Why does driver_plot need tables to be loaded?

vulder (changes requested):
The PR was quite large, but actually not that difficult to understand :laughing: Knowing the plot impl the table one is actually quite straight forward

tests/TEST_INPUTS/paper_configs/test_artefacts_driver/artefacts.yaml
boehmseb: You could also argue for other options such as `view` whether they need to be stored. So we simply persist all options in the artefact file. Excluding them would require to distinguish yet another type of options, e.g., by filtering a hard-coded list of unwanted options. On a positive note, this way an artefact will not change when a default value changes.

tests/table/test_blame_interaction_graph_table.py
boehmseb: This misnomer stems from the CLI option name, which is named `--case-study`. Thanks to code reuse, this is used in many plots and tables and would need to be replaced everywhere; doable, but laborious.

tests/table/test_case_study_metrics_table.py
boehmseb: The table uses the newest revision from the case study, hence it is stable (that's also why the expected result changed below).

varats/varats/table/table.py
boehmseb: Apparently, pylint does not understand this correctly. We don't suppress this warning anywhere else, so I didn't bother here either.

varats/varats/table/table.py
boehmseb: I agree in principle, but since it has only this one usage, I think that giving it its own alias is a bit overkill.

varats/varats/tools/driver_plot.py
boehmseb: TL;DR: We always need to initialize plots and tables when we access the paper config because of artefacts.

Somewhere during construction of the CLI, the paper config is loaded so a plot can ask you to provide a valid case study as a parameter. When the paper config is loaded, it also loads the artefact file. The internal representation of an artefact stores which plot or table it can generate (and also sanity-checks the type's existance during loading). And since an artefact file can store plots _and_ tables we need to initialize both registries.

tests/TEST_INPUTS/paper_configs/test_artefacts_driver/artefacts.yaml
vulder: kk

tests/table/test_blame_interaction_graph_table.py
vulder: Yeah, seems currently not worth it :rofl: Maybe a clean up task for a rainy day

tests/table/test_case_study_metrics_table.py
vulder: I see, hmm. So from an functionality perspective, the table no longer is customization and only defaults to the current case study version. I guess that's ok, I did not understand the initial change like that.

varats/varats/table/table.py
vulder: Hmm, I never saw this error before, is it new?
The more important question: Is pylint correct here? I would assume not... what do you think?

varats/varats/table/table.py
vulder: Do you expect that other code will need this type?
If such cases are to be expected, providing a type def would make it easier and more consistent later on.

varats/varats/tools/driver_plot.py
vulder: Oh :laughing: 
If this is the case, we probably should do one of the following:
a) Remove the loop (hard)
b) provide an aggregated initializer function so the CLI only needs to call one and new usages don't fall into this tar pit^^

varats/varats/table/table.py
boehmseb: Pylint is incorrect here. What we do here is perfectly fine as you can see in the [official python docs](https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__). Mypy also flagged this at some point, but they have fixed this already.

varats/varats/table/table.py
boehmseb: I don't expect this. Currently, we don't even use this function (but that could change in the future).

varats/varats/tools/driver_plot.py
boehmseb: It's not really a loop, but I agree that we should improve how we initialize all of our registries. This, however, sounds like its own PR and we should discuss this first.

varats/varats/table/table.py
vulder: ok then let's keep it that way

vulder (approved):
Expect for the pylint warning in `varats/varats/ts_utils/cli_util.py`

LGTM

vulder (approved):

PR #621, opened 2022-03-19
Additional verification for VaRA in buildsetup

LuAbelt:
Added additional verification to the `verify_install` step for VaRA

Additionally, VaRA now runs the `check_vara` target to verify that it not only compiles but also passes unittests.

Not sure if we really want this in a separate `verify_build` but in `verify_install` we currently don't have access to the build directory in which we run `ninja`.

resolves https://github.com/se-sic/VaRA/issues/801

### DIFF
--- a/varats/varats/tools/research_tools/research_tool.py
+++ b/varats/varats/tools/research_tools/research_tool.py
@@ -492,6 +492,17 @@ class ResearchTool(tp.Generic[SpecificCodeBase]):
             True, if the tool was correctly installed
         """
 
+    def verify_build(
+        self, build_type: BuildType, build_folder_suffix: tp.Optional[str]
+    ) -> bool:
+        """
+        Verify if the research tool was built correctly for a given build_type.
+
+        Returns:
+            True, if the build was correct.
+        """
+        return True
+
     def container_install_dependencies(
         self, image_context: 'containers.BaseImageCreationContext'
     ) -> None:
--- a/varats/varats/tools/research_tools/research_tool.py
+++ b/varats/varats/tools/research_tools/research_tool.py
@@ -490,16 +490,21 @@ class ResearchTool(tp.Generic[SpecificCodeBase]):
             True, if the tool was correctly installed
         """
 
+    @abc.abstractmethod
     def verify_build(
         self, build_type: BuildType, build_folder_suffix: tp.Optional[str]
     ) -> bool:
         """
         Verify if the research tool was built correctly for a given build_type.
 
+        Args:
+            build_type: which type of build should be used, e.g., debug,
+                        development or release
+            build_folder_suffix: a suffix that is appended to the build folder
+
         Returns:
             True, if the build was correct.
         """
-        return True
 
     def container_install_dependencies(
         self, image_context: 'containers.BaseImageCreationContext'
--- a/varats/varats/tools/research_tools/vara.py
+++ b/varats/varats/tools/research_tools/vara.py
@@ -5,6 +5,7 @@ import math
 import os
 import re
 import shutil
+import subprocess
 import typing as tp
 from pathlib import Path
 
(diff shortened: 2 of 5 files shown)


varats/varats/tools/research_tools/research_tool.py
```diff
@@ -490,6 +490,17 @@ def verify_install(self, install_location: Path) -> bool:
...
+    def verify_build(
+        self, build_type: BuildType, build_folder_suffix: tp.Optional[str]
+    ) -> bool:
+        """
+        Verify if the research tool was built correctly for a given build_type.
+
+        Returns:
+            True, if the build was correct.
+        """
+        return True
```
vulder:
```suggestion
```

Should we really provide a default impl here? Could be better to force every `research_tool` to implement the method by making it abstract. If the tool wants to default to `true` they can.

varats/varats/tools/research_tools/research_tool.py
```diff
@@ -490,6 +490,17 @@ def verify_install(self, install_location: Path) -> bool:
             True, if the tool was correctly installed
         """
 
+    def verify_build(
+        self, build_type: BuildType, build_folder_suffix: tp.Optional[str]
+    ) -> bool:
+        """
+        Verify if the research tool was built correctly for a given build_type.
+
```
vulder, line 499:
```suggestion

            build_type: which type of build should be used, e.g., debug,
                        development or release
            build_folder_suffix: a suffix that is appended to the build folder
```

PS: not sure if this is correctly aligned, gh makes it hard so see this

varats/varats/tools/research_tools/vara.py
```diff
@@ -374,8 +374,58 @@ def verify_install(self, install_location: Path) -> bool:
         status_ok &= (install_location / "bin/opt").exists()
         status_ok &= (install_location / "bin/phasar-llvm").exists()
 
+        # Check that clang++ can display it's version
+
+        clang = local[str(install_location / "bin/clang++")]
+
+        (ret, stdout, _) = clang.run("--version")
```
vulder:
```suggestion
        # Check that clang++ can display it's version
        clang = local[str(install_location / "bin/clang++")]
        ret, stdout, _ = clang.run("--version")
```
Python does not require braces to unpack the return values

varats/varats/tools/research_tools/vara.py
```diff
@@ -374,8 +374,58 @@ def verify_install(self, install_location: Path) -> bool:
...
+
+        vara_name = self.code_base.get_sub_project("vara-llvm-project").name
+        status_ok &= (ret == 0)
```
vulder:
```suggestion
        status_ok &= (ret == 0)
```
What's the idea behind putting braces around the comparision?

varats/varats/tools/research_tools/vara.py
```diff
@@ -374,8 +374,58 @@ def verify_install(self, install_location: Path) -> bool:
...
+
+        status_ok &= (vara_name in stdout)
+
+        # Check that phasar-llvm can display it's version
+        phasar_llvm = local[str(install_location / "bin/phasar-llvm")]
+
+        (ret, stdout, _) = phasar_llvm.run("--version")
+
+        status_ok &= (ret == 0)
```
vulder:
```suggestion
        phasar_llvm = local[str(install_location / "bin/phasar-llvm")]
        ret, stdout, _ = phasar_llvm.run("--version")
        status_ok &= ret == 0
```

varats/varats/tools/research_tools/vara.py
```diff
@@ -374,8 +374,58 @@ def verify_install(self, install_location: Path) -> bool:
...
 
+    def verify_build(
+        self, build_type: BuildType, build_folder_suffix: tp.Optional[str]
+    ) -> bool:
+        """
+        Verifies whether vara was built correctly for the given target.
+
+        Args:
+            build_type: which type of build should be used, e.g., debug,
+                        development or release
```
vulder, line 403:
```suggestion
            build_type: which type of build should be used, e.g., debug,
                        development or release
            build_folder_suffix: a suffix that is appended to the build folder
```

varats/varats/tools/research_tools/vara.py
```diff
@@ -374,8 +374,58 @@ def verify_install(self, install_location: Path) -> bool:
...
+            return False
+
+        build_folder_path = build_type.build_folder(build_folder_suffix)
+        full_path /= build_folder_path
+
+        ninja = local["ninja"].with_cwd(full_path)
+
+        (ret, _, _) = ninja.run("check-vara")
+
+        return bool(ret == 0)
```
vulder:
```suggestion
        ninja = local["ninja"].with_cwd(full_path)
        ret, _, _ = ninja.run("check-vara")
        return ret == 0
```

vulder (changes requested):
Impl looks good, thx.

I only have a few small issue/questions.

varats/varats/tools/research_tools/vara.py
LuAbelt: If we remove the explicit cast to `bool`, mypy will complain about the function returning `Any` even though it was declared as `bool`

boehmseb (approved):

vulder (approved):

PR #623, opened 2022-03-20
Removed check_required_args

LuAbelt:
Remove definition and usages of `@check_required_args`

Closes https://github.com/se-sic/VaRA/issues/746

### DIFF
--- a/varats/varats/tools/research_tools/vara.py
+++ b/varats/varats/tools/research_tools/vara.py
@@ -12,7 +12,6 @@ from benchbuild.utils.cmd import ln, mkdir
 from plumbum import local
 from PyQt5.QtCore import QProcess
 
-from varats.plot.plot_utils import check_required_args
 from varats.tools.research_tools.cmake_util import set_cmake_var
 from varats.tools.research_tools.research_tool import (
     CodeBase,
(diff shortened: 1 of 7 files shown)


boehmseb (changes requested):
Can you please make the two variables that were previously checked  for the function `setup()` in the reasearch tools proper arguments? We can then get rid of the `kwargs` there.

The rest looks good.

vulder (approved):

varats/varats/tools/research_tools/vara.py
```diff
@@ -219,7 +218,7 @@ def setup(self, source_folder: tp.Optional[Path], **kwargs: tp.Any) -> None:
 
         self.code_base.clone(self.source_location())
         self.code_base.setup_vara_remotes()
-        self.code_base.checkout_vara_version(version, use_dev_branches)
+        self.code_base.checkout_vara_version(int(version), use_dev_branches)
```
boehmseb: you might need to create a new variable of type `int` and assign to it in lines 207 and 209 to avoid this type error.

vulder: ~~Or maybe wrap `int()` around the config load in line 212~~

boehmseb (approved):

PR #631, opened 2022-03-28
report aggregates for time and TEF

jonas-kaufmann:


varats-core/varats/report/gnu_time_report.py
```diff
@@ -20,18 +20,21 @@
...
 from varats.utils.util import static_vars
 
 
 class WrongTimeReportFormat(Exception):
-    """Thrown if the a time report could not be parsed."""
+    """Thrown if a time report could not be parsed."""
 
 
-class TimeReport(BaseReport, shorthand="TR", file_type=""):
+class TimeReport(BaseReport, shorthand="TR", file_type="txt"):
```
vulder, line 37: Why do you think we should a txt as a file type for the time reports?

varats-core/varats/report/gnu_time_report.py
```diff
@@ -197,3 +200,38 @@ def _parse_max_resident_size(line: str) -> int:
...
+):
+    """Context Manager for parsing multiple time reports stored inside a zip
+    file."""
+
+    def __init__(self, path: Path) -> None:
+        super().__init__(path, TimeReport)
+
+    @property
+    def wall_clock_times(self) -> tp.List[float]:
+
```
vulder:
```suggestion
```

varats-core/varats/report/gnu_time_report.py
```diff
@@ -197,3 +200,38 @@ def _parse_max_resident_size(line: str) -> int:
...
+        ]
+
+    @property
+    def wall_clock_time_mean(self) -> float:
+
+        return np.mean(self.wall_clock_times)
+
+    @property
+    def wall_clock_time_std(self) -> float:
+
```
vulder:
```suggestion
```

varats-core/varats/report/gnu_time_report.py
```diff
@@ -197,3 +200,38 @@ def _parse_max_resident_size(line: str) -> int:
...
+    @property
+    def wall_clock_times(self) -> tp.List[float]:
+
+        return [
+            report.wall_clock_time.total_seconds() for report in self.reports
+        ]
+
+    @property
+    def wall_clock_time_mean(self) -> float:
+
```
vulder:
```suggestion
```

varats-core/varats/report/gnu_time_report.py
```diff
@@ -197,3 +200,38 @@ def _parse_max_resident_size(line: str) -> int:
...
+    @property
+    def wall_clock_times(self) -> tp.List[float]:
+
+        return [
+            report.wall_clock_time.total_seconds() for report in self.reports
+        ]
+
+    @property
+    def wall_clock_time_mean(self) -> float:
```
vulder:
```suggestion
    def mean_wall_clock_time(self) -> float:
```
I think this would read better.

```
df = ...
report = ...
df["mean"] = report.mean_wall_clock_time
```

varats-core/varats/report/gnu_time_report.py
```diff
@@ -197,3 +200,38 @@ def _parse_max_resident_size(line: str) -> int:
...
+
+        return np.mean(self.wall_clock_times)
+
+    @property
+    def wall_clock_time_std(self) -> float:
```
vulder: same here

varats-core/varats/report/report.py
```diff
@@ -557,3 +560,46 @@ def __contains__(self, report_type: tp.Type[BaseReport]) -> bool:
...
     def __iter__(self) -> tp.Iterator[tp.Type[BaseReport]]:
         return iter(self.report_types)
+
+
+T = tp.TypeVar('T', bound=BaseReport)
+
+
+class ReportAggregate(
+    BaseReport, tp.Generic[T], shorthand="Agg", file_type="zip"
+):
```
vulder, line 570: I'm still not fully convinced we should implement the Aggregated report with a context manager, as the typical report usage pattern in varats does not use `with`, but just hands report objects from `a` to `b`.

In my opinion, using a finalizer to clean up the tmpdir after object destruction is the better way.
See the example here https://docs.python.org/3.6/library/weakref.html#comparing-finalizers-with-del-methods

varats-core/varats/report/report.py
```diff
@@ -557,3 +560,46 @@ def __contains__(self, report_type: tp.Type[BaseReport]) -> bool:
 
     def __iter__(self) -> tp.Iterator[tp.Type[BaseReport]]:
         return iter(self.report_types)
+
+
+T = tp.TypeVar('T', bound=BaseReport)
+
+
+class ReportAggregate(
+    BaseReport, tp.Generic[T], shorthand="Agg", file_type="zip"
```
vulder: @boehmseb Can you please take a look at the typing here, I'm not 100% sure how mypy likes typechecking when using run-time types (parameters) as generic `T`s.

tests/report/test_aggregated_reports.py
```diff
@@ -0,0 +1,65 @@
...
+            tmp_file = Path(tmp_dir) / "TimeAggregateTwoReportsTest.zip"
+            time_aggregate = TimeReportAggregate(tmp_file)
+            with ZippedReportFolder(tmp_file) as time_reports_dir:
+
+                for i in range(num_reports):
+
+                    with open(
+                        Path(time_reports_dir) / f"time_report_{i}.txt", "w"
+                    ) as time_report_file:
+
```
vulder:
```suggestion
```

tests/report/test_aggregated_reports.py
```diff
@@ -0,0 +1,65 @@
...
+        with tempfile.TemporaryDirectory() as tmp_dir:
+
+            # Write time reports to `tempdir`.
+            tmp_file = Path(tmp_dir) / "TimeAggregateTwoReportsTest.zip"
+            time_aggregate = TimeReportAggregate(tmp_file)
+            with ZippedReportFolder(tmp_file) as time_reports_dir:
+
+                for i in range(num_reports):
+
```
vulder:
```suggestion
```

tests/report/test_aggregated_reports.py
```diff
@@ -0,0 +1,65 @@
...
+    """Test if we can write time reports to `AggregatedReport.tempdir`, zip the
+    contents and read them again afterwards."""
+
+    def test_two_equal_reports(self) -> None:
+        """Test if we correctly parse duration event types."""
+
+        num_reports = 2
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+
```
vulder:
```suggestion
```

tests/report/test_aggregated_reports.py
```diff
@@ -0,0 +1,65 @@
...
+            # Write time reports to `tempdir`.
+            tmp_file = Path(tmp_dir) / "TimeAggregateTwoReportsTest.zip"
+            time_aggregate = TimeReportAggregate(tmp_file)
+            with ZippedReportFolder(tmp_file) as time_reports_dir:
+
```
vulder:
```suggestion
```

vulder (changes requested):
Overall, this looks good, will be a nice addition to the tool suite :smile: , we just need to figure out the thing about the context mgr for finaliztion.

varats-core/varats/report/report.py
boehmseb: Not sure where you see a potential problem. Mypy does not complain on this PR and I also can't see any problems.

vulder: Yeah, it does not complain that's what makes me anxious :laughing: 
With dynamic types for `T` you cannot really do a lot of type checking at mypy time and I was wondering if that could lead to problems in certain scenarios.

boehmseb: From what I see, this should be fine. We do similar things in other locations. A better name for `T` would be nice though.

vulder (approved):

PR #632, opened 2022-04-11
Set  mounting parameters according to the Buildah version when building Containers

Sinerum:
Set the `rw` mounting parameter when the installed buildah version requires it and not if it does not. Also update the required benchbuild version to 3.6.1. to include this fix in benchbuild as well.

### DIFF
--- a/varats/varats/containers/containers.py
+++ b/varats/varats/containers/containers.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from tempfile import TemporaryDirectory
 
 from benchbuild.environments import bootstrap
+from benchbuild.environments.adapters.common import buildah_version
 from benchbuild.environments.domain.commands import (
     CreateImage,
     fs_compliant_name,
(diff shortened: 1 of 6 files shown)


boehmseb (approved):

boehmseb:
Only this one test needs fixing. It just needs to include the `,rw` in the expected output.

Sinerum:
> Only this one test needs fixing. It just needs to include the `,rw` in the expected output.

It needs to be removed, but that would mean the test would fail depending on the buildah version would that be okay?

varats/varats/containers/containers.py
```diff
@@ -60,7 +61,7 @@ def distro(self) -> Distro:
             'apt', 'install', '-y', 'wget', 'gnupg', 'lsb-release',
             'software-properties-common', 'python3', 'python3-dev',
             'python3-pip', 'musl-dev', 'git', 'gcc', 'libgit2-dev',
-            'libffi-dev', 'libyaml-dev'
+            'libffi-dev', 'libyaml-dev', 'graphviz-dev'
```
boehmseb, line 64: Can you please also add this dependency to the install guide in `docs/source/tutorials/getting_started.rst` and `docs/source/vara-ts/vara-buildsetup.rst`?

boehmseb (changes requested):

vulder:
> > Only this one test needs fixing. It just needs to include the `,rw` in the expected output.
> 
> It needs to be removed, but that would mean the test would fail depending on the buildah version would that be okay?

Neither, just use the buildah version check to determine if `rm` should be there or not.

vulder (approved):
Except for the missing test adaption this PR looks good to me.

PR #633, opened 2022-04-11
Casestudy Generation Gui

Sinerum:
Adds a gui for generating casestudies. the gui can be started using `vara-cs gen-gui`

resolves se-sic/VaRA#795


varats/varats/gui/cs_gen/main_window.py
```diff
@@ -0,0 +1,353 @@
...
+    ShortCommitHash,
+    create_commit_lookup_helper,
+    FullCommitHash,
+)
+from varats.utils.settings import vara_cfg
+
+
+class GenerationStrategie(Enum):
+    """Enum for the Startegie used when Generating a CaseStudy."""
+    SELECTREVISION = 0
```
vulder:
```suggestion
    SELECT_REVISION = 0
```
To follow the naming schema, this needs an `_` or?

vulder (changes requested):
Did you write the gui without a ui file? If not, you should add the ui file for generation to the appropriate folder (`uicomponents/`) and add the generation call to the script.

Furthermore, I think we should name the `main_window` files a bit more specific.

Otherwise, the GUI looks nice, thx.

vulder (approved):
thx, lgtm now :grinning:

varats/varats/tools/driver_casestudy.py
```diff
@@ -14,6 +14,7 @@
 from varats.data.discover_reports import initialize_reports
 from varats.data.reports.szz_report import SZZReport
 from varats.experiment.experiment_util import VersionExperiment
+from varats.gui.cs_gen.case_study_generation import start_gui
```
boehmseb: We should be careful with imports that require Qt since as soon as a module imports `PyQt5` it will not work in our containers. That means that with this import, we can no longer run any `vara-cs` commands in a container. As far as I know we don't do this now anywhere maybe apart from manual debugging, but who knows what will happen in the future.
Maybe making this a local import could avoid this issue, but I'm not certain.
@vulder what do you think?

boehmseb (commented):
Do we have any chance of testing anything of this, e.g., the filter logic or similar?

varats/varats/tools/driver_casestudy.py
vulder: Yeah, that could become a problem, we probably should also include a test at some point to ensure we don't break anything on accident.

For this specific case, maybe we should split of the `gui` tools into separate drivers? This would at least make it clear on which qt may depend.

Far fetched idea for the future: Introduce a new namespace `varats-gui`, which implements all `gui` tools and gui related stuff AND only the setup.py for this namespace is allowed to depend on `qt`.

boehmseb: For now i vote for the separate driver module. I think we did (or at least discussed) something similar with vara-buildsetup.
The namespace idea sounds like something that should be done after our long desired jupyter idea as it would require the same kind of API layer.

vulder: Works for me. Then let's move the code into a separated driver module and talk about the rework afterwards :+1:

@Sinerum If you're interested in joining out "new and fancy idea", we could meet in the following weeks in our office to discuss more details :wink:

boehmseb (approved):

PR #638, opened 2022-05-02
Remove extra_test.sh from ci

Sinerum:
resolves se-sic/VaRA#853


tests/tools/test_driver_casestudy.py
```diff
@@ -185,6 +185,94 @@ def test_vara_cs_gen_specific(self):
...
+                '6c6da57ae2aa962aabde6892442227063d87e88c'
+            ]
+        )
+        self.assertEqual(0, result.exit_code, result.exception)
+        case_study_path = Path(
+            vara_cfg()["paper_config"]["folder"].value +
+            "/test_ext/xz_0.case_study"
+        )
+        self.assertTrue(case_study_path.exists())
+        case_study = load_case_study_from_file(case_study_path)
```
vulder, line 225: Just to make sure, The case study had one revision already did it not?
Because you check for two different ones afterwards but only select one.

What could improve this test is add two constants at the top of the test func `new_commit = ...` and `already_existing_commit = ...` and the afterwards use the constants.

tests/tools/test_driver_casestudy.py
```diff
@@ -185,6 +185,94 @@ def test_vara_cs_gen_specific(self):
...
+        self.assertEqual(1, case_study.num_stages)
+
+    @run_in_test_environment(
+        UnitTestFixtures.create_file_fixture(
+            TEST_INPUTS_DIR / "paper_configs/test_casestudy_status",
+            Path("paper_configs/test_ext")
+        )
+    )
+    def test_vara_cs_gen_to_extend_new_stage(self):
+        """Test the extend-functionality of vara-cs gen."""
```
vulder, line 241: same for this test

vulder (changes requested):
Overall, this looks good thx for remove the old "extra tests" :laughing:

tests/tools/test_driver_casestudy.py
Sinerum: Yes since the test is about the extending feature we need to make sure the original commits are still present.

vulder: Yeah, makes sense :+1:

vulder (approved):

boehmseb (approved):

PR #639, opened 2022-05-03
Make contains_source_code work on initial commits

Sinerum:
change the` contains_source_code` function to handle initial commits of a repository properly

### DIFF
--- a/varats-core/varats/utils/git_util.py
+++ b/varats-core/varats/utils/git_util.py
@@ -282,8 +282,8 @@ def contains_source_code(
     if not churn_config:
         churn_config = ChurnConfig.create_c_style_languages_config()
 
-    return_code = git[__get_git_path_arg(repo_folder), "diff", "--exit-code",
-                      "--quiet", f"{commit.hash}~", commit.hash, "--",
+    return_code = git[__get_git_path_arg(repo_folder), "show", "--exit-code",
+                      "-m", "--quiet", commit.hash, "--",
                       churn_config.get_extensions_repr('*.')] & RETCODE
 
     if return_code == 0:


boehmseb (approved):

varats-core/varats/utils/git_util.py
```diff
@@ -282,8 +282,8 @@ def contains_source_code(
     if not churn_config:
         churn_config = ChurnConfig.create_c_style_languages_config()
 
-    return_code = git[__get_git_path_arg(repo_folder), "diff", "--exit-code",
-                      "--quiet", f"{commit.hash}~", commit.hash, "--",
+    return_code = git[__get_git_path_arg(repo_folder), "show", "--exit-code",
+                      "-m", "--quiet", commit.hash, "--",
```
vulder, line 286:
```suggestion
                      "-m", "--quiet", commit.hash, "--",
```
What does `-m` do? I cannot find this in my git docs.
Is it the same as `-M`? Maybe, we could use the log parameter form, if case one exists.

vulder (approved):

varats-core/varats/utils/git_util.py
Sinerum: `-m` makes merge commits show a diff. See https://git-scm.com/docs/git-show#Documentation/git-show.txt--m .
It is a short hand for  `--diff-merges=on`

vulder: Ok, then we just just use `--diff-merges=on` in the code, otherwise one needs the docs to understand the part.

PS: is it required to have the same behavior as diff? I would assume you added it for a reason, so show would be output similar to diff without it.

Sinerum: I just reread the documentation and with git version 2.30.2 which is the current version in Debians packagemanager, the only way to have merge commits shown correctly is using `-m` which generates a diff for each parrent.
[Here](https://git-scm.com/docs/git-show/2.30.0#_combined_diff_format) is the correct documentation for that version.

Since this option stays consitent between git versions we probably need to use it.

vulder: Ok, then we'll keep it like this.

PR #640, opened 2022-05-06
Add instrumentation point printer experiment

jonas-kaufmann:
Implements experiment which collects the source code locations of instrumentation points of VaRA's feature regions.

Requires se-sic/VaRA#875.

### DIFF
--- a/varats/varats/experiments/vara/instrumentation_printer_experiment.py
+++ b/varats/varats/experiments/vara/instrumentation_printer_experiment.py
@@ -1,0 +1,139 @@
+"""Implements experiment which collects the source code locations of
+instrumentation points of VaRA's feature regions."""
+
+import typing as tp
+from pathlib import Path
+
+from benchbuild import Project
+from benchbuild.extensions import compiler, run, time
+from benchbuild.utils import actions
+from benchbuild.utils.cmd import opt
+
+from varats.data.reports.empty_report import EmptyReport
+from varats.experiment.experiment_util import (
+    VersionExperiment,
+    ExperimentHandle,
+    exec_func_with_pe_error_handler,
+    get_default_compile_error_wrapped,
+    create_default_analysis_failure_handler,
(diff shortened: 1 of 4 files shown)


varats/varats/experiments/vara/instrumentation_printer_experiment.py
```diff
@@ -0,0 +1,139 @@
...
+    NAME = "CollectInstrumentationPoints"
+    DESCRIPTION = "Runs utility pass on LLVM-IR to extract instrumentation point information."
+
+    def __init__(self, project: Project, experiment_handle: ExperimentHandle):
+        super().__init__(obj=project, action_fn=self.analyze)
+        self.__experiment_handle = experiment_handle
+
+    def analyze(self) -> actions.StepResult:
+        if not self.obj:
+            return actions.StepResult.ERROR
```
vulder:
```suggestion
```
This is actually no longer needed, as BB guarantees this.

varats/varats/experiments/vara/instrumentation_printer_experiment.py
```diff
@@ -0,0 +1,139 @@
...
+
+            # Store the collected information in report.
+            run_cmd = opt[opt_params] > str(
+                vara_result_folder / str(result_file)
+            )
+
+            exec_func_with_pe_error_handler(
+                run_cmd,
+                create_default_analysis_failure_handler(
+                    self.__experiment_handle, project, EmptyReport,
```
vulder: Maybe, we should add a special report type for this kind of output.
If you want the quick option, just add a new report class declaration with a "speaking" class name similar to EmptyReport, with this, we at least have a unique shorthand.
If we want to impl the parsing later, we can always add it when needed.

vulder (changes requested):

vulder (approved):
Looks good, clean up the pylint remarks and we'll merge this :)

LGTM

PR #642, opened 2022-05-08
USDT execution stats experiment, projects gzip and bzip2

jonas-kaufmann:
- New experiment 'VaraUS', which captures tracing execution statistics using VaRA's USDT instrumented USDT probes and the included script 'UsdtExecutionStats.bt'
- Fixed bug in gzip, which always executes local machine's binary
- Added project bzip2
- Updated project FeaturePerfCSCollection to include binaries 'SimpleSleepLoop' and 'SimpleBusyLoop'

Example output of experiment:
```
Attaching 8 probes...

@nanoseconds_between_probes: 
[256, 512)             1 |                                                    |
[512, 1K)              0 |                                                    |
[1K, 2K)               0 |                                                    |
[2K, 4K)               3 |                                                    |
[4K, 8K)               1 |                                                    |
[8K, 16K)              0 |                                                    |
[16K, 32K)             0 |                                                    |
[32K, 64K)           522 |                                                    |
[64K, 128K)        98550 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[128K, 256K)         915 |                                                    |
[256K, 512K)          13 |                                                    |

@num_probes_fired[finalize]: 1
@num_probes_fired[init]: 1
@num_probes_fired[end]: 2
@num_probes_fired[begin]: 100002
```

resolves se-sic/VaRA#884


varats-core/varats/report/gnu_time_report.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, path: Path) -> None:
                         TimeReport._parse_wall_clock_time(line)
                     continue
 
-                print("Not matched: ", line)
+                # print("Not matched: ", line)
```
vulder, line 69:
```suggestion
                print("Not matched: ", line)
```
Left over comment or should we remove this line?

varats/varats/data/reports/usdt_stats_report.py
```diff
@@ -0,0 +1,8 @@
+"""Implements report for VaRA's bpftrace script 'UsdtExecutionStats.bt'."""
+
+from varats.report.report import BaseReport
+
+
+class VaraUsdtStatsReport(BaseReport, shorthand="VaraUS", file_type="txt"):
```
vulder: I'm currently debating in my head what a better name would be.
UsdtStats is correct, but I think we should be a bit more specific and not focus on the technology to much.
What are your ideas? Maybe something in the direction of `InstStatistics`, not sure :thinking:

varats/varats/projects/c_projects/gzip.py
```diff
@@ -91,18 +91,23 @@ def compile(self) -> None:
         """Compile the project."""
         gzip_version_source = local.path(self.source_of_primary)
 
+        # Build binaries in separate dir because executing the binary with path 'gzip' will execute '/usr/bin/gzip' independent of the current working directory.
+        mkdir("-p", gzip_version_source / "build")
+
```
vulder: Good catch :+1:

PS: could it be that the line of the comment is to long?

varats/varats/projects/c_projects/gzip.py
```diff
@@ -91,18 +91,23 @@ def compile(self) -> None:
...
 
-        clang = bb.compiler.cc(self)
         with local.cwd(gzip_version_source):
-            with local.env(CC=str(clang)):
-                bb.watch(local["./bootstrap"])()
-                bb.watch(local["./configure"])()
+            bb.watch(local["./bootstrap"])()
+
+        clang = bb.compiler.cc(self)
+        with local.cwd(gzip_version_source / "build"), local.env(CC=str(clang)):
```
vulder:
```suggestion
        c_compiler = bb.compiler.cc(self)
        with local.cwd(gzip_version_source / "build"), local.env(CC=str(c_compiler)):
```
When we already hear, fixing this we can clean up the old "bad" name :laughing:

varats/varats/experiments/vara/usdt_stats_experiment.py
```diff
@@ -0,0 +1,160 @@
...
+
+        vara_result_folder = get_varats_result_folder(project)
+        binary: ProjectBinaryWrapper
+
+        for binary in project.binaries:
+            if binary.type != BinaryType.EXECUTABLE:
+                continue
+
+            # Get workload to use.
+            # TODO Change this.
```
vulder:
```suggestion
            # TODO (se-sic/VaRA#841): refactor to bb workloads if possible
```

varats/varats/experiments/vara/usdt_stats_experiment.py
```diff
@@ -0,0 +1,160 @@
...
+            workload_provider = WorkloadProvider.create_provider_for_project(
+                project
+            )
+            workload = workload_provider.get_workload_for_binary(binary.name)
```
vulder: You need to first check if a provider was available, for some projects it might not be :)

varats/varats/experiments/vara/usdt_stats_experiment.py
```diff
@@ -0,0 +1,160 @@
...
+
+                # Assertion: Can be run without sudo password prompt. To guarentee this, add an entry to /etc/sudoers.
+                bpftrace_cmd = bpftrace["-o", report_file, bpftrace_script,
+                                        binary.path]
+
+                bpftrace_runner: Future
+                with local.as_root():
+                    bpftrace_runner = bpftrace_cmd & BG
+
+                sleep(1)  # give bpftrace time to start up
```
vulder: Is there another way we could guarantee this?

What happens when 1 sec is not enough? Could happen on an old and slow machine.

varats-core/varats/provider/workload/workload_provider.py
```diff
@@ -0,0 +1,81 @@
...
+        ],
+        f"{Bzip2.NAME},bzip2": [
+            "--compress", "--best", "--verbose", "--keep", "--force",
+            "/home/jonask/Repos/WorkloadsForConfigurableSystems/compression/countries-land-1m.geo.json"
+        ],
+        f"{Gzip.NAME},gzip": [
+            "--force", "--keep", "--name", "--recursive", "--verbose", "--best",
+            "/home/jonask/Repos/WorkloadsForConfigurableSystems/compression/countries-land-10km.geo.json"
+        ],
+    }
```
vulder: We could slightly refactor this to make it "a bit" more reusable.
Put a option into the varats config `workload_base_location`, which you can set to `/home/jonask/Repos/WorkloadsForConfigurableSystems`. Then in the provider we concatenate the location on the rest of the path. With this, we have the option that you send you workloads to me/or someone an still run the experiment ;)

varats-core/varats/provider/workload/workload_provider.py
```diff
@@ -0,0 +1,81 @@
...
+            a default provider instance
+        """
+        raise AssertionError(
+            "All usages should be covered by the project specific provider."
+        )
+
+    def get_workload_for_binary(self, binary_name: str) -> tp.List[str]:
+        """Get a list of arguments to execute the given binary with."""
+        key = f"{self.project.NAME},{binary_name}"
+        return self.WORKLOADS.get(key, None)
```
vulder, line 88: You have two options here:
a) return an empty list
b) return an optional list (however, I'm not sure if `Provider` requires the signature)

vulder (changes requested):
Very nice work :+1:

varats-core/varats/report/gnu_time_report.py
jonas-kaufmann: I'd say let's remove it. For example, when parsing 100 time reports, stdout will be full of 'Not matched'.

vulder (approved):

PR #644, opened 2022-05-19
Fixes `dataframe_to_table()` default case

jonas-kaufmann:


boehmseb (approved):
The PR name does not completely fit. Something like "Fix dataframe tabulation default case" fits better. But the change looks good.

vulder (approved):

PR #650, opened 2022-06-03
Fix our container-ready slurm script template.

boehmseb:
The previous template did not handle cleanup correctly.

### DIFF
--- a/varats/varats/tools/templates/slurm_container.sh.inc
+++ b/varats/varats/tools/templates/slurm_container.sh.inc
@@ -43,15 +43,18 @@ exec 2>&1
 
 # parent of node dir (= BB_SLURM_NODE_DIR) is used as home
 node_dir=$(dirname {{ prefix }})
+lockfile="${node_dir}.lock"
+clean_lockfile="${node_dir}.clean-in-progress.lock"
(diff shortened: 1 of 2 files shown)


varats/varats/tools/templates/slurm_container.sh.inc
```diff
@@ -43,15 +43,18 @@ exec 2>&1
 
 # parent of node dir (= BB_SLURM_NODE_DIR) is used as home
 node_dir=$(dirname {{ prefix }})
+lockfile="${node_dir}.lock"
+clean_lockfile="${node_dir}.clean-in-progress.lock"
+clean_lockdir="${node_dir}"
 
 # Prepare cluster node
-exec 9> {{ lockfile }}
+exec 9> ${lockfile}
```
vulder, line 51: Don't we also need the `mkdir` for `clean_lockdir`?

boehmseb: No because lines 46-48 guarantee that `clean_lockdir` is the same as `node_dir`.
I only kept them separate to stay closer to BenchBuild's template.

vulder (approved):

PR #652, opened 2022-06-04
adds feature performance analysis experiments

jonas-kaufmann:


jonas-kaufmann:
The newly added `FeaturePerfAnalysisTef` experiment should cover the functionality from [`varats/varats/experiments/vara/feature_perf_runner.py`](https://github.com/se-sic/VaRA-Tool-Suite/blob/vara-dev/varats/varats/experiments/vara/feature_perf_runner.py) with the difference of collecting gnu time statistics. What's your opinion, should I remove it?

varats-core/varats/experiment/experiment_util.py
```diff
@@ -307,15 +307,20 @@ def get_file_name(
...
-            self.__experiment.shorthand(), project_name, binary_name,
+            self.experiment.shorthand(), project_name, binary_name,
             project_revision, project_uuid, extension_type
         )
 
+    @property
+    def experiment(self) -> 'VersionExperiment':
+        """Experiment this handle is for."""
+        return self._experiment
+
```
vulder: We should not export the `Experiment` here. The idea behind the handle was to control which information can be accessed by a step and how. The `Experiment` has many details that we don't want to expose to the underlying steps or a project.
I checked your usages below, you only need the `shorthand`, which would be ok to export (i.e., make accessible).
So, I would remove this part here and only add a property for the `shorthand`.

varats/varats/experiments/vara/feature_perf_analysis_experiments.py
```diff
@@ -0,0 +1,277 @@
...
+class FeaturePerfAnalysisDry(VersionExperiment, shorthand="FPA_Dry"):
+    """Captures baseline runtime (without any instrumentation)."""
+
+    NAME = "FeaturePerfAnalysisDry"
+
+    REPORT_SPEC = ReportSpecification(TimeReportAggregate, TEFReport)
+
+    # Indicate whether to trace binaries and whether USDT markers should be used
+    TRACE_BINARIES = False
+    USE_USDT = False
```
vulder: I'm currently a bit undecided if modeling these two values as class variables is a good idea. I normally only use object variables for these this and pass the correct state via the `__init__` method.

What was you reasoning behind it?

vulder (changes requested):
The overall Idea is good, we should talk about this a bit more in our 1o1 tomorrow, as I also was working on unifying feature perf analyses. Maybe we can merge/extend your abstraction to handle everything.

varats-core/varats/experiment/experiment_util.py
jonas-kaufmann: Makes sense, going to change that.

varats/varats/experiments/vara/feature_perf_analysis_experiments.py
jonas-kaufmann: Actually don't know the reason anymore. Guess it was somehow simpler to do it with class variables in the past but I then removed the reason for it. Gonna change that since in my opinion passing the parameters in `__init__` is the proper way to do it.

PR #661, opened 2022-07-07
Add containers with older flex versions for older revisions of doxygen

Sinerum:
depends on PolyJIT/benchbuild#554
resolves se-sic/VaRA#536


varats/varats/projects/cpp_projects/doxygen.py
```diff
@@ -28,26 +29,43 @@ class Doxygen(VProject):
...
+                 (
+                     GoodBadSubgraph([
+                         "093381b3fc6cc1e97f0e737feca04ebd0cfe538d"
+                     ], ["cf936efb8ae99dd297b6afb9c6a06beb81f5b0fb"],
+                                     "Needs flex <= 2.5.4 and >= 2.5.33"),
+                     get_base_image(ImageBase.DEBIAN_10).run(
+                         'apt', 'install', '-y', 'cmake', 'flex-old', 'bison',
+                         'qt5-default'
+                     )
+                 )]
```
vulder: Nice, that you found this.
Just one question: Do these ranges not overlap? That looks strange that they would have a gap where flex-old was no longer required.

vulder: @Sinerum ping

vulder: @Sinerum ping

Sinerum: The starting commits are on two different branches the ending commit is after they have been merged. I'm not sure why I put it in two SubGraphs. I think it should work with one.

vulder: kk, if we get this cleaned up, we'll can merge this fix :)

varats-core/varats/paper/paper_config.py
```diff
@@ -260,8 +260,7 @@ def __init__(
 
     def versions(self) -> tp.List[bb.source.base.Variant]:
         cache_path = self.fetch()
-        git_rev_list = git['rev-list', '--abbrev-commit', '--abbrev=10',
-                           '--all']
+        git_rev_list = git['rev-list', '--all']
```
vulder: Why did you change this here? We, similar to benchbuild, us 10 chars long short commit hashes as the default.

Sinerum: Because they don't use short hashes for their revision range list and this version string is checked against the revision range when looking for the correct container to use. As seen in the mentioned PR. This caused no container to be found. So either the benchbuild revision range needs to use the short version or we need a more verbose system to check for eqaulity similar to ours.

vulder: A simpler way of checking equality would be nice but we definitively need a benchbuild patch here as changing the length on our side will not work and break a lot of our use cases.

Sinerum: Okay then I will add that to the PR on benchbuild and revert this

vulder:
Does this PR fix an issues? If yes, please not the issue number in the top comment of the PR

boehmseb (approved):
Can be merged once the formatting is fixed.

varats/varats/projects/cpp_projects/doxygen.py
```diff
@@ -28,26 +29,34 @@ class Doxygen(VProject):
...
+                     GoodBadSubgraph([
+                         "a6238a4898e20422fe6ef03fce4891c5749b1553",
+                         "093381b3fc6cc1e97f0e737feca04ebd0cfe538d"
+                     ], ["cf936efb8ae99dd297b6afb9c6a06beb81f5b0fb"],
+                                     "Needs flex <= 2.5.4 and >= 2.5.33"),
+                     get_base_image(ImageBase.DEBIAN_10).run(
+                         'apt', 'install', '-y', 'cmake', 'flex-old', 'bison',
+                         'qt5-default'
+                     )
+                 )]
```
boehmseb: For some reason, github did not include that in my review:

The indentation here is very misleading. If yapf cannot be convinced to layout this better, we should do it by hand and disable yapf for this block.

vulder (approved):

vulder (approved):

PR #663, opened 2022-07-18
Added fast_downward project

bnico99:


vulder:
@bnico99  @boehmseb  Did you two discuss the issue regarding version exploration (`--all`)?
What is the current state of this discussion/this PR?

boehmseb:
> @bnico99 @boehmseb Did you two discuss the issue regarding version exploration (`--all`)? What is the current state of this discussion/this PR?

We only discussed the issue regarding annotated and lightweight tags.

bnico99:
@vulder @boehmseb I added the `--all` parameter since otherwise, the release commits of FastDownward are not in the commit list. Example:

```
git --no-pager log --pretty=format:'%H %D' | grep e907baf
git --no-pager log --all --pretty=format:'%H %D' | grep e907baf
e907baf9b847c28710873077b4670aa9e5310d57 tag: release-21.12.0, origin/release-21.12
```

If I understand it correctly, this is due to the fact that the release commits are not located in the `main` branch but in separate `release-*` branches.

Do you see any problems with adding the `--all` parameter?

vulder:
When sampling and finding revisions to "experiment on" this would introduce a lot of (potentially) broken commits. I'm not sure if we should enable this `by default`. However, we could extend the function in a way that makes this configurable for the use, like a `include_all_revs` bool parameter with an `false` default value.

@boehmseb What is your opinion?

boehmseb:
An extra option defaulting to `False` sounds reasonable.

boehmseb:
The problem with adding such a parameter is that the id we generate for commits is then no longer unique.
Also, existing case studies use the current commit map logic and will break when this changes.

I think the best solution is the following:
- Add the `--all` to the commit map generation.
- Add a function that creates a view on the commit map that only shows the commits that are shown with the old behavior, but still uses the new ids. We can use this function when we only want to see commits from the master branch, e.g., when sampling.
- Our case study file format is versioned. We can bump this version and add some logic to the case study loading code that translates old commit ids to the new ones when it sees an outdated case study version.

@bnico99 if you have questions how and where to implement this just contact @vulder or me.

vulder (approved):
LGTM, and works :+1:

thx 4 contributing

boehmseb (approved):

PR #664, opened 2022-07-25
Fix phasar research tool to enable container build

boehmseb:
This PR enables Phasar to be built in a container.

@MMory @pdschubert: There are two things that should be changed in phasar's `install-llvm.sh` script for this to work properly:
- the script currently uses `sudo` for installation, this is not necessary when installing to a user-writeable directory (as we use here)
- Martin suggested using source tarballs instead of cloning llvm to speed up the process, especially since the build is done in a temporary directory


vulder (approved):

vulder:
Should we merge this after the phasar changes were worked in?

vulder:
@fabianbs96 @boehmseb What's the current state of the phasar changes? Can we start to integrate this branch?

boehmseb:
The only remaining issue is to decide how to handle the llvm install.
Phasar currently needs to be built against llvm 14.

The options are:
- make the user responsible for having the correct llvm version installed and only install llvm 14 by ourself when using containers
- always install llvm 14 by ourself
- make the llvm install configurable via CLI (probably not a good idea because this would require reserach-tool-specific command line options)

vulder:
If we can easily switch to 14 it's probably the simplest thing to do, especially as our LLVM state is also 14.

vulder:
@fabianbs96 @boehmseb What's the current state of this PR? We should resolve the llvm-14 issue to merge this one otherwise the branch will drift more and and more.

PR #665, opened 2022-07-25
fta-locations-eval-table 

Keller-1998:
The fta-locations-eval-table contains information about the false-tainted locations. I also decided to switch from sets to lists in the evaluation process in order to recognize multiple taints at one location.


varats/varats/projects/cpp_projects/ect.py
```diff
@@ -0,0 +1,72 @@
...
+        PaperConfigSpecificGit(
+            project_name="ect",
+            remote="https://github.com/fhanau/Efficient-Compression-Tool.git",
+            local="ect",
+            refspec="origin/HEAD",
+            limit=None,
+            shallow=False
+        )
+    ]
+
```
vulder: Does the tool have any dependencies? If `yes`, it would be cool to add a container install spec like:
```
    CONTAINER = get_base_image(
        ImageBase.DEBIAN_10
    ).run('apt', 'install', '-y', 'cmake', 'g++', 'git', 'make', 'python3')
```

vulder (changes requested):
Looks good :+1:

varats/varats/projects/cpp_projects/ect.py
vulder: + Please fix the broken test, there is just a small config specification missing probably

varats/varats/table/table.py
```diff
@@ -147,8 +147,14 @@ def table_file_name(self, table_format: TableFormat) -> str:
         """
         table_ident = ''
         if 'case_study' in self.table_kwargs:
-            case_study: 'CaseStudy' = self.table_kwargs['case_study']
-            table_ident = f"{case_study.project_name}_{case_study.version}_"
+            case_studies = self.table_kwargs['case_study']
```
vulder: @boehmseb is this change necessary?

boehmseb: `Plot.table_kwargs['case_study']` should never be a list.
Therefore, this change should not be necessary.
If there is a plot that can be created for different subsets of case studies, I suggest that plot should override this function.

Keller-1998: With the CLI Option `REQUIRE_MULTI_CASE_STUDY`, the case studies are passed as a list. Therefore, I thought this might be a bug. 

I just pushed a commit that implements your suggestion.

boehmseb: It's not a bug. The intended usage of `REQUIRE_MULTI_CASE_STUDY` is that the plot/table generator creates one plot/table per case study.
A plot/table typically requires either a single case study or all case studies from a paper config at once.

Keller-1998: Ohh, alright, got it!

PR #666, opened 2022-07-31
Table/Plot related fixes and improvements

boehmseb:
- fix some blame interaction graph related plots and tables
- update to pandas 1.3 and its new html/latex table generation
- add progress bar to artefact, plot, and table generation


tests/table/test_bug_overview_table.py
```diff
@@ -26,26 +26,25 @@ def test_basic_repo_latex_booktabs(self) -> None:
 
         # each bug must be matched separately since order is unclear
         result_bug_regex =\
-            re.compile(
+            re.compile(r"None\s*"
```
vulder, line 29: What's the meaning behind the None? Is this some sort of empty column?

vulder (approved):

tests/table/test_bug_overview_table.py
boehmseb: It's the issue number which is not available for this test. I moved it from the last column where it was before and used it as the index in an effort to make the output order more deterministic (which did not work for this test obviously)

PR #668, opened 2022-08-18
Use experiment type in addition to report type for result file selection

boehmseb:


tests/paper/test_paper_config_manager.py
```diff
@@ -12,21 +12,52 @@
...
+
+
+def mocked_get_commit_map(
+    project_name: str,  # pylint: disable=unused-argument
+    cmap_path: tp.Optional[Path] = None,  # pylint: disable=unused-argument
+    end: str = "HEAD",  # pylint: disable=unused-argument
+    start: tp.Optional[str] = None
+) -> CommitMap:  # pylint: disable=unused-argument
+    """
+    Mock function to replace a lazy commit map loader callback.
```
boehmseb:
```suggestion
    Create a dummy commit map.
```

varats/varats/tables/time_stats_table.py
```diff
@@ -25,12 +26,13 @@ def tabulate(self, table_format: TableFormat, wrap_table: bool) -> str:
...
-                get_case_study_file_name_filter(case_study), False
-            )
+            for experiment_type in self.table_kwargs["experiment_type"]:
+                report_files = get_processed_revisions_files(
+                    project_name, experiment_type, TimeReportAggregate,
+                    get_case_study_file_name_filter(case_study), False
+                )
 
-            for report_file in report_files:
+                report_file = report_files[0]
```
vulder, line 35: That looks wrong, why do we suddenly only need one report file?

vulder (changes requested):

varats/varats/tables/time_stats_table.py
boehmseb: Well, not more wrong than before, at least :sweat_smile: 

As I understood the old version, you retrieve all report files because you expect that different experiments have produced each one time report.
That is now handled by explicitly specifying the experiment types via a command line parameter.
And then I assumed, we only want to compare one report file per experiment type (hence the `[0]).
This, of course, also assumes, that there was only one revision analyzed per project and I think that this assumption was also made in the old version.
A better solution would be to either specify a revision via commandline, or extend the table to have a separate entry per found revision.

vulder: @Kaufi-Jonas What did you envision here?

jonas-kaufmann: Yes, @boehmseb is right. I used the assumption here that I am only working with a single revision per case study. If there are more, the table will look off in that it contains two rows for the same binary and experiment with no way to differentiate them. I want to use the table in my bachelor thesis and therefore want to avoid cluttering it up with unnecessary information. Now that I think about it I agree though that this prevents it from being generally useful. I will change this and in some form include the revision information.

vulder (approved):

PR #672, opened 2022-09-03
Support non-master branches

bnico99:


varats-core/varats/mapping/commit_map.py
```diff
@@ -29,11 +28,73 @@ class AmbiguousCommitHash(Exception):
...
+                )
+                wanted_out = git(
+                    "--no-pager", "log", "--all", "--pretty=format:'%H'",
+                    search_range
+                )
+            else:
+                full_out = git("--no-pager", "log", "--pretty=format:'%H'")
+                wanted_out = git(
+                    "--no-pager", "log", "--pretty=format:'%H'", search_range
+                )
```
boehmseb: Shouldn't the commands look like this?

```suggestion
            if master:
                full_out = git(
                    "--no-pager", "log", "--all", "--pretty=format:'%H'"
                )
                wanted_out = git(
                    "--no-pager", "log", "--pretty=format:'%H'",
                    search_range
                )
            else:
                full_out = git("--no-pager", "log", "--all", "--pretty=format:'%H'")
                wanted_out = git(
                    "--no-pager", "log", "--all", "--pretty=format:'%H'", search_range
                )
```

boehmseb (commented):
This already looks quite good. The only thing I found is that I think you mixed up the log commands a bit.

As for the failing tests: 
- TestCommitMap: you need to find out why the range is invalid
- TestPaperConfigManager: I think the best is to create a subclass of `CommitMap` that overrides the relevant functions to return the test data and make the mocked function return an instance of that subclass.

varats-core/varats/mapping/commit_map.py
bnico99: Yes, you're absolutely right, I somehow mixed that up, thanks! I fixed it in a new commit.

varats-core/varats/mapping/commit_map.py
```diff
@@ -29,11 +28,75 @@ class AmbiguousCommitHash(Exception):
...
+            search_range += self.start + ".."
+        search_range += self.end
+
+        with local.cwd(self.git_path):
+            old_head = get_current_branch()
+            git("checkout", self.refspec)
+            if master:
+                full_out = git(
+                    "--no-pager", "log", "--all", "--pretty=format:'%H'"
+                )
```
boehmseb: Since `full_out` is the same for both branches you should pull it out of the `if` to simplify, but also to make it more obvious that the commit ids are indeed always the same regardless of which "view" we use.

boehmseb (changes requested):
One small refactoring remaining, but otherwise looks good.

varats-core/varats/mapping/commit_map.py
bnico99: Valid point. I pushed a new commit that fixes this.

boehmseb (approved):

vulder (approved):

PR #691, opened 2022-10-20
Implements error status tracking for zipped reports

vulder:


varats-core/varats/report/report.py
```diff
@@ -377,6 +386,16 @@ def uuid(self) -> str:
 
         raise ValueError(f'File {self.filename} name was wrongly formatted.')
 
+    @property
+    def file_suffix(self) -> str:
```
boehmseb, line 390: Why use a name different to what is used in our codebase, i.e., `file_ext`? Is it to comply with the Path API?

varats-core/varats/utils/report_utils.py
```diff
@@ -0,0 +1,26 @@
+"""Additional report utilities for moving and renaming report files."""
+import varats.report.report as r
+
+
+# TODO: find a good name
+def adapted_report_filename_status(
```
boehmseb: No good ideas here either.
But how about making this a member function of `ReportFilename` and calling it `with_status()`?

varats-core/varats/report/report.py
vulder: Yes, they referee to it as suffix, which also sounds a bit better to me as the internal `file_ext`, especially as we also have a `status_extension`

varats-core/varats/utils/report_utils.py
vulder: I was tempted to do that yeah. Initially, I wanted the function to also rename the file but that does not work well.
+ I thought maybe we should not put everything as a member function.
But, anyhow, I'm fine with a member function if you like. Would also follow in the design footsteps of the Path API, which is nice.

boehmseb: Would be an option at least. And it would solve the naming problem :smiley:

boehmseb (approved):

PR #692, opened 2022-10-20
Fixes doc building errors due to cryptography cycles

vulder:


boehmseb (approved):
Also works on my setup.

PR #693, opened 2022-10-21
Add (experimental) support for different configs in FeaturePerfRunner

bnico99:


varats-core/varats/base/commandline_option.py
```diff
@@ -20,6 +23,9 @@ def render(self, config: Configuration) -> str:
         Returns: command line string for the given `Configuration`
         """
 
+    def get_option_name(self) -> str:
+        return self.option_name
```
vulder: That does not work (type error), in some cases you set `None` which makes this method not always return `str`.

I would not move the option_name acessor to the base class as not all sub classes have an option name.

varats/varats/experiments/vara/feature_experiment.py
```diff
@@ -119,6 +133,24 @@ def __str__(self, indent: int = 0) -> str:
             f"* {self.project.name}: Run instrumentation verifier", indent * " "
         )
 
+    def create_command_line_options_from_fm(self) -> list[CommandlineOption]:
```
vulder: Overall for initial work/testing this should work.

However, we should discuss in the next meeting with Kalti how we correctly encode the different types of cli options on the FM side.

bnico99:
Obsolete because of https://github.com/se-sic/VaRA-Tool-Suite/pull/700

PR #694, opened 2022-10-25
BlameReport: Parse file name from function

boehmseb:


varats/varats/data/reports/blame_report.py
```diff
@@ -224,14 +225,15 @@ def create_blame_result_function_entry(
         yaml document section."""
         demangled_name = str(raw_function_entry['demangled-name'])
         num_instructions = int(raw_function_entry['num-instructions'])
+        file_name = str(raw_function_entry.get('file'))
```
vulder: This is problematic in the case where no file is specified. `get` correctly handles this and returns `None`, however, `str(None)` results in `'None'`. So, we always get a string and not an optional string.
We need to separate the conversion logic.

vulder (changes requested):

varats/varats/data/reports/blame_report.py
boehmseb: You are right, stupid mistake.

vulder (approved):

PR #700, opened 2022-10-31
Pulls in new configuration aware experiment enumeration

vulder:
resolves se-sic/VaRA#840


varats-core/varats/experiment/experiment_util.py
```diff
@@ -640,3 +658,62 @@ def create_new_failed_result_filepath(
...
+            project.active_revision.variant_by_name(FeatureSource.LOCAL_KEY
+                                                   ).version
+        )
+
+    return None
+
+
+def get_extra_config_options(project: VProject) -> tp.List[str]:
+    """
+    Get a extra program options that where specified in the particular
```
boehmseb:
```suggestion
    Get extra program options that where specified in the particular
```

varats-core/varats/experiment/experiment_util.py
```diff
@@ -640,3 +658,62 @@ def create_new_failed_result_filepath(
         exp_handle, report_type, project, binary, FileStatusExtension.FAILED,
         config_id
     )
+
+
+def get_current_config_id(project: VProject) -> tp.Optional[int]:
+    """
+    Get, if available, the current config id of project. Should the project be
+    not configuration specific None is returned.
```
boehmseb:
```suggestion
    not configuration specific ``None`` is returned.
```

varats-core/varats/utils/config.py
```diff
@@ -0,0 +1,36 @@
...
+import typing as tp
+from pathlib import Path
+
+from varats.base.configuration import Configuration
+from varats.mapping.configuration_map import ConfigurationMap
+from varats.paper.case_study import (
+    CaseStudy,
+    load_configuration_map_from_case_study_file,
+)
+from varats.paper_mgmt.paper_config import PaperConfig
```
boehmseb: `PaperConfig` is not part of varats-core. You are importing from `varats` here.

varats-core/varats/experiment/experiment_util.py
```diff
@@ -20,12 +21,19 @@
...
 from benchbuild.utils.actions import Step, MultiStep, StepResult, run_any_child
 from benchbuild.utils.cmd import prlimit, mkdir
 from plumbum.commands import ProcessExecutionError
 
 import varats.revision.revisions as revs
+from varats.base.configuration import (
+    PlainCommandlineConfiguration,
+    ConfigurationOption,
+)
+from varats.paper_mgmt.paper_config import get_paper_config
```
boehmseb: `get_paper_config` is not part of varats-core. You are importing from varats here.

varats-core/varats/project/sources.py
```diff
@@ -0,0 +1,52 @@
+"""Implements additional sources for projects."""
+
+import typing as tp
+
+import benchbuild as bb
+import plumbum as pb
+from benchbuild.source import Variant
+from benchbuild.source.base import Revision
+
+from varats.paper_mgmt.paper_config import get_paper_config
```
boehmseb: `get_paper_config` is not part of varats-core. You are importing from varats here.

varats-core/varats/project/sources.py
```diff
@@ -0,0 +1,52 @@
...
+        super().__init__(local=self.LOCAL_KEY, remote=remote)
+
+    def version(self, target_dir: str, version: str) -> pb.LocalPath:
+        # TODO: do we need this? Maybe an optional path?
+        return pb.LocalPath('.')
+
+    def versions(self) -> tp.List[Variant]:
+        raise NotImplementedError()
+
+    def fetch(self) -> pb.LocalPath:
```
boehmseb, line 29: It seems odd that this inherits from a class named `FetchableSource` and `fetch` is not allowed :sweat_smile:

varats-core/varats/project/sources.py
```diff
@@ -0,0 +1,52 @@
...
+    """Feature source that automatically enumerates all configurations."""
+
+    LOCAL_KEY = "config_info"
+
+    def __init__(self, remote: tp.Dict[str, str]) -> None:
+        super().__init__(local=self.LOCAL_KEY, remote=remote)
+
+    def version(self, target_dir: str, version: str) -> pb.LocalPath:
+        # TODO: do we need this? Maybe an optional path?
+        return pb.LocalPath('.')
```
boehmseb, line 24: How is this information used by BB or by experiments? I have currently no idea what this function should do for a feature source.

varats/varats/projects/c_projects/xz.py
```diff
@@ -62,6 +63,7 @@ class Xz(VProject):
                 shallow=False
             )
         ),
+        FeatureSource(remote={"1.0": "remotefoo"}),
```
boehmseb: Are these some dummy values because the constructor expects an argument or do they actually have a meaning?

boehmseb (commented):
Still a couple of open questions.
The biggest issue I found is that parts of your implementation in varats-core require the PaperConfig, which is only implemented in varats.

varats-core/varats/project/sources.py
vulder: From what I know, it find the local content, which we don't have. However, None is not a valid return value so I just return the local dir.

varats-core/varats/project/sources.py
vulder: Yeah, the BB interface is currently a bit convoluted. We need to iron this out later on. For features, we don't need anything here.

varats-core/varats/project/sources.py
boehmseb: Would it be an option to raise an error then instead?

varats/varats/projects/c_projects/xz.py
vulder: This is related to the "Fetchable Source" API issue. Currently, this has no meaning for us, we actually would prefer not to add something here.

boehmseb: But then I think it would make more sense that the `FeatureSource` class does not expect this parameter and just passes some dummy value to its super class.

varats-core/varats/project/sources.py
vulder: Then the error get's raised, as this function is called during execution. It's just not a problem if we return '.' as nobody depends on it.

varats-core/varats/utils/config.py
vulder: Yeah, that sucks I guess. Question is: How should we fix this? Making PaperConfig core is not really easy and making everything else, including the sampling of a Version experiment, varats is also :hankey: 
I'm open for ideas.

varats-core/varats/project/sources.py
boehmseb: I guess that's ok then

varats/varats/projects/c_projects/xz.py
vulder: We'll, we want/need the version part of the equation. It's just that we cannot currently only pass a key but need the dict.
This can only be sorted out nicely but adapting BB interfaces into what they actually need instead of what we currently have.

varats-core/varats/utils/config.py
boehmseb: Do you remember the reasons why the paper config never became part of core?
There are the artefacts, but they can be rather easily be extracted from the paper config.
Other than that I see no dependencies of the PaperConfig to non-core varats modules.

vulder: You're right, it seems like it's possible now. I'm not sure, I just remember there where deps we could not get out of it but maybe they have been removed overtime :thinking:

vulder: If I look closer, maybe it was Artefacts that could not easily be removed.

varats/varats/projects/c_projects/xz.py
boehmseb: But we control the (constructor) interface of the `FeatureSource` class, right?
We could design it in a way so that it only takes the required parameters and then the constructor can take care of passing this to the BB API.

boehmseb (commented):
The only thing I still don't like is that the constructor of the `FeatureSource` takes some meaningless dummy values. I understand that the BB API is not a perfect fit for this case, but this our class so I don't see why we cannot make its API nicer to the user.

vulder:
> The only thing I still don't like is that the constructor of the `FeatureSource` takes some meaningless dummy values. I understand that the BB API is not a perfect fit for this case, but this our class so I don't see why we cannot make its API nicer to the user.

I mean, yeah I can only pass the version if you insist. This Bug fix is outstanding since Friday and blocks people from working. And we'll fix the BB API in the next two weeks anyhow, but yeah.

vulder:
Hope this interface is good enough for testing/fixing now.

boehmseb (approved):

varats/varats/projects/c_projects/xz.py
boehmseb: It wasn't that obvious to me that this interface was only for testing and a fix is incoming soon anyway.

PR #707, opened 2022-11-08
Adds ECT + lepton 

Keller-1998:
Adds the C++ projects ECT and lepton.


vulder (changes requested):
Can you please not format the listing in the bb_config file. It's easier to comprehend if everything is in a list.
Should the auto formatter do this automatically, please add a disable comment so nothing get's formatted :)

vulder (approved):

PR #714, opened 2022-11-22
Allow Git Blame overhead measurement in BlameReportExperiment

boehmseb:
- Adds possibility to adapt existing steps to work with ZippedExperimentSteps
- Inject blame measurement flag in BlameReportExperiment compile step

### DIFF
--- a/varats/varats/experiments/vara/blame_report_experiment.py
+++ b/varats/varats/experiments/vara/blame_report_experiment.py
@@ -6,6 +6,7 @@ BlameReport.
 """
 
 import typing as tp
+from pathlib import Path
 
 from benchbuild import Project
 from benchbuild.utils import actions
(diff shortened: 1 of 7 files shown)


vulder (approved):

varats/varats/experiments/vara/blame_report_experiment.py
```diff
@@ -99,18 +102,32 @@ def analyze(self) -> actions.StepResult:
...
+        return actions.StepResult.OK
+
+
 class BlameReportExperiment(VersionExperiment, shorthand="BRE"):
     """Generates a blame report of the project(s) specified in the call."""
 
     NAME = "GenerateBlameReport"
 
-    REPORT_SPEC = ReportSpecification(BR)
+    REPORT_SPEC = ReportSpecification(BR, WallTimeReportAggregate)
```
vulder, line 124: Do we want to keep this?

vulder (changes requested):
Overall, everything looks nice and tidy.

The only thing we should discuss is: Should we keep the WTReport for BlameReports? As we don't plan on expanding/by default using the measurement infra on the vara side.

varats/varats/experiments/vara/blame_report_experiment.py
boehmseb: I don't think we really need this. But we could maybe leave this branch lying around somewhere,

vulder: Ok, I'll keep the branch and just close the PR

vulder:
Keep the changes alive but remove the PR

PR #718, opened 2022-11-29
WIP: GenerateCoverage Experiment

danjujan:


tests/helper_utils.py
```diff
@@ -8,6 +8,7 @@
 from functools import wraps
```
vulder: Why do the imports of this file change but no conent?

tests/plots/test_llvm_coverage_plot.py
```diff
@@ -0,0 +1,127 @@
...
+    RunConfig,
+    ConfigCoverageReportMapping,
+)
+
+CODE_REGION_1 = CodeRegion.from_list([9, 79, 17, 2, 4, 0, 0, 0], "main")
+
+
+class TestCodeRegion(unittest.TestCase):
+
+    def test_feature_config_report_map(self):
```
vulder: I'm not sure what this tests, the beginning looks a bit like testing the config handling part of the configuration, which should be tested in separation and in the config module.

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
+"""Code region tree and coverage report."""
+
+from __future__ import annotations
```
vulder, line 3: From which python version is this? As we have update to min 3.9 we maybe don't need this anymore

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+                    region.childs.sort()
+                    for child in childs_to_move:
+                        child.parent = region
+                        node.childs.remove(child)
+                node.childs.append(region)
+                node.childs.sort()
+                region.parent = node
+                # Actually this is possible,
+                # e.g. a for loop can be executed
+                # more often than its function.
```
vulder:
```suggestion
                # Actually this is possible, e.g., a for loop can
                # be executed more often than its function.
```

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+                #assert region.count <= node.count
```
vulder: Still needed?

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+            raise ValueError("The given region is not a subregion!")
+        if region in self:
+            raise ValueError("The given region exists already!")
+
+        # Find the right child to append to
+        # Should be the first code region where region is a subregion
+        # when traversing the tree in postorder
+        for node in self.iter_postorder():
+            if node.is_subregion(region):
+                if len(node.childs) > 0:
```
vulder:
```suggestion
                if node.childs:
```
This should be enough, as empty lists are false

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+                    # check which childs should become childs of regions
+                    childs_to_move = []
+                    for child in node.childs:
+                        if region.is_subregion(child):
+                            childs_to_move.append(child)
+                    region.childs.extend(childs_to_move)
+                    region.childs.sort()
+                    for child in childs_to_move:
+                        child.parent = region
+                        node.childs.remove(child)
```
vulder:
```suggestion
                    for child in node.childs:
                        if region.is_subregion(child):
                            childs_to_move.append(child)

                    region.childs.extend(childs_to_move)
                    region.childs.sort()

                    for child in childs_to_move:
                        child.parent = region
                        node.childs.remove(child)

```

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+                # e.g. a for loop can be executed
+                # more often than its function.
+                #assert region.count <= node.count
+                break
+
+    def merge(self, region: CodeRegion) -> None:
+        """Merges region into self by adding all counts of region to the counts
+        of self."""
+        for x, y in zip(self.iter_breadth_first(), region.iter_breadth_first()):
+            assert x == y, "CodeRegions are not identical"
```
vulder: As far as I now, a PEP suggests to use raise assertion error over assert, to also trigger in non debug cases

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+                # Coverage increased
+                x.count = 1
+
+    def is_identical(self, other: object) -> bool:
+        """Is the code region equal and has the same coverage?"""
+        if not isinstance(other, CodeRegion):
+            return False
+        if not (self == other and self.count == other.count):
+            return False
+        for code_region_a, code_region_b in zip(
```
vulder:
```suggestion
        if not isinstance(other, CodeRegion):
            return False

        if not (self == other and self.count == other.count):
            return False

        for code_region_a, code_region_b in zip(
```

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+            if not (
+                code_region_a == code_region_b and
+                code_region_a.count == code_region_b.count
+            ):
+                return False
+        return True
+
+    # Compare regions only depending on their
+    # start lines and columns + their type
+
```
vulder: This rnd. comment in the class seems a bit lost, should it be put somewhere else?

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+            return False
+        for code_region_a, code_region_b in zip(
+            self.iter_breadth_first(), other.iter_breadth_first()
+        ):
+            if not (
+                code_region_a == code_region_b and
+                code_region_a.count == code_region_b.count
+            ):
+                return False
+        return True
```
vulder:
```suggestion
                return False

        return True
```

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+
+    # Compare regions only depending on their
+    # start lines and columns + their type
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, CodeRegion):
+            return False
+        return (
```
vulder:
```suggestion
        if not isinstance(other, CodeRegion):
            return False

        return (
```

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+        )
+
+    def __lt__(self, other: CodeRegion) -> bool:
+        if (
+            self.start.line < other.start.line or
+            self.start.line == other.start.line and
+            self.start.column < other.start.column
+        ):
+            return True
+        return False
```
vulder:
```suggestion
            return True

        return False
```
I personally, like a few empty lines here and there to separate logical blocks, otherwise it's harder to spot for the eye

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+class FunctionCodeRegionMapping(tp.Dict[str, CodeRegion]):
+    """Mapping from function names to CodeRegion objects."""
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, FunctionCodeRegionMapping):
+            return False
+        for self_value, other_value in zip(self.values(), other.values()):
+            if not self_value.is_identical(other_value):
+                return False
+        return True
```
vulder:
```suggestion
    def __eq__(self, other: object) -> bool:
        if not isinstance(other, FunctionCodeRegionMapping):
            return False

        for self_value, other_value in zip(self.values(), other.values()):
            if not self_value.is_identical(other_value):
                return False

        return True
```

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+                code_region_a.diff(code_region_b)
+
+    def _import_functions(
+        self, json_file: Path,
+        filename_function_mapping: FilenameFunctionMapping
+    ) -> FilenameFunctionMapping:
+        with json_file.open() as file:
+            j = json.load(file)
+        # Compatibility check
+        try:
```
vulder: `j` could be undefined, producing a hard to interpret run-time error

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+            j_type = j["type"]
+            j_version = j["version"].split(".")
+            assert j_type == "llvm.coverage.json.export" and j_version[0] == "2"
+        except Exception as err:
```
vulder: Can you make the catch a bit more precise?

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+                assert code_region_a == code_region_b
+
+                code_region_a.diff(code_region_b)
+
+    def _import_functions(
+        self, json_file: Path,
+        filename_function_mapping: FilenameFunctionMapping
+    ) -> FilenameFunctionMapping:
+        with json_file.open() as file:
+            j = json.load(file)
```
vulder: Maybe give `j` a nicer more "clear" name

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+        absolute_path = j["absolute_path"]
+        data: tp.Dict[str, tp.Any] = j["data"][0]
+        # files: tp.List = data["files"]
+        functions: tp.List[tp.Any] = data["functions"]
+        totals: tp.Dict[str, tp.Any] = data["totals"]
+
+        for function in functions:
+            name: str = function["name"]
+            # count: int = function["count"]
+            # branches: list = function["branches"]
```
vulder, line 871: not needed, or why are these outcommented?

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+            j_version = j["version"].split(".")
+            assert j_type == "llvm.coverage.json.export" and j_version[0] == "2"
+        except Exception as err:
+            raise NotImplementedError(
+                "Cannot import functions. Json format unknown"
+            ) from err
+
+        absolute_path = j["absolute_path"]
+        data: tp.Dict[str, tp.Any] = j["data"][0]
+        # files: tp.List = data["files"]
```
vulder, line 864: not needed?

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+            for function in function_region_mapping:
+                counted_functions += 1
+                code_region = function_region_mapping[function]
+                for region in code_region.iter_breadth_first():
+                    if region.kind == CodeRegionKind.CODE:
+                        counted_code_regions += 1
+                        if region.is_covered():
+                            covered_regions += 1
+                        else:
+                            notcovered_regions += 1
```
vulder:
```suggestion
                            notcovered_regions += 1

```

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+    NOTE: The colored representation differs a bit!
+    """
+    result = []
+    for file in list(report.filename_function_mapping):
+        function_region_mapping = report.filename_function_mapping[file]
+        path = Path(file)
+        tmp_value = _cov_show_file(path, base_dir, function_region_mapping, [])
+        if not tmp_value[-1].endswith("\n"):
+            # Add newline if file does not end with one
+            tmp_value.append("\n")
```
vulder:
```suggestion

        if not tmp_value[-1].endswith("\n"):
            # Add newline if file does not end with one
            tmp_value.append("\n")

```

varats/varats/data/reports/llvm_coverage_report.py
```diff
@@ -0,0 +1,708 @@
...
+    buffer = []
+    for line_number, segments in segments_dict.items():
+        if len(segments) > 1:
+            # Workaround: Ignore counts for last segment with whitespaces
+            # and single ';' that ends with "\n"
+            segments[-1] = (None, segments[-1][1]
+                           ) if segments[-1][1].endswith("\n") and (
+                               str.isspace(segments[-1][1].replace(";", "", 1))
+                           ) else segments[-1]
+        #buffer.append(str(segments) + "\n")
```
vulder: no longer needed?

varats/varats/experiments/vara/llvm_coverage_experiment.py
```diff
@@ -0,0 +1,174 @@
...
+                    )
+                    (llvm_cov > str(json_name))()
+
+                    # Add absolute path to json to compute
+                    # relative filenames later in the report
+                    with open(json_name) as file:
+                        coverage = json.load(file)
+                    coverage["absolute_path"] = str(
+                        Path(self.project.source_of_primary).resolve()
+                    )
```
vulder:
```suggestion
                        coverage = json.load(file)

                    coverage["absolute_path"] = str(
                        Path(self.project.source_of_primary).resolve()
                    )

```

varats/varats/experiments/vara/llvm_coverage_experiment.py
```diff
@@ -0,0 +1,174 @@
...
+                    with open(json_name, "w") as file:
+                        json.dump(coverage, file)
+
+        return actions.StepResult.OK
+
+
+# Please take care when changing this file, see docs experiments/just_compile
```
vulder: Maybe also add a quite note why?

varats/varats/plots/llvm_coverage_plot.py
```diff
@@ -0,0 +1,323 @@
...
+            return self.x
+        if isinstance(self.x, str):
+            return True
+        raise NotImplementedError()
+
+    def __repr__(self) -> str:
+        return repr(self.x)
+
+
+class RunConfig(tp.FrozenSet[tp.Tuple[str, ConfigValue]]):
```
vulder: Why do you need to build a config class here, is the one that parses the config format not working. If not, can we maybe adapt it?

vulder (changes requested):
Hmm, I'm not 100% sure about the zip files for the tests. On the one side, it's good for testing, on the other, we don't know how to update/fix them when stuff breaks.
Maybe, we should at least put some doc/wording in a file that specifies how the information can be regenerated?
+ I hope, this does not make the tests brittel

@boehmseb what's your take on that?

Overall, the implementation already looks quite good.

varats/varats/experiments/vara/llvm_coverage_experiment.py
boehmseb: I think this was copied by mistake. This comment is there to keep the compile experiment in sync with the docs.

boehmseb:
> Hmm, I'm not 100% sure about the zip files for the tests. On the one side, it's good for testing, on the other, we don't know how to update/fix them when stuff breaks. Maybe, we should at least put some doc/wording in a file that specifies how the information can be regenerated?
> 
>     * I hope, this does not make the tests brittle
> 
> 
> @boehmseb what's your take on that?

We kind of brought this on ourselves by introducing .zip files as reports.
The corresponding report class should make obvious why there is a .zip lying around.
So technically, this is no different to including a normal report as test input.
I understand your concern, but I'm not sure whether an additional readme of some sorts can provide additional information that is not already encoded by the test and the report class.

varats/varats/data/reports/llvm_coverage_report.py
danjujan: No, that was a comment/assertion for me while developing the code region tree. Removed it.

tests/helper_utils.py
danjujan: Probably an oversight while I developed https://github.com/se-sic/VaRA-Tool-Suite/pull/770. Removed them.

varats/varats/data/reports/llvm_coverage_report.py
danjujan: This is still necessary. Otherwise the following is not possible. See https://peps.python.org/pep-0563/ and https://peps.python.org/pep-0649/
```python
class CodeRegion
    def from(cls, x) -> CodeRegion:
       ...
```

varats/varats/data/reports/llvm_coverage_report.py
danjujan: Correct. Removed.

varats/varats/experiments/vara/llvm_coverage_experiment.py
danjujan: Yeay, copy pasted from the JustCompile experiment. Removed.

varats/varats/data/reports/llvm_coverage_report.py
danjujan: Changed.

varats/varats/data/reports/llvm_coverage_report.py
danjujan: This comment refers to the following comparison functions `__eq__`, `__lt__`, `__gt__`,`__le__` and `__ge__`. I could make it a docstring for `__eq__` instead or simply remove it.

varats/varats/data/reports/llvm_coverage_report.py
danjujan: Not needed for now. These are parts in the json that could be interesting for someone who wants to extend the parsing in the future.

varats/varats/data/reports/llvm_coverage_report.py
danjujan: Not needed for now. Same answer as above.

tests/plots/test_llvm_coverage_plot.py
danjujan: This only tests if `ConfigCoverageReportMapping` will use the right configs for diffing.

varats/varats/data/reports/llvm_coverage_report.py
danjujan: Renamed `j` to `coverage_json`

varats/varats/data/reports/llvm_coverage_report.py
danjujan: Done.

varats/varats/data/reports/llvm_coverage_report.py
danjujan: A `NotImplementedError` is now raised when json decoding fails.

danjujan:
> > Hmm, I'm not 100% sure about the zip files for the tests. On the one side, it's good for testing, on the other, we don't know how to update/fix them when stuff breaks. Maybe, we should at least put some doc/wording in a file that specifies how the information can be regenerated?
> > ```
> > * I hope, this does not make the tests brittle
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @boehmseb what's your take on that?
> 
> We kind of brought this on ourselves by introducing .zip files as reports. The corresponding report class should make obvious why there is a .zip lying around. So technically, this is no different to including a normal report as test input. I understand your concern, but I'm not sure whether an additional readme of some sorts can provide additional information that is not already encoded by the test and the report class.

As @boehmseb said. The zip files were obtained by copy pasting them from the results folder after running the `GenerateCoverage` experiment.

vulder:
That part I know and "buy", I'm more concerned to how we can update them if needed. We should add docu to the tests how to regenerate the files, project + experiment.
But what also concerns me, it looks like some of the paths where edited by had, which could introduce brittleness as this rewriting is not done automatically. I would be fine with the zips if we have a small description in the test files that tells a developer how to regenerate these files, because from what I see, it's a bit more difficult than "just" running the experiment.

By the way, as soon as we have a "docu format" or idea how to specify these, we should probably also document the other ones.

danjujan:
> That part I know and "buy", I'm more concerned to how we can update them if needed. We should add docu to the tests how to regenerate the files, project + experiment. But what also concerns me, it looks like some of the paths where edited by had, which could introduce brittleness as this rewriting is not done automatically. I would be fine with the zips if we have a small description in the test files that tells a developer how to regenerate these files, because from what I see, it's a bit more difficult than "just" running the experiment.
> 
> By the way, as soon as we have a "docu format" or idea how to specify these, we should probably also document the other ones.

I regenerated the files and noted down my steps. https://github.com/se-sic/VaRA-Tool-Suite/blob/fba20d72770514086ea65262a491378c5ea11963/tests/TEST_INPUTS/results/FeaturePerfCSCollection/HowToReproduce.md. Feedback welcome.

PR #719, opened 2022-12-04
Improved docu for container

bnico99:


docs/source/tutorials/container_guide.rst
```diff
@@ -4,7 +4,7 @@ Container Guide
 VaRA-TS supports running experiments inside containers to make them more portable and reproducible.
 This guide walks you through the steps how to prepare and execute experiments in containers and assumes that your system has already been set up for executing containers.
 You can find the instructions for setting up containers on your system :ref:`here <Running BenchBuild in a Container>`.
-More details about the inner workings of the container support in VaRA-TS can be found :ref:`here <Container support>`.
+More details about the inner workings of the container support in VaRA-TS can be found :ref:`here <Container support>`. Note that, by default, the most recent version of VaRA-TS will be used inside the container. If you want to install VaRA-TS from source, you have to specify this in the ``.varats.yml`` configuration by setting the value of ``from_source`` to true and the value of ``varats_source`` to the directory containing the source code of the VaRA-TS.
```
boehmseb:
```suggestion
More details about the inner workings of the container support in VaRA-TS can be found :ref:`here <Container support>`.
Note that, by default, the most recent version of VaRA-TS available from PyPI will be used inside the container.
If you want to use your local development version of VaRA-TS instead, you have to specify this in the ``.varats.yml`` configuration file by setting the value of ``from_source`` to true and the value of ``varats_source`` to the directory containing the source code of the VaRA-TS.
```

docs/source/tutorials/container_guide.rst
```diff
@@ -15,6 +15,20 @@ More details about the inner workings of the container support in VaRA-TS can be
 
        vara-buildsetup build vara --container=DEBIAN_10
 
+   Note that the underlying tools do not support network file systems. In the ``.benchbuild.yml`` configuration, you need to set the values of ``root`` and ``runroot`` to a directory that is locally available on the current machine. For example, you can use the following values:
```
boehmseb:
```suggestion
   Note that the underlying tools may not support network file systems (i.e., if you are using podman in `rootless mode <https://github.com/containers/podman/blob/master/rootless.md>`_).
   In the ``.benchbuild.yml`` configuration, you need to set the values of ``root`` and ``runroot`` to a directory that is locally available on the current machine.
   For example, you can use the following values:
```

boehmseb (commented):
Thanks for the contribution. This makes a nice addition to the container guide.

A small note: we usually put each sentence in a new line to make it easier to track changes.

vulder (approved):

PR #720, opened 2022-12-09
Remove packaging.LegacyVersion usage

danjujan:
I created a new Python virtualenvironment today and encountered the following error while running `vara-cs cleanup`:
```
ImportError: cannot import name 'LegacyVersion' from 'packaging.version'
```
Turns out that our dependency `packaging` got a new release and removed the already deprecated `LegacyVersion`.
https://github.com/pypa/packaging/blob/main/CHANGELOG.rst#220---2022-12-07

Fixed this by removing our usage of `LegacyVersion`.

### DIFF
--- a/varats-core/varats/provider/cve/cve.py
+++ b/varats-core/varats/provider/cve/cve.py
@@ -15,7 +15,7 @@ import zipfile
 from datetime import datetime
 
 import requests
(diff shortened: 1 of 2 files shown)


vulder (approved):
lgtm

PR #721, opened 2022-12-11
Reduce memory overhead of TEFReport

bnico99:
Since currently the TEFReport loads the whole JSON file into main memory before parsing it, it is easy to run out of memory for huge tracefiles (for example, the main memory usage for a 30 GB tracefile was >200 GB). I implemented a TEFReport that streams the JSON file and, additionally, reduces saving duplicate information by introduing a key-value store that stores function/region names. This way, TraceEvents only need to include an int instead of a (potentially large) string.

Note that I merged #720 into my branch such that the tests do not fail, so not all changes in this branch are related to the above description.


varats-core/varats/report/tef_report.py
```diff
@@ -57,8 +60,8 @@ def __init__(self, json_trace_event: tp.Dict[str, tp.Any]) -> None:
         self.__tid = int(json_trace_event["tid"])
 
     @property
-    def name(self) -> str:
-        return self.__name
+    def name_id(self) -> int:
+        return self.__name_id
```
vulder: The name id part should not be in the public API, as it is only a implementation detail. As user should still just call `.name` and get the name transparently, without knowing that locally only an ID was stored.

varats-core/varats/report/tef_report.py
```diff
@@ -123,11 +122,28 @@ def stack_frames(self) -> None:
...
-    def _parse_trace_events(
-        raw_event_list: tp.List[tp.Dict[str, tp.Any]]
-    ) -> tp.List[TraceEvent]:
-        return [TraceEvent(data_item) for data_item in raw_event_list]
+    @property
+    def key_name_map(self) -> tp.List[str]:
+        return self.__key_name_map
+
+    def get_name_by_id(self, name_id: int) -> str:
+        return self.key_name_map[name_id]
```
vulder: Should not be public, i.e., only be used by the `TraceEvent` class.

varats-core/varats/report/tef_report.py
```diff
@@ -1,11 +1,12 @@
 """Report module to create and handle trace event format files, e.g., created
 with chrome tracing."""
 
-import json
 import typing as tp
 from enum import Enum
 from pathlib import Path
 
+import ijson
```
vulder, line 9: If ijson does not support type hints, we need to adapt the mypy file to ignore types from there.

vulder (changes requested):

varats-core/varats/report/tef_report.py
bnico99: @vulder I was not sure how to handle that best. The "problem" is that in this implementation, the TraceEvent does not know its name, it only knows its name id since the id-name mapping is saved in the TEFReport. Should the `name` method take the `TEFReport` as an argument or should we add a `TEFReport` attribute to the `TraceEvent` or did you imaging something entirely different?

varats-core/varats/report/tef_report.py
```diff
@@ -123,11 +122,28 @@ def stack_frames(self) -> None:
...
+
+    def get_name_by_id(self, name_id: int) -> str:
+        return self.key_name_map[name_id]
+
+    def _parse_json(self) -> None:
+        trace_events = list()
+        with open(self.path, "r", encoding="utf-8") as f:
+            for trace_event in ijson.items(f, "traceEvents.item"):
+                if trace_event["name"] in self.__key_name_map:
+                    name_id = self.__key_name_map.index(trace_event["name"])
```
boehmseb: Depending on how often you have to look up an index like this and how large this list can become it might be better to use an actual map instead.

varats-core/varats/report/tef_report.py
```diff
@@ -123,11 +122,28 @@ def stack_frames(self) -> None:
...
+                else:
+                    self.__key_name_map.append(trace_event["name"])
+                    name_id = len(self.__key_name_map) - 1
+                trace_events.append(TraceEvent(trace_event, name_id))
+        # TODO: The following way to extract the timestampUnit is certainly not the best solution. However, I am not yet sure what's the best alternative.
+        with open(self.path, "r", encoding="utf-8") as f:
+            for timestamp_unit in ijson.items(f, "timestampUnit"):
+                self.__timestamp_unit: str = str(timestamp_unit)
+        self.__trace_events: tp.List[TraceEvent] = trace_events
```
boehmseb: You could, at least, reuse the already open file instead of opening it a second time. But other than that, I also see no better way.

bnico99: @boehmseb I re-opened the file because the pointer needs to be reset to the first byte (due to how ijson works - at least I could not figure another way to do it via the high-level APIs that ijson offers; it might be possible with lower-level APIs of ijson though) and `f.seek(0)` made some tests fail, though I am honestly not sure why.

varats-core/varats/report/tef_report.py
bnico99: @boehmseb I used a list instead of a map since a list makes it easy to use incremental indices (by just appending) and I did not expect the list to reach a size (since it just contains the name of each traced region once) such that the performance influence is substantial. Additionally, with a map, we would still have to do a reverse-lookup at this point (to get the ID of existing entries).

boehmseb: If you think the benefits of using a list here outweigh the complexity of the lookup than that's Ok.

varats-core/varats/report/tef_report.py
boehmseb: I see. Then there probably is no better way. Although I'm a little bit surprised that the library works this way. This is not apparent from their documentation.

varats-core/varats/report/tef_report.py
bnico99: @vulder Fixed the leaky abstraction as discussed.

varats-core/varats/report/tef_report.py
bnico99: Fixed.

varats-core/varats/report/tef_report.py
vulder: That sounds strange, normally seek or a similar method should be able to reset the file to the beginning :thinking:

bnico99: @boehmseb It might be possible with some lower-level APIs that the library offers. I will have a look into it. If it is not possible or feasible, I will have a look at the `f.seek(0)` problem (@vulder)

bnico99: @boehmseb I implemented a different approach that makes use of lower-level APIs that ijson offers. It requires a bit more logic on our side (instead of ijson's side), but it prevents having to read the (potentially huge) file twice.

varats-core/varats/report/tef_report.py
```diff
@@ -122,27 +132,39 @@ def stack_frames(self) -> None:
...
-        with open(self.path, "r", encoding="utf-8") as f:
-            for timestamp_unit in ijson.items(f, "timestampUnit"):
-                self.__timestamp_unit: str = str(timestamp_unit)
+        trace_events: tp.List[TraceEvent] = list()
+        with open(self.path, "rb") as f:
+            parser = ijson.parse(f)
+            trace_event: tp.Dict[str, str] = {}
+            for prefix, event, value in parser:
+                if event == "map_key":
+                    key = value
```
boehmseb, line 143: I'm surprised that mypy does not complain that `key` might be used in line 164 without prior definition here.
It still might be a good idea to define it with an empty string or so outside the loop.

varats-core/varats/report/tef_report.py
boehmseb: Looks good. The question now is whether all these `startswith` checks are in the end faster than re-opening the file once :sweat_smile:
The correct way would probably to profile both versions on a large trace file but I'm not sure whether thta's worth the effort.

boehmseb (approved):

varats-core/varats/report/tef_report.py
bnico99: Done

varats-core/varats/report/tef_report.py
bnico99: @boehmseb Fair point. I did not test it yet. If you insist, I can try to figure it out. Otherwise, I would leave it this way for now.

vulder (approved):

boehmseb (approved):

PR #723, opened 2022-12-14
Add sample FeaturePerfRegression case study

bnico99:


varats/varats/projects/cpp_projects/dummy_case_study.py
```diff
@@ -0,0 +1,90 @@
+"""Project file for the feature performance case study collection."""
```
vulder: This file is not a real cpp_project, we should put it in the perf examles

varats/varats/projects/perf_tests/feature_perf_regression.py
```diff
@@ -0,0 +1,112 @@
...
+        pass
+
+    def compile(self) -> None:
+        """Compile the project."""
+        feature_perf_regression_source = local.path(
+            self.source_of(self.primary_source)
+        )
+
+        # Decrease instruction threshold and disable exceptions
+        self.cflags += ["-fvara-instruction-threshold=1", "-fno-exceptions"]
```
vulder:
```suggestion
        self.cflags += ["-fvara-instruction-threshold=1", "-fno-exceptions"]
```
The threshold is experiment specific for feature instr. and should not be part of the project.

I'm not sure what the best solution is here :thinking:

varats/varats/projects/perf_tests/feature_perf_regression.py
```diff
@@ -0,0 +1,112 @@
...
+    """Test project for identifying performance regressions across revisions."""
+
+    NAME = 'FeaturePerfRegression'
+    GROUP = 'perf_tests'
+    DOMAIN = ProjectDomains.TEST
+
+    SOURCE = [
+        PaperConfigSpecificGit(
+            project_name="FeaturePerfRegression",
+            remote="https://github.com/bnico99/FeaturePerfRegression.git",
```
vulder: We should move this repo to `se-sic` so that it's preserved into the future.

vulder (changes requested):

varats/varats/projects/perf_tests/feature_perf_regression.py
bnico99: @vulder I transferred the repo (I though starting the transferring process opens a request, but it seems like it has directly been transferred: https://github.com/se-sic/FeaturePerfRegression)

varats/varats/projects/perf_tests/feature_perf_regression.py
bnico99: Maybe the threshold could be added as a command line option that is passed to the experiment or incorporated into the case_study file somehow, if that makes sense to you?

varats/varats/projects/perf_tests/feature_perf_regression.py
vulder: Yes, seems it worked. Update the `remote` and we can merge it afterwards :)

varats/varats/projects/perf_tests/feature_perf_regression.py
vulder: The case_study file is a bit to generic for that and would then need some form of custom section :thinking: 

@Kaufi-Jonas at least added a variable way to pass the threshold to the base class for correctly setting it in the cflags, however, this still does not solve the problem of choosing a correct value. From the conceptual point of view, it's the job of the experiment to define "how fine grained" measurements should be made. So, the problem is that we currently don't have a good way for the experiment to determine this. One idea I currently have floating in my head would be to use the 'Domain' of the project. If it's a test project set it to 0/1 if not use the default. For more elaborate work one could come up with a better strategy to select the threshold, however, that would also require some form of project introspection to determine it.

vulder (approved):

PR #724, opened 2022-12-20
Support repetitions in FeaturePerfRunner and FeaturePerfAnalysisTable and add RegionPerfAnalysisTable

bnico99:


bnico99:
Depends on https://github.com/se-sic/VaRA-Tool-Suite/pull/730 and https://github.com/se-sic/VaRA-Tool-Suite/pull/738

PR #715, opened 2022-11-23
Fix varats bb config vals being overwritten

boehmseb:

PR #728, opened 2023-01-04
Update several blame interaction related plots

boehmseb:
Waiting on #714 and #715


varats/varats/tables/case_study_metrics_table.py
```diff
@@ -31,26 +31,27 @@ def tabulate(self, table_format: TableFormat, wrap_table: bool) -> str:
...
+            authors = num_project_authors(project_name, revision)
 
             cs_dict = {
                 project_name: {
                     "Domain":
                         str(project_cls.DOMAIN)[0].upper() +
                         str(project_cls.DOMAIN)[1:],
                     "LOC":
-                        calc_repo_loc(project_repo, rev_range),
+                        loc,
```
vulder: Maybe we should include both LOC numbers here? For some it's important to see the full project, however, we probably also want to see the locs that are "just" part of the repo without it's dependencies.

vulder (changes requested):

varats/varats/tables/case_study_metrics_table.py
boehmseb: Depends on what our goal is. If we want to make the table from the SEAL paper exactly reproducible then we should leave it as is and maybe include both in the future when needed.
Otherwise, it's easy enough to include both.

vulder: I'm fine with both. The only thing I currently see is that we should indicate what `LOC` means. Maybe `LOC_WS`
for `with submodule`? Got a better idea?

boehmseb: I added both locs and named the columns `LOC (repo)` and `LOC (project)`.

vulder (approved):

PR #730, opened 2023-01-09
Add function relative id printer

bnico99:


bnico99:
Depends on https://github.com/se-sic/VaRA/pull/980

vulder (commented):
Let's discuss this part in a meeting next week or so. I'm not 100% what parts we'll need in the future and which ones we only need for your thesis.

PR #734, opened 2023-01-12
Adapts folder layout of config id specific result files

vulder:
resolves se-sic/VaRA#977


varats/varats/projects/test_projects/example_test_repo.py
```diff
@@ -28,7 +28,7 @@ class ExampleTestRepo(VProject):
             project_name="example_test_repo",
             remote="BasicTestRepos/ExampleRepo",
             local="example_repo",
-            refspec="origin/HEAD",
+            refspec="HEAD",
```
boehmseb, line 31: What's the reason for this change?

vulder: It was wrong it the first place and does not work :wink:

boehmseb (approved):

PR #738, opened 2023-01-19
Add new instr_type that directly aggregates runtime per region/region interaction

bnico99:


varats/varats/experiments/vara/agg_region_interaction_perf_runner.py
```diff
@@ -0,0 +1,66 @@
...
+        self, project: VProject
+    ) -> tp.MutableSequence[actions.Step]:
+        """
+        Returns the specified steps to run the project(s) specified in the call
+        in a fixed order.
+
+        Args:
+            project: to analyze
+        """
+        instr_type = "rit_trace_event"  # trace_event
```
vulder: This can be simplified as soon as we merge https://github.com/se-sic/VaRA-Tool-Suite/pull/748

vulder (commented):
Looks reasonable, but we should wait to integrate this until the instrumentation are merged.

boehmseb (approved):
lgtm.

But what's up with mypy here?

PR #748, opened 2023-02-17
rework feature tracing experiments and time-stats table, add feature sampling experiment

jonas-kaufmann:
Requires https://github.com/se-sic/VaRA/pull/972 and https://github.com/se-sic/vara-llvm-project/pull/84.

Even though this contains a lot of changes, the commit history is clean meaning each commit touches different files. I recommend reviewing based on that.


varats-core/varats/experiment/experiment_util.py
```diff
@@ -343,6 +340,10 @@ def report_spec(self) -> ReportSpecification:
         """Experiment report specification."""
         return self.__experiment.report_spec()
 
+    def shorthand(self) -> str:
```
vulder: Why do you need to access this directly? I try to keep such low-level accesses low and more use high-level APIs.

varats/varats/data/reports/feature_tracing_stats_report.py
```diff
@@ -0,0 +1,8 @@
+"""Module for FeatureTracingStatsReport."""
+
+from varats.report.report import BaseReport
+
+
+class FeatureTracingStatsReport(BaseReport, shorthand="FTS", file_type="txt"):
```
vulder, line 6: Is this now specific to tracing?

varats/varats/ts_utils/click_param_types.py
```diff
@@ -72,11 +72,12 @@ def convert(
         if isinstance(value, str):
             values = list(map(str.strip, value.split(",")))
 
-        return [
-            item for v in values for item in self.__choices[
+        converted = []
```
vulder: @boehmseb Can you take an additional look at this CLI changes.

varats/varats/experiments/vara/feature_experiment.py
```diff
@@ -31,6 +40,23 @@
...
+    PRINT = "print"
+    """Print tracing information to stdout."""
+    TEF = "trace_event"
+    """Produce a trace file."""
+    USDT = "usdt"
+    """Insert USDT probes."""
+    USDT_RAW = "usdt_raw"
+    """Insert raw USDT probes."""
+    VERIFY = "instr_verify"
+    """Verify markers."""
```
vulder:
```suggestion
    """Don't add any instrumentation."""
    NONE = ""
    """Print tracing information to stdout."""
    PRINT = "print"
    """Produce a trace file."""
    TEF = "trace_event"
    """Insert USDT probes."""
    USDT = "usdt"
    """Insert raw USDT probes."""
    USDT_RAW = "usdt_raw"
    """Verify markers."""
    VERIFY = "instr_verify"
```
we normally put comments above

varats/varats/projects/c_projects/gzip.py
```diff
@@ -78,6 +78,14 @@ class Gzip(VProject, ReleaseProviderHook):
                     "https://github.com/simonepri/geo-maps/releases/"
                     "download/v0.6.0/countries-land-1km.geo.json"
             }
+        ),
+        HTTP(
```
vulder: If you need more, you can just download the whole zip archive with `HTTPUntar` from benchbuild.

varats/varats/experiments/vara/feature_instrumentation_points.py
```diff
@@ -0,0 +1,111 @@
...
+)
+from varats.experiments.vara.feature_experiment import FeatureExperiment
+from varats.project.varats_project import VProject
+from varats.report.report import ReportSpecification
+
+
+class CollectInstrumentationPoints(actions.ProjectStep):  # type: ignore
+    """See `DESCRIPTION`."""
+
+    NAME = "CollectnstrumentationPoints"
```
vulder:
```suggestion
    NAME = "CollectInstrumentationPoints"
```

varats/varats/experiments/vara/feature_instrumentation_points.py
```diff
@@ -0,0 +1,111 @@
...
+    RunWLLVM,
+    get_cached_bc_file_path,
+)
+from varats.experiments.vara.feature_experiment import FeatureExperiment
+from varats.project.varats_project import VProject
+from varats.report.report import ReportSpecification
+
+
+class CollectInstrumentationPoints(actions.ProjectStep):  # type: ignore
+    """See `DESCRIPTION`."""
```
vulder: Just write a quick class desc, the description might overlap but in the docs this is rendered differently.

varats/varats/experiments/vara/feature_instrumentation_points.py
```diff
@@ -0,0 +1,111 @@
...
+
+    NAME = "FeatureInstrumentationPoints"
+
+    REPORT_SPEC = ReportSpecification(FeatureInstrumentationPointsReport)
+
+    def actions_for_project(
+        self, project: VProject
+    ) -> tp.MutableSequence[actions.Step]:
+
+        project.cflags += self.get_vara_feature_cflags(project)
```
vulder: With your action rework in the base class, could this code not be simplified?

varats/varats/experiments/vara/feature_perf_sampling.py
```diff
@@ -0,0 +1,197 @@
...
+    FeatureInstrType,
+)
+from varats.project.project_util import BinaryType
+from varats.project.varats_project import VProject
+from varats.report.gnu_time_report import TimeReport, TimeReportAggregate
+from varats.report.report import ReportSpecification
+
+
+class SampleWithPerfAndTime(ProjectStep):  # type: ignore
+    """See `DESCRIPTION`."""
```
vulder: same

varats/varats/experiments/vara/feature_perf_sampling.py
```diff
@@ -0,0 +1,197 @@
...
+        sampling_rate: int = 997,
+        num_iterations: int = 2
+    ):
+        super().__init__(project=project)
+        self.experiment_handle = experiment_handle
+        self.sampling_rate = sampling_rate
+        self.num_iterations = num_iterations
+
+    def __call__(self) -> StepResult:
+        vara_results_dir = get_varats_result_folder(self.project)
```
vulder:
```suggestion
        vara_results_dir = get_varats_result_folder(self.project)

```

varats/varats/experiments/vara/feature_perf_sampling.py
```diff
@@ -0,0 +1,197 @@
...
+                print(
+                    f"No workload for project={self.project.name} "
+                    f"binary={binary.name}. Skipping."
+                )
+                continue
+            if len(workloads) > 1:
+                raise RuntimeError(
+                    "Currently, only a single workload is supported. "
+                    f"project={self.project.name} binary={binary.name}"
+                )
```
vulder, line 90: Why is only one workload supported? Could we not run multiple after each other?

varats/varats/experiments/vara/feature_perf_sampling.py
```diff
@@ -0,0 +1,197 @@
...
+                                   "--user-callchains", "-o", perf_report_file,
+                                   run_cmd]
+
+                    with cleanup(workload):
+                        run_cmd()
+        return StepResult.OK
+
+
+class TranslateCallStackSamples(ProjectStep):  # type: ignore
+    """See `DESCRIPTION`."""
```
vulder: same

varats/varats/experiments/vara/feature_perf_tracing.py
```diff
@@ -0,0 +1,412 @@
...
+    though so using bpftrace is recommended except when dealing with very high
+    event frequencies.
+    """
+    NONE = 0
+    BPFTRACE = 1
+    BCC = 2
+
+
+class TraceFeaturePerfAndTime(ProjectStep):  # type: ignore
+    """See `DESCRIPTION`."""
```
vulder: same

varats/varats/experiments/vara/feature_perf_tracing.py
```diff
@@ -0,0 +1,412 @@
...
+
+    NAME = "TraceFeaturePerfWithTime"
+    DESCRIPTION = """Trace feature performance and collect execution time using
```
vulder: I think it's better to use this as a class comment, the Step description should be a small desc., as they are shown during BB execution.

varats/varats/experiments/vara/feature_perf_tracing.py
```diff
@@ -0,0 +1,412 @@
...
+                print(
+                    f"No workload for project={self.project.name} "
+                    f"binary={binary.name}. Skipping."
+                )
+                continue
+            if len(workloads) > 1:
+                raise RuntimeError(
+                    "Currently, only a single workload is supported. "
+                    f"project={self.project.name} binary={binary.name}"
+                )
```
vulder, line 112: same question as above

varats/varats/experiments/vara/feature_tracing_stats.py
```diff
@@ -0,0 +1,130 @@
...
+from varats.project.varats_project import VProject
+from varats.report.report import ReportSpecification
+from varats.tools.research_tools.vara import VaRA
+
+
+class CaptureTracingStats(actions.ProjectStep):  # type: ignore
+    """See `DESCRIPTION`."""
+
+    NAME = "CaptureTracingStats"
+    DESCRIPTION = """Collect execution statistics about the events produced
```
vulder: same

varats/varats/projects/c_projects/bzip2.py
```diff
@@ -34,9 +37,55 @@ class Bzip2(VProject):
...
+            }
+        ),
+        HTTP(
+            local="countries-land-100m.geo.json",
+            remote={
+                "1.0":
+                    "https://github.com/simonepri/geo-maps/releases/"
+                    "download/v0.6.0/countries-land-100m.geo.json"
+            }
         )
```
vulder: You can simplify this with the BB HTTPuntar source, which can download the countries archive and automatically unpack it for you.

varats/varats/projects/c_projects/bzip2.py
```diff
@@ -34,9 +37,55 @@ class Bzip2(VProject):
...
+                "--verbose",
+                "--keep",
+                "countries-land-1m.geo.json",
+                "countries-land-10m.geo.json",
+                "countries-land-100m.geo.json",
+                creates=[
+                    "countries-land-1m.geo.json.bz2",
+                    "countries-land-10m.geo.json.bz2",
+                    "countries-land-100m.geo.json.bz2"
+                ]
```
vulder: Do I understand this correctly, you want to compress all files in one pass?
This is probably related to the "only one workload" part :thinking:

varats/varats/projects/perf_tests/feature_perf_cs_collection.py
```diff
@@ -74,18 +85,17 @@ def binaries_for_revision(
             only_valid_in=RevisionRange("162db88346", "master")
         )
         binary_map.specify_binary(
-            "build/bin/SimpleSleepLoop",
+            "build/bin/SimpleBusyLoop",
             BinaryType.EXECUTABLE,
             only_valid_in=RevisionRange(
-                "c77bca4c6888970fb721069c82455137943ccf49",
-                "27f17080376e409860405c40744887d81d6b3f34"
```
vulder, line 81: Why do you remove this one? With the range ending at `27f1` it's only considered in the commit range it was valid in the repo. Do you think this binary does not/never made any sense at all? Which would be a reason to fully remove it.

varats/varats/projects/c_projects/bzip2.py
```diff
@@ -34,9 +37,55 @@ class Bzip2(VProject):
...
 
+    WORKLOADS = {
+        WorkloadSet(WorkloadCategory.MEDIUM): [
+            Command(
+                SourceRoot("bzip2") / RSBinary("bzip2"),
+                "--compress",
+                "--best",
+                "--verbose",
+                "--verbose",
+                "--verbose",
```
vulder:
```suggestion
                "--vvv",
```
same?

vulder (changes requested):
Overall this looks quite good, I only found a few nits we should clean up before merging.

varats/varats/ts_utils/click_param_types.py
```diff
@@ -174,11 +175,11 @@ def create_multi_experiment_type_choice(
     """
     initialize_experiments()
     value_dict = {
-        k: [v]
+        k: [v]  # make value a list to be consistent with entry for 'all'
         for k, v in ExperimentRegistry.experiments.items()
         if not __is_experiment_excluded(k)
     }
-    value_dict["all"] = list(value_dict.values())
+    value_dict["all"] = [v[0] for v in value_dict.values()]
```
boehmseb: Better properly flatten the list:
```suggestion
    value_dict["all"] = [experiment for value in value_dict.values() for experiment in value]
```

varats/varats/ts_utils/click_param_types.py
boehmseb: For this change, I don't understand why it should be necessary. To me it seems that the two versions are equivalent.

varats-core/varats/experiment/experiment_util.py
jonas-kaufmann: That's for the filename of the compiled binaries I copy into the results directory during the feature tracing and sampling experiments. I currently assemble a filename manually there, which I agree, doesn't make sense as we, for example, use the information which commit the binary originates from. I will change that.

varats/varats/data/reports/feature_tracing_stats_report.py
jonas-kaufmann: It has always been specific to tracing of feature performance. It provides statistics about how many tracing events are produced over the whole execution and what the distribution of the amount of time between the occurrence of these events looks like. The old name was `VaraInstrumentationStatsReport`, which I found to not express the high-level idea well enough.

varats/varats/ts_utils/click_param_types.py
jonas-kaufmann: They are equivalent but I find the original version quite hard to read / understand / counter-intuitive. The main reasons were that it was basically one large expression and that I had to take a close look at the two intermingled list comprehensions to grasp how the data flows.

varats/varats/experiments/vara/feature_experiment.py
jonas-kaufmann: These are intended as doc strings that should be visible from IntelliSense. If you put them above then at least VS Code doesn't correctly relate them.

varats/varats/projects/c_projects/gzip.py
jonas-kaufmann: I am unsure how to use this. Having done a quick Google search, I didn't find a URL to download all release files as a single zip file. Instead, people seem to use self-written scripts to download all release files one by one. There's the option to download all of the source code as a zip file but that doesn't contain the workload files.

varats/varats/experiments/vara/feature_instrumentation_points.py
jonas-kaufmann: Yes, it definitely can, thanks for pointing that out.

varats/varats/experiments/vara/feature_perf_sampling.py
jonas-kaufmann: I suppose the changes necessary to support multiple workloads would indeed be straightforward. I have not worked with multiple workloads yet and didn't want to invest the time to adjust everything I have written to work with them. This is also the case for the `TimeStatsTable`.

varats/varats/experiments/vara/feature_perf_tracing.py
jonas-kaufmann: Ah, I was always wondering what the `DESCRIPTION` was about. Makes sense, I am going to change that.

varats/varats/projects/c_projects/bzip2.py
jonas-kaufmann: I can't find the URL to the countries archive (see above), can you point me to it?

varats/varats/projects/c_projects/bzip2.py
jonas-kaufmann: The issue with bzip2 is that it compresses quite fast, even on the strongest compression setting. As such, I needed multiple input files in a single workload to get it to the 30s mark of total execution time as even the largest JSON files didn't suffice.

varats/varats/projects/perf_tests/feature_perf_cs_collection.py
jonas-kaufmann: Yes, I actually think that one shouldn't use it. In order to get interesting results, you have to sleep for a small amount of time (only a few milliseconds or even in the microsecond domain). The variation from run to run then becomes quite high. However, it would make sense to include it again when updated to use the busy sleep implementation. The clear advantage over SimpleBusyLoop would then be that you could specify a concrete amount of time to sleep for each iteration instead of how often to increment the counter.

varats/varats/projects/c_projects/bzip2.py
jonas-kaufmann: Yes, if the binary offers long arguments, I much prefer those though because they immediately spell out what their effect is. This means that when quickly glancing at the workload you don't have to do a `<binary> --help` first to understand the configuration.

varats/varats/ts_utils/click_param_types.py
jonas-kaufmann: In my opinion, the two intermingled list comprehensions are quite hard to understand and somewhat counter-intuitive. As the definition of value_dict is just above and therefore not defined somewhere outside this function, I prefer my simpler expression. If you want to have this rewritten, I would propose to do it as a for loop explicitly concatenating arrays via the `+=` operator instead.

varats/varats/experiments/vara/feature_experiment.py
vulder: That's weird, if find placing string "after" the value strings below the var as highly confusing :thinking:
1) I read the name/value and afterwards can only see the doc -> reading flow is bad
2) If the string below somehow get's formatted to the above (or the line ends in `\`) the values change. But this would, without new lines, also be the case for the above variant.

However, if this is the only option to get the docs to work, we should at least add linebreaks for the comments and an empty line inbetween.

```
class Foo(Enum):

    FOO = "f"
    """
    Foo doc
    """

    BAR = "b"
    """
    Bar doc
    """
```

varats/varats/projects/c_projects/gzip.py
vulder: I see :see_no_evil: Well, I guess then the zip version does not work.

Could you maybe write a small source class that can download and handle multiple files, such as http?

```
class MultiHTTP(
            local="example.cnf",
            remote={
                "1.0":
                    "https://github.com/simonepri/geo-maps/releases/"
                    "download/v0.6.0/"
            }
            files = ["countries-land-1km.geo.json", ...]
)
```
I mean, the core of the problem is that when we want 10 different files, we need to add 10 sources, which turn the unique proj/exp name into `...1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,...`. Something, I would like to not have :laughing: 

Maybe @simbuerg has a better idea? @Kaufi-Jonas If want, maybe the BB people would like such as contribution.

varats/varats/experiments/vara/feature_perf_sampling.py
vulder: ok, then let's keep it at 1 for the moment

varats/varats/projects/c_projects/bzip2.py
vulder: Nice for `bzip2` I guess :rofl: 

Ok, then let's keep it that way. Maybe just not your reasoning in a comment, so people don't forget why this was choosen.

varats/varats/projects/perf_tests/feature_perf_cs_collection.py
vulder: Ok, them let's completely remove it

varats/varats/projects/c_projects/bzip2.py
vulder: Isn't `-vvv` common to chain multiple verboses together? I, at least, no of no tool that uses multiple `v`s otherwise.

3x `--verbose` is kind "verbose"

varats/varats/ts_utils/click_param_types.py
boehmseb: I actually prefer the conciseness of the list comprehension here and once you know in which order the `for`s are processed by python it's not that hard to understand. But maybe, we can use @vulder's opinion as a tie-breaker here.

varats/varats/ts_utils/click_param_types.py
vulder: Nested list comprehension are quite common in python and list comprehensions are also deemed to be "pythonic". After you know how they are build, they are often super easy to read/understand.
We also use them all over the place in the TS, so I would prefer to keep it.

varats/varats/experiments/vara/feature_experiment.py
boehmseb: Unfortunately, they indeed have to be after the variable when using docstrings.
However, there is an alternative with special comments we could try.
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute

varats/varats/ts_utils/click_param_types.py
boehmseb: I don't like loops for such simple tasks, they take up so much space :smiley: 
But that's just my opinion.
I still vote for fully flattening the list since that respects the semantics of the type rather than some implementation detail on our side.

varats/varats/ts_utils/click_param_types.py
vulder: We use list comprehensions all over the place and should keep doing this.

varats/varats/projects/c_projects/gzip.py
jonas-kaufmann: Done :)

https://github.com/PolyJIT/benchbuild/pull/546

varats/varats/experiments/vara/feature_experiment.py
jonas-kaufmann: I quickly tried replacing the doc strings with comments starting with `#:  ` as @boehmseb suggested. The issue with this approach is that VS Code's Intellisense feature doesn't handle them resulting in no documentation at all being shown. In my opinion, if you don't have another idea, the best option is to stick with doc strings below the declarations. I already inserted the newlines.

varats/varats/projects/c_projects/gzip.py
jonas-kaufmann: The pull request has been merged and a benchbuild release has been created. I therefore added a commit to make use of the new `HTTPMultiple` source class. I also removed the `TODO: auto unzipper for BB?` comments.

vulder (approved):

boehmseb (approved):

PR #751, opened 2023-03-03
Ensure research tools report their installation status correctly after build

Sinerum:
resolves se-sic/VaRA#974


vulder (approved):

varats/varats/tools/research_tools/research_tool.py
```diff
@@ -519,13 +519,47 @@ def build(
         """
 
     @abc.abstractmethod
+    def install_binaries(self) -> tp.List[str]:
```
boehmseb: Maybe a bit picky, but maybe we should call this `get_install_binaries` otherwise one could think this function actually installs the binaries.

boehmseb (approved):

PR #753, opened 2023-03-13
Make LoadRepositoryForTest importable from varats

danjujan:
My [CoveragePlot](https://github.com/se-sic/VaRA-Tool-Suite/blob/2275f635df6dc0360809ffddf36fafe91374963a/varats/varats/plots/llvm_coverage_plot.py#L296) needs to checkout the state of the git repo used during the experiment, since it has to read the src files. LoadRepositoryForTest provides the needed functionality. However, it cannot be imported easyily.
Move it to the varats utils to solve this and return the `repo_path` in the `__enter__` method to avoid calling `get_local_project_git_path` in later code.

### DIFF
--- a/varats-core/varats/utils/git_commands.py
+++ b/varats-core/varats/utils/git_commands.py
@@ -1,10 +1,17 @@
 """This module contains wrapper for basic git commands."""
 import typing as tp
 from pathlib import Path
+from types import TracebackType
 
 from benchbuild.utils.cmd import git
 
-from varats.utils.git_util import get_current_branch, CommitHash
+from varats.project.project_util import get_local_project_git_path
+from varats.utils.git_util import (
+    get_current_branch,
+    CommitHash,
+    ShortCommitHash,
+    get_head_commit,
+)
 
 
 def add_remote(repo_folder: Path, remote: str, url: str) -> None:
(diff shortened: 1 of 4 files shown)


varats-core/varats/utils/git_commands.py
```diff
@@ -147,3 +154,24 @@ def download_repo(
     output = git("-C", dl_folder, args)
     for line in output.split("\n"):
         post_out(line)
+
+
+class LoadRepository():
```
boehmseb: This class doesn't really fit in this module since `git_commands` is intended for simple wrappers around git commands.
iIstead, this should be in `git_util`.
The problem is that `git_util` must not import `git_commands`.
A good solution would be rewrite this context manager using pygit2 and its checkout functionality.

boehmseb: Let's find a better name for this. Maybe something like `RepositoryAtCommit` or similar.

boehmseb (changes requested):

varats-core/varats/utils/git_commands.py
```diff
@@ -1,6 +1,7 @@
 """This module contains wrapper for basic git commands."""
 import typing as tp
 from pathlib import Path
+from types import TracebackType
```
vulder:
```suggestion
```

varats-core/varats/utils/git_commands.py
danjujan: `RepositoryAtCommit` sounds good enough for me.

vulder (approved):

varats-core/varats/utils/git_util.py
```diff
@@ -1105,3 +1106,26 @@ def branch_has_upstream(
...
+    def __init__(self, project_name: str, revision: ShortCommitHash) -> None:
+        self.__repo = pygit2.Repository(
+            get_local_project_git_path(project_name)
+        )
+        self.__initial_head = self.__repo.head
+        self.__revision = self.__repo.get(revision.hash)
+
+    def __enter__(self) -> Path:
+        self.__repo.checkout_tree(self.__revision)
+        return self.__repo.path
```
boehmseb:
```suggestion
        return Path(self.__repo.path)
```

boehmseb (approved):

PR #756, opened 2023-03-15
Add first work for author disambiguation

vulder:


vulder:
@Sinerum Did you pick those changes and integrate them somewhere? Or what is the plan to work in authors disambig?

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,162 @@
...
+            self.__authors.append(Author(self.new_author_id(), name, mail))
+            self.look_up_invalid = True
+            return
+        if len(ambiguos_authors) > 1:
+            existing_author = reduce(
+                lambda accu, author: accu.merge(author), ambiguos_authors
+            )
+        else:
+            existing_author = ambiguos_authors[0]
+        existing_author.add_data(name, mail)
```
vulder:
```suggestion
            existing_author = ambiguos_authors[0]

        existing_author.add_data(name, mail)
```

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,162 @@
...
+        self._look_up_invalid = True
+
+    def _gen_lookup_dicts(self):
+        """Generate the dicts for name and mail lookups."""
+        for author in self.__authors:
+            for mail in author.mail_addresses:
+                self.mail_dict[mail] = author
+            for name in author.names:
+                self.name_dict[name] = author
+        self._look_up_invalid = False
```
vulder:
```suggestion
                self.name_dict[name] = author

        self._look_up_invalid = False
```

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,162 @@
...
+        test = git["shortlog", "-sne", "--all"]().strip().split("\n")
+        for line in test:
+            match = name_regex.match(line)
+            if not match:
+                LOG.info("Invalid author format.")
+                continue
+            name = match.group(1)
+            email = match.group(2)
+            author_map.add_entry(name, email)
+        return author_map
```
vulder:
```suggestion
            author_map.add_entry(name, email)

        return author_map
```

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+if __name__ == '__main__':
+    a = generate_author_map(
+        "/home/simon/Documents/Vara/vara-ts/VaRA-Tool-Suite"
+    )
+    print(a)
```
vulder:
```suggestion
```

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,162 @@
...
+        ambiguos_authors = [
+            author for author in self.__authors
+            if name in author.names or mail in author.mail_addresses
+        ]
+
+        if not ambiguos_authors:
+            self.__authors.append(Author(self.new_author_id(), name, mail))
+            self.look_up_invalid = True
+            return
+        if len(ambiguos_authors) > 1:
```
vulder:
```suggestion
            return

        if len(ambiguos_authors) > 1:
```

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,162 @@
...
+    @property
+    def authors(self):
+        return self.__authors
+
+    def get_author(self, name: str, mail: str):
+        """Get an author by name and mail Throws AmbiguousAuthor exception if no
+        author matches the combination."""
+        if self._look_up_invalid:
+            self._gen_lookup_dicts()
+        if self.mail_dict[mail] == self.name_dict[name]:
```
vulder:
```suggestion
            self._gen_lookup_dicts()

        if self.mail_dict[mail] == self.name_dict[name]:
```

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def merge(self, other: ['Author']) -> ['Author']:
+        if other.id < self.id:
+            other.names.append(self.names)
+            other.mail_addresses.append(self.mail_addresses)
+            return other
+        else:
+            self.names.append(other.names)
+            self.mail_addresses.append(other.mail_addresses)
+            return self
```
vulder:
```suggestion

        self.names.append(other.names)
        self.mail_addresses.append(other.mail_addresses)
        return self
```

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,162 @@
...
+from plumbum import local
+
+LOG = logging.getLogger(__name__)
+
+
+class AmbiguousAuthor(Exception):
+    """Raised if an ambiguous author is encountered."""
+
+
+name_regex = re.compile("\s*\d+\t(.*) <(.*)>")
```
vulder:
```suggestion
name_regex = re.compile("\s*\d+\t(.*) <(.*)>")
```
@Sinerum can you make the constant upper case?

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,160 @@
...
+    def get_author(self, name: str, mail: str):
+        """Get an author by name and mail Throws AmbiguousAuthor exception if no
+        author matches the combination."""
+        if self._look_up_invalid:
+            self._gen_lookup_dicts()
+
+        if self.mail_dict[mail] == self.name_dict[name]:
+            return self.name_dict[name]
+        else:
+            raise AmbiguousAuthor
```
vulder:
```suggestion

        raise AmbiguousAuthor
```

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,160 @@
...
+    def mail_addresses(self):
+        return self.__mail_addresses
+
+    def add_data(self, name: str, mail: str):
+        if not name in self.names:
+            self.names.append(name)
+        if not mail in self.mail:
+            self.mail_addresses.append(mail)
+
+    def merge(self, other: ['Author']) -> ['Author']:
```
boehmseb: Why the `['Author']` here instead of `'Author'`?

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,160 @@
...
+    def authors(self):
+        return self.__authors
+
+    def get_author(self, name: str, mail: str):
+        """Get an author by name and mail Throws AmbiguousAuthor exception if no
+        author matches the combination."""
+        if self._look_up_invalid:
+            self._gen_lookup_dicts()
+
+        if self.mail_dict[mail] == self.name_dict[name]:
```
boehmseb: What if one or both keys cannot be found in the map?
That's a case that would also deserve a small test ;)

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,160 @@
...
+            self._gen_lookup_dicts()
+        return self.mail_dict[email]
+
+    @property
+    def authors(self):
+        return self.__authors
+
+    def get_author(self, name: str, mail: str):
+        """Get an author by name and mail Throws AmbiguousAuthor exception if no
+        author matches the combination."""
```
boehmseb: If the exception is raised if no match is found, then "ambiguous" is probably not the best name for the exception.
Ambiguous means that there are multiple possible matches.
Maybe call it `AuthorLookupError` or something along those lines.

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,160 @@
...
+
+        self._look_up_invalid = False
+
+    def __repr__(self):
+        return f"{self.name_dict} \n {self.mail_dict}"
+
+
+def generate_author_map(path: Path) -> AuthorMap:
+    """Generate an AuthorMap for the repository at the given path."""
+    with local.cwd(path):
```
boehmseb: Use `git -C` instead of `local.cwd`

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,160 @@
...
+        author_map = AuthorMap()
+        test = git["shortlog", "-sne", "--all"]().strip().split("\n")
+        for line in test:
+            match = name_regex.match(line)
+            if not match:
+                LOG.info("Invalid author format.")
```
boehmseb: In what cases does this happen? Wouldn't be a warning more appropriate?

tests/mapping/test_author_map.py
```diff
@@ -0,0 +1,29 @@
...
+
+class TestAuthorMap(unittest.TestCase):
+
+    def test_get_author_by_email(self):
+        initialize_projects()
+        git_path = get_local_project_git_path("xz")
+        amap = generate_author_map(git_path)
+        self.assertEqual(
+            amap.get_author_by_email("jim@meyering.net"),
+            Author(19, "Jim Meyering", "meyering@redhat.com")
```
boehmseb: Checking the Id seems very brittle as it can easily change when new commits are made.
Also, you have implemented equality on authors such that it only checks for the id anyway.
Therefore, you should check whether the lists of names and email addresses match instead on using the equality.

boehmseb (changes requested):
In addition to my other remarks, there are a lot of return-type annotations missing.

varats-core/varats/mapping/author_map.py
Sinerum: Acctualy the docu is wrong the error is thrwon only when multiple authors match the combination. If there is no entry in either map a LookupError is thrown by the dict.

varats-core/varats/mapping/author_map.py
Sinerum: I am unsure what should happen in that case for now a Look up error is thrown but we may want change that to either return an author if either map has one and only throw an error if both do not contain a matching author

boehmseb: If you explicitly look up using name and email and we cannot find something, then I think we should return `None`. Same goes for the `by_name` and `by_email` versions btw.
I'm not a fan of throwing exceptions around in such a case.

vulder: For me, not finding a name is expected behavior for such an API, hence returning an optional (i.e., `None`) seems like the most telling design. Especially, as a caller I see from the func signature that the author I'm looking might not be there.

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,165 @@
...
+        Returns None if no author or multiple authors were found matching the
+        combination of name and mail-address.
+        """
+        if self._look_up_invalid:
+            self._gen_lookup_dicts()
+        mail_author = self.mail_dict.get(mail, None)
+        if mail_author == (name_author := self.name_dict.get(name, None)):
+            return name_author
+        else:
+            return None
```
vulder:
```suggestion

        return None
```
@Sinerum You don't need an else if you return within the then clause

varats-core/varats/mapping/author_map.py
```diff
@@ -0,0 +1,159 @@
...
+
+        Returns None if no author or multiple authors were found matching the
+        combination of name and mail-address.
+        """
+        if self._look_up_invalid:
+            self._gen_lookup_dicts()
+        mail_author = self.mail_dict.get(mail, None)
+        if mail_author == (name_author := self.name_dict.get(name, None)):
+            return name_author
+        return None
```
vulder:
```suggestion
        if self._look_up_invalid:
            self._gen_lookup_dicts()

        mail_author = self.mail_dict.get(mail, None)
        if mail_author == (name_author := self.name_dict.get(name, None)):
            return name_author

        return None
```

boehmseb (approved):

Sinerum (approved):

PR #777, opened 2023-03-25
Optimizes costly imports at run time

jonas-kaufmann:
These are the performance optimizations on the VaRA-TS side for https://github.com/se-sic/VaRA/issues/1003. Together with https://github.com/PolyJIT/benchbuild/pull/555, I achieved roughly a 2x improvement in execution time when invoking the wrapped clang.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ requests_cache.sqlite
 env
 venv
 build/
+tmp/
 .tox/
 docs/source/.varats.yaml
 docs/**/*.inc
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,8 +22,13 @@ repos:
     hooks:
     - id: docformatter
       args: ['--in-place', '--wrap-summaries=80', '--wrap-descriptions=80', '--pre-summary-newline']
+-   repo: https://github.com/MarcoGorelli/auto-walrus
+    rev: v0.2.2
+    hooks:
+    -   id: auto-walrus
 -   repo: https://github.com/pre-commit/mirrors-yapf
     rev: 'v0.32.0'  # Use the sha / tag you want to point at
     hooks:
     -   id: yapf
+
 exclude: ^tests/TEST_INPUTS/.*$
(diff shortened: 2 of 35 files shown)


danjujan (approved):
LGTM and reduces xz's compile time by 20% in my case.

jonas-kaufmann:
Together with the optimizations on the benchbuild side, you should be seeing more. One key change that I added to https://github.com/PolyJIT/benchbuild/pull/555 yesterday is to prevent executing database code and related imports if not required. You probably need to add the entry to your benchbuild config to disable the database or regenerate it via `vara-gen-bbconfig`.

You need the following:
```yaml
db:
    [...]
    enabled:
        default: false
        desc: Whether the database is enabled.
        value: false
    [...]
```

danjujan:
You are right. I forgot the benchbuild side. After running
```bash
pip install --upgrade git+https://github.com/Kaufi-Jonas/benchbuild.git@j-optimize_perf
vara-gen-bbconfig --update --test-projects
```
my xz build time dropped to 73s. Originally it took about 200s.
Thank you very much!

vulder:
Most of the gains come from the BB side, hence we should not add to many local imports to the tool suite. I think the research_tool ones are not really needed. For project/report related ones I can see us making an exception if it's  necessary.

jonas-kaufmann:
Agreed. Those are actually an artifact and don't make any difference. These changes are from when I started working on this and was trying some things out. I never intended to push these but somehow they slipped through.

The remaining local imports are scipy, numpy, and pygit2, ordered descending starting with worst import time. Are you okay with keeping these in?

jonas-kaufmann:
Oh, and we still need a benchbuild release. With this, we could close https://github.com/se-sic/VaRA/issues/1003 together with this PR.

vulder (commented):
project_util and gnu_time_report looks good, for git_util i'm not sure if we can ensure that this change is keept

jonas-kaufmann:
The import time for pygit2 is just 15 ms on my machine so I just reverted the changes for that. We also have a benchbuild release now. From my side, we can merge this now.

boehmseb (approved):

vulder (approved):

PR #780, opened 2023-03-27
Add missing labels to workloads in FeaturePerfCSCollection

danjujan:
https://github.com/se-sic/VaRA-Tool-Suite/pull/748 added new workloads without labels for FeaturePerfCSCollection. However, `create_workload_specific_filename` expects labels https://github.com/se-sic/VaRA-Tool-Suite/blob/06c2d3466a862039f5e08361e1899a15750a1593/varats-core/varats/experiment/workload_util.py#L106 and fails if `cmd.label` is `None`.
It might be a good idea to add a check for missing labels when creating a workload. But for now just define some labels.

### DIFF
--- a/varats/varats/projects/perf_tests/feature_perf_cs_collection.py
+++ b/varats/varats/projects/perf_tests/feature_perf_cs_collection.py
@@ -56,14 +56,21 @@ class FeaturePerfCSCollection(VProject):
             ),
             Command(
                 SourceRoot("FeaturePerfCSCollection") /
-                RSBinary("SimpleFeatureInteraction"), "--enc", "--compress"
+                RSBinary("SimpleFeatureInteraction"),
+                "--enc",
+                "--compress",
+                label="SFI-enc-compress"
             )
         ],
         WorkloadSet(WorkloadCategory.MEDIUM): [
             Command(
                 SourceRoot("FeaturePerfCSCollection") /
-                RSBinary("SimpleBusyLoop"), "--iterations", str(10**7),
-                "--count_to", str(5 * 10**3)
+                RSBinary("SimpleBusyLoop"),
+                "--iterations",
+                str(10**7),
+                "--count_to",
+                str(5 * 10**3),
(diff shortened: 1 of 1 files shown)


varats/varats/projects/perf_tests/feature_perf_cs_collection.py
```diff
@@ -56,14 +56,21 @@ class FeaturePerfCSCollection(VProject):
...
             Command(
                 SourceRoot("FeaturePerfCSCollection") /
-                RSBinary("SimpleBusyLoop"), "--iterations", str(10**7),
-                "--count_to", str(5 * 10**3)
+                RSBinary("SimpleBusyLoop"),
+                "--iterations",
+                str(10**7),
+                "--count_to",
+                str(5 * 10**3),
+                label="SBL-iterations-count-to"
```
jonas-kaufmann: Thanks for catching this.

```suggestion
                label="SBL-iterations-10M-count-to-5K"
```
SimpleBusyLoop only has these two configuration options and both are required. They directly influence how many feature tracing events SBL will generate together with their frequency. I therefore recommend that we add this information to the label.

jonas-kaufmann (changes requested):
Otherwise LGTM

jonas-kaufmann (approved):

vulder (approved):

vulder:
> #748 added new workloads without labels for FeaturePerfCSCollection. However, `create_workload_specific_filename` expects labels
> 
> https://github.com/se-sic/VaRA-Tool-Suite/blob/06c2d3466a862039f5e08361e1899a15750a1593/varats-core/varats/experiment/workload_util.py#L106
> 
> and fails if `cmd.label` is `None`.
> It might be a good idea to add a check for missing labels when creating a workload. But for now just define some labels.

I'll relay this to the BB devs. For me a label would always make sense, if no label is provided we could just take the name of the command as a label. In the end, everything should have some form of printable name.

@simbuerg

PR #784, opened 2023-04-12
Generate Feature-Blame-Report

sisteu56:


varats/varats/projects/cpp_projects/commit_feature_interactions_test_repo.py
```diff
@@ -0,0 +1,53 @@
...
+class CFI(VProject):  # type: ignore
+    """test repo for commit-feature interactions."""
+    NAME = 'CFI'  # The name of the project
+    GROUP = 'cpp_projects'  # The group this project belongs to
+    DOMAIN = ProjectDomains.TEST  # The application domain of this project
+
+    SOURCE = [
+        PaperConfigSpecificGit(
+            project_name="CFI",
+            remote="https://github.com/sisteu56/cfi_test_repo.git",
```
boehmseb: We should integrate this into our test repo infrastructure.
@sisteu56 we can discuss this in our next meeting.

varats/varats/projects/cpp_projects/commit_feature_interactions_test_repo.py
```diff
@@ -0,0 +1,53 @@
...
+            refspec="main",
+            limit=None,
+            shallow=False
+        )
+    ]
+
+    test_files = ["main.cpp", "helper.cpp"]
+
+    @property
+    def binaries(self) -> tp.List[ProjectBinaryWrapper]:
```
boehmseb: You should override the function `binaries_for_revision` instead of this one.
This function has the advantage that it can declare binaries per revision.

PS: could you tell me where you have copied this from? We should fix that location so that it isn't used as a reference in the future.

varats/varats/data/reports/feature_blame_report.py
```diff
@@ -0,0 +1,271 @@
...
+from varats.report.report import BaseReport
+
+
+class FeatureCommitRegion():
+    """Data containing feature and commit regions."""
+
+    def __init__(self, feature: str, commit_hash: str, repo: str) -> None:
+        self.__feature = feature
+        self.__commit_hash = commit_hash
+        self.__repo = repo
```
boehmseb: Use `CommitRepoPair` from `git_util.py` instead.

varats/varats/data/reports/feature_blame_report.py
```diff
@@ -0,0 +1,271 @@
...
+
+    @staticmethod
+    def create_feature_commit_region(
+        raw_region_data: tp.Dict[str, tp.Any]
+    ) -> 'FeatureCommitRegion':
+        """Creates a 'FeatureCommitRegion' from the corresponding yaml document
+        section."""
+        feature = raw_region_data["feature"]
```
boehmseb: On the C++ side you have a vector of features and store it in the yaml with the key `features`. Here you only expect one feature with the key `feature` instead.

varats/varats/data/reports/feature_blame_report.py
```diff
@@ -0,0 +1,271 @@
...
+        return FeatureCommitRegionInstruction(
+            instruction, location, feature_commit_regions
+        )
+
+    def print(self) -> None:
+        """'FeatureCommitRegionInstruction' prints itself."""
+        print("    -FEATURE COMMIT REGION INSTRUCTION")
+        print("      -INSTRUCTION: " + self.__instruction)
+        print("      -LOCATION: " + self.__location)
+        for FeatureCommitRegion in self.__feature_commit_regions:
```
boehmseb: You cannot name a variable the same as an existing class. Use `feature_commit_region` instead.
Note that the naming conventions for variables in Python (`snake_case`) are different from the ones in VaRA/LLVM (`PascalCase`).

varats/varats/data/reports/feature_blame_report.py
```diff
@@ -0,0 +1,271 @@
...
+                FeatureCommitRegionInstruction.
+                create_feature_tainted_instruction(raw_inst_entry)
+            )
+        return FeatureBlameResultFunctionEntry(name, demangled_name, inst_list)
+
+    def print(self) -> None:
+        """'FeatureBlameResultFunctionEntry' prints itself."""
+        print("  -FEATURE BLAME RESULT FUNCTION ENTRY")
+        print("    -FUNCTION: " + self.demangled_name)
+        for FeatureCommitRegionInstruction in self.__feature_commit_region_insts:
```
boehmseb: Same as above

varats/varats/data/reports/feature_blame_report.py
```diff
@@ -0,0 +1,271 @@
...
+                        raw_feature_blame_report['result-map'][raw_func_entry]
+                    )
+                )
+                self.__function_entries[new_function_entry.name
+                                       ] = new_function_entry
+
+    def print(self) -> None:
+        """'FeatureBlameReport' prints itself."""
+        print("FEATURE BLAME REPORT")
+        for FeatureBlameResultFunctionEntry in self.__function_entries.values():
```
boehmseb: Same as above.

varats/varats/data/reports/feature_blame_report.py
```diff
@@ -0,0 +1,271 @@
...
+    ) -> FeatureBlameResultFunctionEntry:
+        """
+        Get the result entry for a specific function.
+
+        Args:
+            mangled_function_name: mangled name of the function to look up
+        """
+        return self.__function_entries[mangled_function_name]
+
+        # def get_feature_locations_dict(self) -> tp.Dict[str, tp.Set[str]]:
```
boehmseb: Always comment out the whole function to avoid type errors. This includes the docstring since this is also a statement in python.

boehmseb (changes requested):

varats/varats/projects/cpp_projects/commit_feature_interactions_test_repo.py
sisteu56: I copied it from here https://vara.readthedocs.io/en/vara-dev/vara-ts-api/projects.html#projects

varats/varats/data/reports/feature_blame_report.py
sisteu56: That's because I changed the FeatureCommitRegion class on the C++ side last week, but forgot to push it

varats/varats/experiments/vara/feature_blame_experiment.py
```diff
@@ -0,0 +1,73 @@
+"""Implements the base feature blame experiment, making it easier to create
```
boehmseb: This is all very similar to the module `blame_experiment.py`.
I suggest merging these two and renaming the module to something like `vara_experiments.py`.
@vulder what's your take on this?

sisteu56: Should I rename generate_basic_feature_blame_experiment_actions and generate_basic_blame_experiment_actions into one function generate_basic_vara_experiment_actions or something like this?

varats/varats/experiments/vara/vara_experiments.py
```diff
@@ -51,6 +52,37 @@ def setup_basic_blame_experiment(
...
+        << RunWLLVM() \
+        << run.WithTimeout()
+
+    # Add own error handler to compile step.
+    project.compile = get_default_compile_error_wrapped(
+        experiment.get_handle(), project, report_type
+    )
+
+    # This c-flag is provided by VaRA and it suggests to use the git-blame
+    # annotation.
```
boehmseb:
```suggestion
    # These flags are provided by VaRA and suggest to use git-blame
    # and feature annotations.
```

varats/varats/experiments/vara/vara_experiments.py
```diff
@@ -71,3 +103,25 @@ def generate_basic_blame_experiment_actions(
     return get_bc_cache_actions(
         project, bc_file_extensions, extraction_error_handler
     )
+
+
+def generate_basic_feature_blame_experiment_actions(
```
boehmseb: That's the same as `generate_basic_blame_experiment_actions`, right?
In that case, you should reuse the existing function.

varats/varats/data/reports/feature_blame_report.py
```diff
@@ -0,0 +1,190 @@
...
+        return self.__feature
+
+    @property
+    def commit(self) -> CommitRepoPair:
+        """commit of this cfi."""
+        return self.__commit
+
+    def is_terminator(self) -> bool:
+        br_regex = re.compile(r'(br( i1 | label ))|(switch i\d{1,} )')
+        return br_regex.search(self.__feature) is not None
```
boehmseb: What do you need this for? I don't think this is a reliable method to identify terminator instructions.

varats/varats/data/reports/feature_blame_report.py
```diff
@@ -0,0 +1,190 @@
...
+        return self.__feature
+
+    @property
+    def commit(self) -> tp.List[CommitRepoPair]:
+        """commit of this cfi."""
+        return self.__commits
+
+    def is_terminator(self) -> bool:
+        br_regex = re.compile(r'(br( i1 | label ))|(switch i\d{1,} )')
+        return br_regex.search(self.__feature) is not None
```
boehmseb: Same as above.

boehmseb (changes requested):
Almost there. A couple of things to clean up left.

varats/varats/data/reports/feature_blame_report.py
sisteu56: It's not needed, it was just left from copying functions from the FeatureAnalysisReport. 
I will remove it

boehmseb (approved):

PR #787, opened 2023-04-25
Avoids using host number of jobs in container.

boehmseb:


vulder (approved):
We just need to think about adding experiment specific slurm options should an experiment require a minimum amount of jobs, but otherwise this looks fine.

PR #788, opened 2023-04-26
F feature blame test repo

sisteu56:


varats/varats/projects/test_projects/feature_test_repo.py
```diff
@@ -70,3 +70,55 @@ def compile(self) -> None:
             bb.watch(cxx_compiler)("main.cpp", "-o", "main")
 
             verify_binaries(self)
+
+class FeatureBlameTestRepo(VProject):
```
boehmseb: We usually name the project the same as the yaml file responsible for creating the repo.
I'm not sure why this convention was not used for the other project in this file.
Anyway, every occurrence of `FeatureBlameTestRepo` or `CommitFeatureInteractionRepo` should be replaced with `CommitFeatureInteractionExample`. I think this is a way better name for the project.

varats/varats/projects/test_projects/feature_test_repo.py
```diff
@@ -70,3 +70,55 @@ def compile(self) -> None:
...
+    ]
+
+    WORKLOADS = {
+        WorkloadSet(WorkloadCategory.EXAMPLE): [
+            Command(
+                SourceRoot("CommitFeatureInteractionRepo") / RSBinary("main"),
+                label="main-no-input"
+            ),
+        ]
+    }
```
boehmseb: We didn't create any workloads for this project yet, so you should delete this.
Workloads are used for performance measurements---which we do not consider in your work.

varats/varats/projects/test_projects/feature_test_repo.py
```diff
@@ -70,3 +70,55 @@ def compile(self) -> None:
...
+        return binary_map[revision]
+
+    def run_tests(self) -> None:
+        pass
+
+    def compile(self) -> None:
+        """Compile the example project."""
+        source = local.path(self.source_of_primary)
+
+        self.cflags += ["-fno-exceptions"]
```
boehmseb, line 110: Do you really need this?
@vulder what was the reason to include this in the other feature test project?

boehmseb (changes requested):

sisteu56:
Docs CI failed nachdem ich heute einen Commit gepusht hatte bei dem eine leere Zeile von der pre-commit hook hinzugefügt wurde.
Davor war Docs CI succesful.

boehmseb:
That fail has nothing to do with your change. A package we use got an update which broke our docs. I pushed a fix here: #793

varats/varats/projects/test_projects/feature_test_repo.py
vulder: This makes debugging and analyzing test projects much easier, as it removes strange control-flow for exceptions. It's not 100% needed but probably good to keep this here.

vulder (approved):

boehmseb (approved):

PR #799, opened 2023-06-06
Fix multi case study plots and tables

Sinerum:
Plots and tables containing multiple case studies no longer crash when generating the filename.
Their file name is now based on all projects in the provided case studies

### DIFF
--- a/varats/varats/table/table.py
+++ b/varats/varats/table/table.py
@@ -153,7 +153,8 @@ class Table:
             if isinstance(case_study, varats.paper.case_study.CaseStudy):
                 table_ident = f"{case_study.project_name}_{case_study.version}_"
             else:
-                table_ident = f"{reduce(lambda x, y: f'{x}{y.project_name}_', case_study, '')}"
+                table_ident = \
+                    f"{reduce(lambda x, y: f'{x}{y.project_name}_', case_study, '')}"
         elif 'project' in self.table_kwargs:
             table_ident = f"{self.table_kwargs['project']}_"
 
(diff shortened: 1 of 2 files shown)


boehmseb (approved):

vulder (changes requested):
Fix looks good but could you please remove the pylint errors before merging :)

varats/varats/table/table.py
```diff
@@ -154,7 +154,7 @@ def table_file_name(self, table_format: TableFormat) -> str:
                 table_ident = f"{case_study.project_name}_{case_study.version}_"
             else:
                 table_ident = \
-                f"{reduce(lambda x, y: f'{x}{y.project_name}_', case_study, '')}"
+            f"{reduce(lambda x, y: f'{x}{y.project_name}_', case_study, '')}"
```
vulder: Sorry to be a bit picky here, but we should the string or somehow else make the line fit. The problem with this formatting ist, da it looks like the `f` string is on the level of the outer if case.

Sinerum: I don't like this either, the only options i see is to shorten the name of the local variable case_study to something like cs_list or extract this String into its own method.

vulder: You can also just move the reduce out of the f-string and create a local variable. That seems like the simplest solution

Sinerum: Or even better just remove the f string as the reduce returns a string 🤦

vulder (approved):

PR #801, opened 2023-06-13
Adapts git util error message and includes more file extensions

Sinerum:
Includes the provided commit hash in the error message for a to short commit hash when creating a CommitHash object.
Adds .cc as possible extension for c++ files.

### DIFF
--- a/varats-core/varats/utils/git_util.py
+++ b/varats-core/varats/utils/git_util.py
@@ -37,7 +37,9 @@ class CommitHash(abc.ABC):
 
     def __init__(self, short_commit_hash: str):
         if not len(short_commit_hash) >= self.hash_length():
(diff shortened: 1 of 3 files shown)


varats-core/varats/utils/git_util.py
```diff
@@ -37,7 +37,7 @@ class CommitHash(abc.ABC):
 
     def __init__(self, short_commit_hash: str):
         if not len(short_commit_hash) >= self.hash_length():
-            raise ValueError("Commit hash too short")
+            raise ValueError(f"Commit hash too short {short_commit_hash}")
```
```suggestion
            raise ValueError(f"Commit hash too short, only got {short_commit_hash}")
```

vulder (approved):

PR #802, opened 2023-06-26
Fixes author merging and adds tests for author merging

Sinerum:
Fix the merging of authors when a new entry provides a link between existing authors. And add tests for this.
Also change author map generation to take a project name instead of a path, to conform with commit_map generation.


boehmseb (approved):
lgtm.

Could you please create an issue for the breaking docs?

PR #806, opened 2023-07-08
Implements feature based perf instrumentation comparision

vulder:


varats-core/varats/report/linux_perf_report.py
```diff
@@ -65,9 +65,10 @@ def __parse_ctx_switches(line: str) -> int:
         return int(line.split(" ")[0].replace(",", ""))
 
     @staticmethod
-    def __parse_branch_misses(line: str) -> tp.Optional[int]:
+    def __parse_branch_misses(line: str) -> int:
+        # TODO: fix return type
```
vulder:
```suggestion
```

varats/varats/experiments/vara/feature_perf_runner.py
```diff
@@ -83,7 +83,9 @@ def actions_for_project(
     ) -> tp.MutableSequence[actions.Step]:
         project.cflags += self.get_vara_feature_cflags(project)
 
-        project.cflags += self.get_vara_tracing_cflags(FeatureInstrType.TEF)
+        project.cflags += self.get_vara_tracing_cflags(
+            FeatureInstrType.TEF, instruction_threshold=1
```
LuAbelt, line 87: Is there a specific reason to choose 1 here, instead of 0?

varats/varats/experiments/vara/feature_experiment.py
```diff
@@ -342,6 +342,10 @@ def run_traced_code(self) -> StepResult:
                                 flush=True
                             )
                             with cleanup(prj_command):
+                                pb_cmd = pb_cmd[get_extra_config_options(
+                                    self.project
+                                )]
+                                print(f"{pb_cmd=}")
```
boehmseb: Is this an overlooked debug print?

varats/varats/data/databases/feature_perf_precision_database.py
```diff
@@ -0,0 +1,932 @@
...
+            f"Could not find matching start for Event {repr(closing_event)}."
+        )
+
+        return None
+
+    found_missing_open_event = False
+    for trace_event in tef_report.trace_events:
+        if trace_event.category == "Feature":
+            if trace_event.event_type == TraceEventType.DURATION_EVENT_BEGIN:
+                # open_events.append(trace_event)
```
boehmseb:
```suggestion
```

varats/varats/data/databases/feature_perf_precision_database.py
```diff
@@ -0,0 +1,932 @@
...
+        report_type: tp.Type[BaseReport]
+    ) -> None:
+        self.__name = name
+        self.__experiment = experiment
+        self.__overhead_experiment = overhead_experiment
+        self.__report_type = report_type
+
+    @property
+    def name(self) -> str:
+        """Hame of the profiler used."""
```
boehmseb:
```suggestion
        """Name of the profiler used."""
```

varats/varats/data/databases/feature_perf_precision_database.py
```diff
@@ -0,0 +1,932 @@
...
+        return self.__name
+
+    @property
+    def experiment(self) -> tp.Type[FeatureExperiment]:
+        """Experiment used to produce this profilers information."""
+        return self.__experiment
+
+    @property
+    def overhead_experiment(self) -> tp.Type[FeatureExperiment]:
+        """Experiment used to produce this profilers information."""
```
boehmseb: Why is this description the same as for `experiment`?

varats/varats/experiments/vara/feature_perf_precision.py
```diff
@@ -0,0 +1,1201 @@
...
+
+
+def get_threshold(project: VProject) -> int:
+    """Get the project specific instrumentation threshold."""
+    if project.DOMAIN is ProjectDomains.TEST:
+        if project.name in [
+            "SynthSAFieldSensitivity", "SynthIPRuntime", "SynthIPTemplate",
+            "SynthIPTemplate2", "SynthIPCombined"
+        ]:
+            print("Don't instrument everything")
```
boehmseb: Debug print?

varats/varats/experiments/vara/feature_perf_precision.py
```diff
@@ -0,0 +1,1201 @@
...
+        """
+        return setup_actions_for_vara_experiment(
+            self,
+            project,
+            FeatureInstrType.USDT,
+            RunBCCTracedWorkloads  # type: ignore[type-abstract]
+        )
+
+
+class RunBackBoxBaseline(OutputFolderStep):  # type: ignore
```
boehmseb: I assume this should be `RunBlackBoxBaseline`?

varats/varats/plots/feature_perf_precision.py
```diff
@@ -0,0 +1,427 @@
...
+        raise UnsupportedOperation
+
+
+class PerfProfDistPlotGenerator(
+    PlotGenerator, generator_name="fperf-precision-dist", options=[]
+):
+    """Generates performance distribution plot."""
+
+    def generate(self) -> tp.List[Plot]:
+
```
boehmseb:
```suggestion
```

varats/varats/experiments/vara/feature_perf_precision.py
```diff
@@ -0,0 +1,1201 @@
...
+def perf_prec_workload_commands(
+    project: VProject, binary: ProjectBinaryWrapper
+) -> tp.List[ProjectCommand]:
+    """Uniformly select the workloads that should be processed."""
+
+    wl_commands = []
+
+    if not project.name.startswith(
+        "SynthIP"
+    ) and project.name != "SynthSAFieldSensitivity":
```
boehmseb, line 64: It's annoying that we have to break the abstraction between projects and experiments here (and also later in the tables). But I also don't have a good practical idea how to avoid this...

boehmseb (changes requested):
Still needs some cleanup (see pylint and mypy complaints).

varats/varats/experiments/vara/feature_perf_precision.py
vulder: I could not avoid this for all cases, some I got removed but depending on what one wants, it's difficult.

boehmseb (approved):

varats/varats/data/databases/feature_perf_precision_database.py
```diff
@@ -0,0 +1,932 @@
...
+    ) -> bool:
+        """Implements t-test."""
+        ttest_res = ttest_ind(old_values, new_values)
+
+        if ttest_res.pvalue < 0.05:
+            return True
+
+        return False
+
+    def __cliffs_delta(  # pylint: disable=W0238
```
LuAbelt, line 206: If I see correctly this is currently not used anywhere, right? I assume it is part of the base class such that people can use it when overriding `is_regression` or one of the `*_regression_check` methods?

varats/varats/data/databases/feature_perf_precision_database.py
```diff
@@ -0,0 +1,932 @@
...
+        get_case_study_file_name_filter(case_study),
+        config_id=0
+    )
+
+    if len(report_files) > 1:
+        raise AssertionError("Should only be one")
+    if not report_files:
+        print(
+            f"Could not find profiling data for {case_study.project_name}"
+            ". config_id=0, profiler=Baseline"
```
LuAbelt, line 512:
```suggestion
            f". config_id={config_id}, profiler=Baseline"
```
See above

varats/varats/data/databases/feature_perf_precision_database.py
```diff
@@ -0,0 +1,932 @@
...
+
+def get_patch_names(case_study: CaseStudy) -> tp.List[str]:
+    """Looks up all patch names from the given case study."""
+    report_files = get_processed_revisions_files(
+        case_study.project_name,
+        fpp.BlackBoxBaselineRunner,
+        fpp.MPRTimeReportAggregate,
+        get_case_study_file_name_filter(case_study),
+        config_id=0
+    )
```
LuAbelt, line 505: Wouldn't it be better to also make the `config_id` a parameter?
```suggestion
def get_patch_names(case_study: CaseStudy, config_id: int = 0) -> tp.List[str]:
    """Looks up all patch names from the given case study."""
    report_files = get_processed_revisions_files(
        case_study.project_name,
        fpp.BlackBoxBaselineRunner,
        fpp.MPRTimeReportAggregate,
        get_case_study_file_name_filter(case_study),
        config_id=config_id
    )
```
There might be scenarios in which you have different patches for different configs (At least this was the case in my thesis)

varats/varats/experiments/vara/feature_perf_precision.py
```diff
@@ -0,0 +1,1204 @@
...
+        experiment.get_handle(),
+        experiment.get_handle().report_spec().main_report, project, binary,
+        get_current_config_id(project)
+    )
+
+    patch_provider = PatchProvider.get_provider_for_project(type(project))
+    patches = patch_provider.get_patches_for_revision(
+        ShortCommitHash(project.version_of_primary)
+    )[IDENTIFIER_PATCH_TAG]
+    print(f"{patches=}")
```
LuAbelt: Is this intended to still be here or was this just a debug print?

varats/varats/plots/feature_perf_precision.py
```diff
@@ -0,0 +1,426 @@
...
+        return [PerfPrecisionPlot(self.plot_config, **self.plot_kwargs)]
+
+
+class PerfPrecisionDistPlot(Plot, plot_name='fperf_precision_dist'):
+    """Precision plot that plots the precision and recall distributions of
+    different profilers."""
+
+    def plot(self, view_mode: bool) -> None:
+        case_studies = get_loaded_paper_config().get_all_case_studies()
```
LuAbelt, line 82: See below

varats/varats/plots/feature_perf_precision.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def calc_missing_revisions(
+        self, boundary_gradient: float
+    ) -> tp.Set[FullCommitHash]:
+        raise UnsupportedOperation
+
+
+class PerfProfDistPlotGenerator(
+    PlotGenerator, generator_name="fperf-precision-dist", options=[]
+):
```
LuAbelt, line 157: Do you think it would be useful/helpful to have another another generator that can generate these plots for individual case studies?

varats/varats/projects/cpp_projects/hyteg.py
```diff
@@ -109,6 +108,8 @@ def compile(self) -> None:
 
         mkdir("-p", hyteg_source / "build")
 
+        update_all_submodules(hyteg_source, recursive=True, init=True)
```
LuAbelt, line 111: Technically, `HyTeG` also performs a submodule update through its CMakeList, but it cannot hurt to keep this in, if at any point they decide to change this.

varats/varats/data/databases/feature_perf_precision_database.py
vulder: Yes, it's not used. My idea was to have all test methods t-test/clifss-delta/u-test etc. follow an "indirect" interface so one can swap them in/out for testing purposes.

varats/varats/experiments/vara/feature_perf_runner.py
vulder: I think there was one, not 100% sure but if my mind does not play tricks on me it was to ignore super small regions, e.g., selects. Selects are one inst and usually optimized on the backend side, hence measuring them (even when we want to measure everything) is not a good idea.

In the future, this should be cleaned up on the llvm side by implementing a better "should I instrument this region" API.

varats/varats/data/databases/feature_perf_precision_database.py
vulder: Hmm, I'm not sure if that can/should happen here in this case. In generell, I would agree with you, however here we fetch the patch names for our regression patches that we compare in the experiment. If the returned names would differ, the experiment setup would have a problem. So, I'm not sure if we should make the config_id a variable part here.

varats/varats/plots/feature_perf_precision.py
vulder: Yes, actual something like that would make analyzing individual case studies easier. I'm not planning to include this in this PR but we can extend/refactor this for future use and have 1) generator for "all" and another one that generates one plot per case study.

varats/varats/projects/cpp_projects/hyteg.py
vulder: If I remember correctly, I had some issue without this but I'm not sure if it is still necessary. But maybe let's keep it as it does not hurt.

varats/varats/plots/feature_perf_precision.py
vulder: I already have a draft :smile: 
I'll post the PR after this one is merged

vulder: Ok, two generators are actually a stupid idea :laughing:, we can make do with one. The plot interface allows one to specify `all` as case study, which then is the same as the "original plot" :rofl:

PR #809, opened 2023-07-11
Implements classification evaluation helper

vulder:


boehmseb (commented):
lgtm

varats/varats/data/metrics.py
```diff
@@ -132,3 +132,122 @@ def min_max_normalize(values: pd.Series) -> pd.Series:
...
+
+
+class ConfusionMatrix:
+    """
+    Helper class to automatically calculate classification results.
+
+                        |  Predicted Positive (PP)  |  Predicted Negative (PN)
+    --------------------|---------------------------|--------------------------
+    Actual Positive (P) |  True Positive      (TP)  |  False Negative     (FN)
+    Actual Negative (N) |  False Positive     (FP)  |  True Negative      (TN)
```
danjujan, line 147: I would find it useful if one could visualize the ConfusionMatrix easily.
Does it make sense to add a `__str__` method in a similar style that also includes the interpretations?
Maybe we could use [tabulate](https://pypi.org/project/tabulate/) for this to be able to generate tables in different formats, e.g. for LaTeX.

danjujan (changes requested):

varats/varats/data/metrics.py
```diff
@@ -132,3 +132,122 @@ def min_max_normalize(values: pd.Series) -> pd.Series:
...
+
+    ###################
+    # Combined values
+
+    @property
+    def TP(self) -> int:  # pylint: disable=C0103
+        return len(
+            set(self.__actual_positive_values
+               ).intersection(self.__predicted_positive_values)
+        )
```
danjujan:
```suggestion
    def TP(self) -> tp.Set[tp.Any]:  # pylint: disable=C0103
        return set(self.__actual_positive_values
               ).intersection(self.__predicted_positive_values)
```
We should add a way to retrieve the classified elements instead of just their number. Either we could change the interface as suggested or introduce seperate methods.

vulder: Why would you need the elements? I did not plan to allow users to access them, as (1) the values should be enough for a eval and (2) during construction, the user already passes correctly attributed lists to us.
What you be your use case where you need access to the elements?

danjujan: I would like to inspect/print the FP and FN elements. If we don't expose them, I have to build the intersections myself. That duplicates code and kind of defeats the purpose of using this confusion matrix implementation. I.e., if I have to classify the elements myself anyway, why should I put them again in the attributed lists just to be able to instantiate a `ConfusionMatrix`? If I have my TP/TN/FP/FN lists already, I can use these directly to compute the needed interpretations.

vulder: Yes, but you don't have to implement all the different values. If you need the TP values a simple intersection is enough :thinking:

boehmseb (approved):
Seems like I just left a comment last time instead of approving...

danjujan (approved):

PR #810, opened 2023-07-13
Add check for missing BC_files folder

LeonievonMann:
se-sic/VaRA#1027


varats-core/varats/utils/settings.py
```diff
@@ -317,6 +317,7 @@ def create_missing_folder_for_cfg(
         config_node = local_cfg[cfg_varname]
         if config_node.has_value() and\
                 config_node.value is not None and\
+                not isinstance(config_node.value, s.ConfigPath) and\
```
boehmseb: Why this Check? This could actually be the reason why the test in the CI fails.

LeonievonMann: The problem is that the value in the bb_cfg is a ConfigPath, so when the value is accessed, a prompt asks the user whether the folder should be created (this causes the errors in the tests I think). If the check would not be there, we would get problems because path.isdir and makedirs do not expect a ConfigPath. Should I maybe change the bb_cfg or add the path to the varats_cfg as a str instead of a ConfigPath?

boehmseb: Can't you simply wrap a `str()` around the `config_node.value`?

LeonievonMann: Yes, I actually already tried that, but the result is the same. The prompt is still being printed which leads to the tests failing.

boehmseb (approved):

PR #804, opened 2023-06-27
Patch Provider Infrastructure

LuAbelt:
Infrastructure to add support for applying patches to projects and measure these

PR #813, opened 2023-07-17
Implements MultiPatchReport

vulder:
The multi patch report groups together reports of the same type that where produced with multiple different patches on a project.

depends se-sic/VaRA-Tool-Suite#804


varats-core/varats/report/multi_patch_report.py
```diff
@@ -0,0 +1,82 @@
...
+    @staticmethod
+    def is_baseline_report(file_name: str) -> bool:
+        return file_name.startswith("baseline_")
+
+    @staticmethod
+    def create_patched_report_name(patch: Patch, base_file_name: str) -> str:
+        return (
+            f"patched_{len(patch.shortname)}_" +
+            f"{patch.shortname}_{base_file_name}"
+        )
```
LuAbelt, line 69: Personally I think it's a bit weird to encode the shortname length in the report name, however I can see why it is necessary to extract the patch name if only given a report file name.

A solution to omit the length here might be to just silently convert underscores in the short names to dashes during parsing of the `.info` files, but not sure if that might be confusing for people if their patches are "renamed" (even though if it is just slightly)

vulder: I would not rename them, not only for confusion reasons but also because then a name with a `-` is problematic. Encoding the length is the only options to allow a free name selection. And it should not really do harm, as user will not see the raw file names anyhow, as they are generated by via the API and after parsing the infos are access only through the API.

LuAbelt (approved):

boehmseb (approved):

PR #814, opened 2023-07-17
Implements ReCompile step

vulder:


LuAbelt (approved):
Lgtm

boehmseb:
I'm wondering what we should do with projects that do not implement a `recompile` function.
I guess, at least, a warning or notification to the user would be nice. Currently, you would get a type error of some sort.

vulder:
Well, we can add a run-time check, but in the end throwing an exception because the project does not implement something seems ok to me. In this case, the exception would be that the project does not have a `recompile` method, which is quite clear indicator for "the project does not support/implement recompile"

What would you throw?

boehmseb:
A `NotImplementedError` seems like a reasonable choice. Or we could create a custom exception.

boehmseb (approved):

PR #815, opened 2023-07-17
Reworks output folder steps into own base class

vulder:


vulder:
I tried to get a decorator working but this does not fit/work :( 
Now we have an isinstance check, but at least now type checking works and steps that don't need an output folder are compatible with the ZippedReportSteps

LuAbelt (approved):

PR #816, opened 2023-07-21
Minor improvements to PatchSet

LuAbelt:
Add proper intersection and union operations to the PatchSet. I tinkered around a bit with the intersection being somewhat based on the tags occuring in both PatchSets, but in the end the classic intersection seems like the best option.

We could discuss whether we want to actually override `getitem` and so, what the behavior of it should be. Currently, it behaves like `all_of`.


vulder (approved):
Looks good.

But at some point you should configure `pre-commit` in you setup, to the code gets automatically formatted etc.

PR #821, opened 2023-07-27
Add tig project

LeonievonMann:


varats/varats/projects/c_projects/tig.py
```diff
@@ -0,0 +1,68 @@
...
+    def compile(self) -> None:
+        """Contains instructions on how to build the project."""
+        tig_version_source = local.path(self.source_of_primary)
+
+        clang = bb.compiler.cc(self)
+        with local.cwd(tig_version_source):
+
+            with local.env(CC=str(clang)):
+                bb.watch(make)("-j", get_number_of_jobs(bb_cfg()))
+                bb.watch(make)("install", "j", get_number_of_jobs(bb_cfg()))
```
vulder: This should probably be skipped as we don't want to install the build tig

varats/varats/projects/c_projects/tig.py
```diff
@@ -0,0 +1,68 @@
...
+        pass
+
+    def compile(self) -> None:
+        """Contains instructions on how to build the project."""
+        tig_version_source = local.path(self.source_of_primary)
+
+        clang = bb.compiler.cc(self)
+        with local.cwd(tig_version_source):
+
+            with local.env(CC=str(clang)):
```
vulder:
```suggestion
        cc_compiler = bb.compiler.cc(self)
        with local.cwd(tig_version_source):
            with local.env(CC=str(cc_compiler)):
```
Calling it compiler is "more correct", as theoretically it can also be gcc or something else.

vulder (changes requested):
Except for the two small things and the missing module doc string, this PR LGTM

boehmseb (approved):

vulder (approved):

PR #822, opened 2023-08-04
Add requires attribute to Command

danjujan:
The `requires` attribute is used by `workload_commands` to check if the command includes all necessary arguments.

### DIFF
--- a/tests/experiment/test_workload_util.py
+++ b/tests/experiment/test_workload_util.py
@@ -1,6 +1,7 @@
 """Test VaRA workload utilities."""
 import unittest
 from pathlib import Path
+from unittest.mock import patch
 
 from benchbuild.command import Command, PathToken, RootRenderer
 from benchbuild.source.base import Revision, Variant
@@ -51,6 +52,30 @@ class TestWorkloadCommands(unittest.TestCase):
             project, binary, [wu.WorkloadCategory.EXAMPLE]
         )
         self.assertEqual(len(commands), 1)
+
+    def test_workload_commands_requires(self) -> None:
+        revision = Revision(Xz, Variant(Xz.SOURCE[0], "c5c7ceb08a"))
+        project = Xz(revision=revision)
+        binary = Xz.binaries_for_revision(ShortCommitHash("c5c7ceb08a"))[0]
+
+        commands = next(
+            wu.filter_workload_index(
+                wu.WorkloadSet(wu.WorkloadCategory.EXAMPLE), project.workloads
+            )
+        )
+        commands[0]._requires = {"--compress"}
+        commands = wu.workload_commands(
+            project, binary, [wu.WorkloadCategory.EXAMPLE]
+        )
+        self.assertEqual(len(commands), 0)
+        with patch(
+            "varats.experiment.workload_util.get_extra_config_options",
+            return_value=["--compress"]
+        ):
+            commands = wu.workload_commands(
+                project, binary, [wu.WorkloadCategory.EXAMPLE]
+            )
+            self.assertEqual(len(commands), 1)
 
 
 class TestWorkloadFilenames(unittest.TestCase):
--- a/tests/experiment/test_workload_util.py
+++ b/tests/experiment/test_workload_util.py
@@ -1,6 +1,5 @@
 """Test VaRA workload utilities."""
 import unittest
-from copy import deepcopy
 from pathlib import Path
 
 from benchbuild.command import Command, PathToken, RootRenderer
--- a/tests/experiment/test_workload_util.py
+++ b/tests/experiment/test_workload_util.py
@@ -55,7 +55,7 @@ class TestWorkloadCommands(unittest.TestCase):
 
     def test_workload_commands_requires(self) -> None:
         revision = Revision(Xz, Variant(Xz.SOURCE[0], "c5c7ceb08a"))
-        project = deepcopy(Xz(revision=revision))
+        project = Xz(revision=revision)
         binary = Xz.binaries_for_revision(ShortCommitHash("c5c7ceb08a"))[0]
 
         commands = wu.workload_commands(
--- a/tests/experiment/test_workload_util.py
+++ b/tests/experiment/test_workload_util.py
@@ -2,7 +2,6 @@
 import unittest
 from copy import deepcopy
 from pathlib import Path
-from unittest.mock import patch
 
 from benchbuild.command import Command, PathToken, RootRenderer
 from benchbuild.source.base import Revision, Variant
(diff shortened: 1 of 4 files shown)


varats/varats/projects/c_projects/xz.py
```diff
@@ -87,7 +88,8 @@ class Xz(VProject):
             Command(
                 SourceRoot("xz") / RSBinary("xz"),
                 "-k",
-                "geo-maps/countries-land-1km.geo.json",
+                output_param=["{output}"],
+                output=SourceRoot("geo-maps/countries-land-250m.geo.json"),
```
danjujan, line 94: Benchbuild appends the output params always after the command arguments. This should be sufficient for our use case. No need to implement insertion points right now.

boehmseb: Not sure whether what you are doing here right now is the intended way.
As I understand it, the `output_param` and `output` are intended for the output file the command produces.
The json given here is the input to the command.

But I see your point about where to insert our own arguments.

tests/experiment/test_workload_util.py
```diff
@@ -52,6 +54,30 @@ def test_workload_commands_tags_selected(self) -> None:
...
+        revision = Revision(Xz, Variant(Xz.SOURCE[0], "c5c7ceb08a"))
+        project = deepcopy(Xz(revision=revision))
+        binary = Xz.binaries_for_revision(ShortCommitHash("c5c7ceb08a"))[0]
+
+        commands = next(
+            wu.filter_workload_index(
+                wu.WorkloadSet(wu.WorkloadCategory.EXAMPLE), project.workloads
+            )
+        )
+        commands[0]._requires = {"--compress"}
```
boehmseb: You can implement this change directly in the project.

varats-core/varats/project/project_util.py
```diff
@@ -382,3 +383,28 @@ def copy_renamed_git_to_dest(src_dir: Path, dest_dir: Path) -> None:
         for name in dirs:
             if name == ".gitted":
                 os.rename(os.path.join(root, name), os.path.join(root, ".git"))
+
+
+class Command(_Command):  # type: ignore [misc]
```
boehmseb: I would prefer to give this class its own name to better distinguish it from the base class.
Users typically don't see the import right away and thus, may get confused.

boehmseb (changes requested):
Looks good overall. Only some small suggestions.

varats-core/varats/experiment/workload_util.py
```diff
@@ -92,8 +93,21 @@ def workload_commands(
         )
     ]
 
+    # Filter commands that have required args set.
+    extra_options = set(get_extra_config_options(project))
+
+    def requires_filter(prj_cmd: ProjectCommand) -> bool:
+        if hasattr(prj_cmd.command, "requires") and prj_cmd.command.requires:
+            args = set(prj_cmd.command._args).union(extra_options)
+            return bool(args.intersection(prj_cmd.command.requires))
```
boehmseb: Is this correct? As far as I can see, this only checks if there is an intersection, but not if all required commands are contained. The following should do the trick:
```suggestion
            return prj.command.requires.issubset(args)
```

tests/experiment/test_workload_util.py
```diff
@@ -52,6 +53,20 @@ def test_workload_commands_tags_selected(self) -> None:
         )
         self.assertEqual(len(commands), 1)
 
+    def test_workload_commands_requires(self) -> None:
+        revision = Revision(Xz, Variant(Xz.SOURCE[0], "c5c7ceb08a"))
+        project = deepcopy(Xz(revision=revision))
```
boehmseb: Why the deepcopy here? You are already creating a new instance of `Xz`.

varats-core/varats/project/project_util.py
```diff
@@ -382,3 +383,28 @@ def copy_renamed_git_to_dest(src_dir: Path, dest_dir: Path) -> None:
         for name in dirs:
             if name == ".gitted":
                 os.rename(os.path.join(root, name), os.path.join(root, ".git"))
+
+
+class WrappedCommand(Command):  # type: ignore [misc]
```
boehmseb: Can we find a better name for this? `WrappedCommand` is quite generic and does not tell the user how it is different to a normal `Command`.
We could either go the same route as we did with projects and call it `VCommand` (for "VaRA-TS specific Command") or bring the required argument aspect into the name.

boehmseb (changes requested):
Still needs a little bit of cleanup.
Also, my comment from last time is still open.

varats-core/varats/experiment/workload_util.py
danjujan: That is intended behavior. It must be in the requires set. Otherwise I would have to duplicate commands for my use case: https://github.com/se-sic/VaRA-Tool-Suite/blob/37793ec0aff6def2c867bc69dde1c71be3d31c33/varats/varats/projects/c_projects/xz.py#L144
If someone needs the all required functionality, we can still add a `requires_all` field.

boehmseb: I see, that is something we should definitely consider here.
It would probably be a good idea to have both, a `requires_all` and a `requires_any`.
The check then should only succeed if both are fulfilled.

tests/experiment/test_workload_util.py
danjujan: Good catch. Will remove that.

varats-core/varats/project/project_util.py
danjujan: I prefer `VCommand`. That is more generic and maybe we want to additional functionality in the future.

varats/varats/projects/c_projects/xz.py
danjujan: This is the code in benchbuild that handles this: https://github.com/PolyJIT/benchbuild/blob/master/benchbuild/command.py#L555-L561
As long as `output` is a `PathToken` `output_param` gets appended to the args. While it might not be the intended use case, we can use this to insert our `extra_args` wherever we want. Just put everything that has to come after the extra_args in the `output_param` list. If an element in the list does not contain `{output}` it stays untouched.
Actually, I think the naming on the benchbuild side is a bit off. This functionality can also used for inputs. So why is `output_param` only for `output`? It could be renamed to `end_params` or so and additionally `{input}` placeholders could be replaced too.

tests/experiment/test_workload_util.py
```diff
@@ -1,5 +1,6 @@
 """Test VaRA workload utilities."""
 import unittest
+from copy import deepcopy
```
boehmseb: This import should be unused now, right?

varats/varats/projects/c_projects/xz.py
```diff
@@ -84,26 +85,34 @@ class Xz(VProject):
...
-                "geo-maps/countries-land-250m.geo.json",
+                # Use output_param to ensure input file
+                # gets appended after all arguments.
+                output_param=["{output}"],
+                output=SourceRoot("geo-maps/countries-land-250m.geo.json"),
                 label="countries-land-250m",
-                creates=["geo-maps/countries-land-250m.geo.json.xz"]
+                creates=["geo-maps/countries-land-250m.geo.json.xz"],
+                requires_any={"--compress", "FooBar"},
+                requires_all={"--compress"},
```
boehmseb: Is this both intended? Looks to me a bit like line 114 is a leftover from debugging :thinking:

boehmseb (approved):

vulder (approved):

PR #823, opened 2023-08-18
Fix pip system-wide install failing when using fuse-overlayfs

danjujan:


danjujan:
Close in favor of https://github.com/se-sic/VaRA-Tool-Suite/pull/824

PR #824, opened 2023-08-21
Use python virtualenv inside containers.

boehmseb:
This avoids the long-standing issue of not being able to update pip inside containers.


danjujan (approved):
Works

vulder (approved):

PR #826, opened 2023-08-23
Handle multiple binaries for same RevisionRange in RevisionBinaryMap

LuAbelt:
Fixes https://github.com/se-sic/VaRA/issues/1033

### DIFF
--- a/varats-core/varats/utils/git_util.py
+++ b/varats-core/varats/utils/git_util.py
@@ -3,6 +3,7 @@ import abc
 import logging
 import re
 import typing as tp
+from collections import defaultdict
 from enum import Enum
 from itertools import chain
 from pathlib import Path
@@ -1031,7 +1032,8 @@ class RevisionBinaryMap(tp.Container[str]):
     def __init__(self, repo_location: Path) -> None:
         self.__repo_location = repo_location
         self.__revision_specific_mappings: tp.Dict[
-            'AbstractRevisionRange', tp.List[ProjectBinaryWrapper]] = {}
+            'AbstractRevisionRange',
+            tp.List[ProjectBinaryWrapper]] = defaultdict(list)
         self.__always_valid_mappings: tp.List[ProjectBinaryWrapper] = []
 
     def specify_binary(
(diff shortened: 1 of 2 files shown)


boehmseb (approved):

varats-core/varats/utils/git_util.py
```diff
@@ -1069,7 +1069,14 @@ def specify_binary(
...
         if validity_range:
-            self.__revision_specific_mappings[validity_range] = wrapped_binary
+            if validity_range in self.__revision_specific_mappings:
+                self.__revision_specific_mappings[validity_range].append(
+                    wrapped_binary
+                )
+            else:
+                self.__revision_specific_mappings[validity_range] = [
+                    wrapped_binary
+                ]
```
vulder: Instead of explicitly checking, just use a DefaultDict and always append.

varats-core/varats/utils/git_util.py
```diff
@@ -1030,8 +1030,8 @@ class RevisionBinaryMap(tp.Container[str]):
 
     def __init__(self, repo_location: Path) -> None:
         self.__repo_location = repo_location
-        self.__revision_specific_mappings: tp.Dict['AbstractRevisionRange',
-                                                   ProjectBinaryWrapper] = {}
+        self.__revision_specific_mappings: tp.Dict[
+            'AbstractRevisionRange', tp.List[ProjectBinaryWrapper]] = {}
```
vulder: It works with fine with list, but maybe a set would work better so wrapper cannot be inserted twice?
However, a `set` would have the drawback of having no guaranteed iteration order, so we would need to convert it to a list and sort bevor returning it.

What's your opinion on that topic?

vulder (changes requested):
Two small things, we should discuss, otherwise this looks good :+1:

varats-core/varats/utils/git_util.py
LuAbelt: I also thought about a `set` first, but I choose `list` first, to keep it consistent with `__always_valid_mappings`. So if we change to `set` here I would suggest to do it for the always valid mappings as well.

Regarding guaranteed iteration order: Is this something we actually need/use at the moment?
From what I see all experiments use all binaries anyways and create separate result files for each binary.

But since we return the binaries as a `list` anyways in `__getitem__`, I also wouldn't mind just returning a sorted version of that list

vulder: Ok, then lest sort the list but keep a `list` for now. By that we reduce a bit of randomness and if we later find problems with with duplicates we can switch to a set afterwards.

vulder (approved):

PR #835, opened 2023-09-12
Fix Patch Provider pulling in parallel

LuAbelt:
Adds some safeguards to avoid pulling concurrently when running experiments on the cluster

Fixes https://github.com/se-sic/VaRA/issues/1038

### DIFF
--- a/varats-core/varats/provider/patch/patch_provider.py
+++ b/varats-core/varats/provider/patch/patch_provider.py
@@ -6,12 +6,9 @@ applied during an experiment to alter the state of the project.
 """
 
 import os
-import random
-import time
 import typing as tp
 import warnings
 from pathlib import Path
-from time import sleep
 
 import benchbuild as bb
 import yaml
--- a/varats-core/varats/provider/patch/patch_provider.py
+++ b/varats-core/varats/provider/patch/patch_provider.py
@@ -17,11 +17,11 @@ import benchbuild as bb
 import yaml
 from benchbuild.project import Project
 from benchbuild.source.base import target_prefix
-from utils.filesystem_util import lock_file
 from yaml import YAMLError
 
 from varats.project.project_util import get_local_project_git_path
 from varats.provider.provider import Provider, ProviderType
+from varats.utils.filesystem_util import lock_file
 from varats.utils.git_commands import pull_current_branch, fetch_repository
 from varats.utils.git_util import (
     CommitHash,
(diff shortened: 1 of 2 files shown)


varats-core/varats/utils/filesystem_util.py
```diff
@@ -13,3 +15,18 @@ def __init__(self, folder: tp.Union[Path, str]) -> None:
...
             "but was already present."
         )
+
+
+@contextmanager
+def lock_file(lock_path: Path, lock_mode: int = fcntl.LOCK_EX) -> tp.Generator:
+    # Ensure that the lock exists
+    lock_path.touch(exist_ok=True)
+
+    open_mode = os.O_RDWR | os.O_CREAT | os.O_TRUNC
```
boehmseb: Shouldn't `os.O_CREAT` already ensure that the lock file exists?

varats-core/varats/utils/filesystem_util.py
```diff
@@ -13,3 +15,18 @@ def __init__(self, folder: tp.Union[Path, str]) -> None:
             f"Folder: '{str(folder)}' should be created "
             "but was already present."
         )
+
+
+@contextmanager
+def lock_file(lock_path: Path, lock_mode: int = fcntl.LOCK_EX) -> tp.Generator:
```
boehmseb, line 21: Interesting that mypy does not complain about the missing generic parameters to `tp.Generator` :thinking:

varats-core/varats/utils/filesystem_util.py
LuAbelt: True, I will remove the `touch`

varats-core/varats/provider/patch/patch_provider.py
```diff
@@ -237,12 +238,10 @@ class PatchProvider(Provider):
...
 
-        # BB only performs a fetch so our repo might be out of date
-        pull_current_branch(self._get_patches_repository_path())
+        self._update_local_patches_repo()
+        repo_path = self._get_patches_repository_path()
 
-        patches_project_dir = Path(
-            self._get_patches_repository_path() / self.project.NAME
-        )
+        patches_project_dir = Path(repo_path / self.project.NAME)
```
vulder:
```suggestion
        patches_project_dir = repo_path / self.project.NAME
```
repo_path is already a path, so if NAME is a string you can directly add it to the path and get a path

vulder (approved):
Otherwise, following the docs, this looks fine.

vulder (approved):

boehmseb (approved):

PR #839, opened 2023-09-24
Adds functionality for patch-based configuration

boehmseb:
Compile-time configurable projects can now be configured using patches.
Therefore, patches can specify a list of `feature_tags`.

Case studies can contain a configuration map with `PatchConfiguration`s that contains a mapping from config id to a list of feature tags associated with this configuration.

Configuration patches are selected by matching their feature tags to the feature tags of the given configuration.

Since there are multiple types of configuration maps now, their type must be specified using the key `config_type` in case study files.

### DIFF
--- a/docs/source/tutorials/configuration_specific_experiments.rst
+++ b/docs/source/tutorials/configuration_specific_experiments.rst
@@ -21,6 +21,7 @@ One just needs to extend the case-study file of a project with a yaml document t
 .. code-block:: yaml
 
   ---
+  config_type: PlainCommandlineConfiguration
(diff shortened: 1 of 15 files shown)


vulder (approved):
Looks good, except for the mypy error. I'm not sure how to type this correctly.

boehmseb:
> Looks good, except for the mypy error. I'm not sure how to type this correctly.

Thing is, you cannot type-hint lambdas. One solution would be to make it a function instead (which is a bit overkill imo).
I don't know why mypy triggers here in the first place. Other locations where we use `map` with a lambda are fine.

PR #844, opened 2023-10-07
Adds token for runtime config params for workload commands.

boehmseb:


vulder (approved):
works and LGTM

we need the BB release + update, afterwards we can merge this

PR #862, opened 2023-10-24
Adds interaction filter for selecting individual commits

boehmseb:


vulder (approved):
This part LGTM. However, do you also have implemented support on the VaRA side for this or do you only plan to use it on the TS side?

boehmseb:
I've implemented this some time ago on both sides and need it now for my experiments. Since here, this was already its own branch, I created the PR. On the VaRA side, the support for this filter is implemented on the branch where my analysis is implemented.

PR #870, opened 2024-01-05
Adds support for trogon.

boehmseb:
Most vara-* commands now support `tui` as a subcommand which opens a text-based UI for selecting command line parameters.

Some limitations apply:
- `vara-run` is not supported because arguments do not play nice with command groups in click
- `vara-plot` and `vara-table` require support for `click.MultiCommand` (Textualize/trogon#71)
- `trogon` contains some more small bugs that limit usability. I opened PRs for them, but until they are merged, you can use the following branch for testing: https://github.com/boehmseb/trogon/tree/trogonFixes

### DIFF
--- a/varats/varats/tools/driver_container.py
+++ b/varats/varats/tools/driver_container.py
@@ -27,7 +27,7 @@ from varats.utils.settings import vara_cfg, save_config, bb_cfg
 LOG = logging.Logger(__name__)
 
 
-@tui()
+@tui()  # type: ignore
 @click.group(
     help="Manage base container images.",
     context_settings={"help_option_names": ['-h', '--help']}
@@ -38,7 +38,9 @@ def main() -> None:
     bb_cfg()
 
 
-@main.command(help="Build base containers for the current research tool.")
+@main.command(
+    help="Build base containers for the current research tool."
+)  # type: ignore
 @click.option(
     "--debug", is_flag=True, help="Debug failed image builds interactively."
 )
@@ -103,7 +105,9 @@ def build(
             export_base_images()
 
 
-@main.command(help="Delete base containers for the current research tool.")
+@main.command(
+    help="Delete base containers for the current research tool."
+)  # type: ignore
 @click.option(
     "--build-type",
     type=EnumChoice(BuildType, case_sensitive=False),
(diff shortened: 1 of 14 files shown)


varats/varats/tools/driver_container.py
```diff
@@ -144,7 +146,8 @@ def delete(
     "-t",
     "--tool",
     type=click.Choice([*get_supported_research_tool_names(), "none"]),
-    default=lambda: vara_cfg()["container"]["research_tool"].value or "none",
+    required=True,
+    # default=lambda: vara_cfg()["container"]["research_tool"].value or "none",
```
LuAbelt, line 156: What is this change for?

boehmseb: Trogon currently cannot evaluate functions as default parameters. Since we can determine a good default only dynamically, I made the option required instead and removed the default.

varats/varats/tools/driver_paper_config.py
```diff
@@ -32,85 +34,56 @@ def main() -> None:
...
-                vara_cfg()["paper_config"], "folder",
-                str(get_varats_base_folder())
-            )
-        )
-        if not (pc_folder_path.exists() and pc_folder_path.is_dir()):
-            LOG.error(
-                f"Paper config folder not set: {pc_folder_path} "
-                "(Path does not exist or is no directory)."
-            )
-            return
```
vulder, line 102: Do we no longer need the checks here?

vulder (commented):
Do I understand this correctly, this is "opt-in" as in one can use the CLI as is, without changes and if wanted with the clicky UI?

PS: also some tests seem not to be happy :D

varats/varats/tools/driver_paper_config.py
boehmseb: No, because by now we take care of creating these directories whenever we save the config.

boehmseb: And the config folder has a sane default value set.

boehmseb:
> Do I understand this correctly, this is "opt-in" as in one can use the CLI as is, without changes and if wanted with the clicky UI?

Yes, this basically adds a new subcommand `tui` to each supported tool that starts the text-based UI (bonus points for a good idea where/how to document this).
 
> PS: also some tests seem not to be happy :D

I have no idea why. That test initially also failed on my machine but for a different reason. All tests pass when I execute them locally.

LuAbelt (approved):

PR #873, opened 2024-01-30
Reworks the table to group together similar categories

vulder:


varats/varats/tables/feature_perf_precision.py
```diff
@@ -35,6 +35,33 @@
 from varats.table.tables import TableFormat, TableGenerator
 from varats.utils.git_util import calc_repo_loc, ChurnConfig
 
+GROUP_SYNTHETIC_CATEGORIES = True
```
LuAbelt, line 38: I am unsure whether it might be better to have this as a command line option for the specific tables.

On the one hand it would be a bit cleaner (from my perspective) from a design POV to have it as a CLI option, but after all most of the handling for the grouping is hard-coded anyway specific to our naming patterns for the case studies, so it won't be generally applicable anyways.

vulder: The idea sounds reasonable, maybe let's keep for now the global and change it later on into a CLI option, reworking a bit the grouping interface so that others could reuse it. I'm thinking in the direction of requiring a callback from the user that "defines" the grouping (does not even have to be str -> str but could be any -> any, as long as the users types match).

But this kind of rework would probably fit better into it's own PR. If agreed, I'll create an issue and maybe even implement it in the next weeks when I find time.

LuAbelt (approved):

PR #874, opened 2024-01-30
Implements generator to automatically generate a performance precision plot per case study

vulder:


varats/varats/plots/feature_perf_precision.py
```diff
@@ -153,12 +154,40 @@ def calc_missing_revisions(
 
 
 class PerfProfDistPlotGenerator(
-    PlotGenerator, generator_name="fperf-precision-dist", options=[]
+    PlotGenerator,
+    generator_name="fperf-precision-dist",
+    options=[REQUIRE_MULTI_CASE_STUDY]
 ):
-    """Generates performance distribution plot."""
+    """Generates performance distribution plot for a given amount of case
```
boehmseb:
```suggestion
    """Generates performance distribution plot for a given list of case
```

boehmseb (approved):

PR #876, opened 2024-02-02
Add experiment CompareASTBlame

LeonievonMann:
Needs se-sic/VaRA#1054

### DIFF
--- a/varats/varats/data/reports/blame_annotations.py
+++ b/varats/varats/data/reports/blame_annotations.py
@@ -36,12 +36,27 @@ class BlameInstruction():
         return self.__varahash
 
 
+class BlameFunctionAnnotations():
+    """Contains all instruction annotations for one function."""
+
+    def __init__(self) -> None:
+        self.__blame_annotations = {}
+
+    def add_annotation(self, name: str, inst: BlameInstruction) -> None:
+        self.__blame_annotations[name] = inst
+
+    @property
+    def blame_annotations(self) -> tp.Dict[str, BlameInstruction]:
+        """Iterate over all blame annotations."""
+        return self.__blame_annotations
+
+
 class BlameAnnotations(BaseReport, shorthand="BA", file_type="yaml"):
     """Report containing debug blame and blame annotations."""
 
     def __init__(self, path: Path) -> None:
         super().__init__(path)
-        self.__blame_annotations = {}
+        self.__functions = {}
 
         with open(path, 'r') as stream:
             documents = yaml.load_all(stream, Loader=yaml.CLoader)
@@ -50,18 +65,22 @@ class BlameAnnotations(BaseReport, shorthand="BA", file_type="yaml"):
 
             raw_blame_report = next(documents)
 
-            for raw_entry in raw_blame_report['annotations']:
-                new_entry = (
-                    BlameInstruction.create_blame_instruction(
-                        raw_blame_report['annotations'][raw_entry]
+            for func in raw_blame_report['functions']:
+                new_func = BlameFunctionAnnotations()
+                func_entry = raw_blame_report['functions'][func]
+                for raw_entry in func_entry['annotations']:
+                    new_entry = (
+                        BlameInstruction.create_blame_instruction(
+                            func_entry['annotations'][raw_entry]
+                        )
                     )
-                )
-                self.__blame_annotations[raw_entry] = new_entry
+                    new_func.add_annotation(raw_entry, new_entry)
+                self.__functions[func] = new_func
 
     @property
-    def blame_annotations(self) -> tp.ValuesView[BlameInstruction]:
+    def functions(self) -> tp.ValuesView[BlameFunctionAnnotations]:
         """Iterate over all blame annotations."""
-        return self.__blame_annotations.values()
+        return self.__functions.values()
 
 
 class ASTBlameReport(BaseReport, shorthand="BAST", file_type="yaml"):
@@ -133,16 +152,20 @@ def compare_blame_annotations(
     based to the AST based blame."""
     ast_report = ASTBlameReport(path)
 
-    for entry in ast_ba.blame_annotations:
-        if entry.dbghash and entry.varahash:
-            ast_report.update_dbg_ast(entry.dbghash != entry.varahash)
+    for func in ast_ba.functions:
+        for entry in func.blame_annotations.values():
+            if entry.dbghash and entry.varahash:
+                ast_report.update_dbg_ast(entry.dbghash != entry.varahash)
 
-    for line_entry, ast_entry in zip(
-        line_ba.blame_annotations, ast_ba.blame_annotations
-    ):
-        if line_entry.varahash and ast_entry.varahash:
-            ast_report.update_line_ast(
-                line_entry.varahash != ast_entry.varahash
-            )
+    for line_func, ast_func in zip(line_ba.functions, ast_ba.functions):
+        line_annotations = line_func.blame_annotations
+        ast_annotations = ast_func.blame_annotations
+        for inst in ast_annotations:
+            if line_annotations[inst].varahash and ast_annotations[inst
+                                                                  ].varahash:
+                ast_report.update_line_ast(
+                    line_annotations[inst].varahash !=
+                    ast_annotations[inst].varahash
+                )
 
     return ast_report
--- a/varats/varats/data/reports/blame_annotations.py
+++ b/varats/varats/data/reports/blame_annotations.py
@@ -9,7 +9,7 @@ from varats.report.report import BaseReport
 
 
 class BlameInstruction():
-    """Collection of debug blame and VaRA blame."""
+    """Debug-based and VaRA-based blame annotations for a single instruction."""
 
     def __init__(
         self, inst: str, dbg_hash: str, vara_computed_hash: str
@@ -22,8 +22,8 @@ class BlameInstruction():
     def create_blame_instruction(
         raw_entry: tp.Dict[str, tp.Any]
     ) -> 'BlameInstruction':
-        """Creates a :class`BlameInstrucion`from the corresponding yaml document
-        section."""
+        """Creates a :class:`BlameInstrucion` from the corresponding yaml
+        document section."""
         inst = str(raw_entry['inst'])
         dbg_hash = str(raw_entry['dbghash'])
         vara_computed_hash = str(raw_entry['varahash'])
@@ -31,7 +31,7 @@ class BlameInstruction():
 
     @property
     def inst(self) -> str:
-        """Intermediate representation of LLVM instruction."""
+        """Textual representation of LLVM-IR instruction."""
         return self.__inst
 
     @property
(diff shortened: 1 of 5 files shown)


varats/varats/data/reports/blame_annotations.py
```diff
@@ -0,0 +1,171 @@
...
+from varats.base.version_header import VersionHeader
+from varats.report.report import BaseReport
+
+
+class BlameInstruction():
+    """Collection of debug blame and VaRA blame."""
+
+    def __init__(self, dbghash: str, varahash: str) -> None:
+        self.__dbghash = dbghash
+        self.__varahash = varahash
```
vulder: Are you reusing another format for the file? varahash seems to be the "current" thing and "dbg" will refer to the ast-ones? I'm only a bit confused why you choose those names.

vulder (commented):
Overall setup looks good. Maybe the file format could be improved to be easier read/parse able but overall the code should work.

varats/varats/data/reports/blame_annotations.py
LeonievonMann: The dbghash is the one calculated with the debug information about the location and is line based. And varahash is taken from the blame annotation made by VaRA. Depending on which flag is used to create the .ll file on which the report is based on, this is either line based or AST based.

varats/varats/experiments/vara/blame_ast_experiment.py
```diff
@@ -0,0 +1,243 @@
...
+        )
+        # Generate blame annotation report
+        analysis_actions.append(
+            BlameAnnotationGeneration(
+                project, self.get_handle(), "linereport-"
+            )
+        )
+
+        # Compile with AST based blame annotations
+        project.cflags += ["-fvara-ast-GB"]
```
boehmseb, line 224: Does this work like intended? I'm not sure, but I think this would also modify the cflags for actions you added before.

LeonievonMann: This would be a problem. How else could I do this?

varats/varats/data/reports/blame_annotations.py
vulder: Ah, I see. Then it's good, maybe call it `__dbg_hash` and `__vara_computed_hash` so it's a bit more distinct (and the word hash is separated in snake_case), otherwise I'm the naming then makes sense. Maybe add a small comment for the getter/variable that the vara_computed_hash can be produced in different ways, so that readers can be aware.

varats/varats/experiments/vara/blame_ast_experiment.py
vulder: I don't think this works, as the c_flags are all collected and afterwards put into the generated compiler wrapper.
I also do not see a "simple" other way to do this in one go, as for one compile run of the project we can only have one type of commit meta data (with or without ast-blame).
The best and probably cleanest option would be to have 2 experiments, 1 producing the ast-blame based data, the other one the normal/baseline data. Afterwards, in your report you could then load both files and join them together, so you can represent the data uniformly to the users (plots/tables etc.).

varats/varats/data/reports/blame_annotations.py
```diff
@@ -0,0 +1,184 @@
...
+    ) -> None:
+        self.__inst = inst
+        self.__dbg_hash = dbg_hash
+        self.__vara_computed_hash = vara_computed_hash
+
+    @staticmethod
+    def create_blame_instruction(
+        raw_entry: tp.Dict[str, tp.Any]
+    ) -> 'BlameInstruction':
+        """Creates a :class`BlameInstrucion`from the corresponding yaml document
```
boehmseb:
```suggestion
        """Creates a :class:`BlameInstrucion` from the corresponding yaml document
```

varats/varats/data/reports/blame_annotations.py
```diff
@@ -0,0 +1,184 @@
...
+        section."""
+        inst = str(raw_entry['inst'])
+        dbg_hash = str(raw_entry['dbghash'])
+        vara_computed_hash = str(raw_entry['varahash'])
+        return BlameInstruction(inst, dbg_hash, vara_computed_hash)
+
+    @property
+    def inst(self) -> str:
+        """Intermediate representation of LLVM instruction."""
```
boehmseb:
```suggestion
        """Textual representation of LLVM-IR instruction."""
```

varats/varats/data/reports/blame_annotations.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @property
+    def dbg_hash(self) -> str:
+        """Blame based on debug information."""
+        return self.__dbg_hash
+
+    @property
+    def vara_computed_hash(self) -> str:
+        """
+        Blame based on IRegion.
```
boehmseb:
```suggestion
        Blame based on VaRA's BlameRegion.
```

varats/varats/data/reports/blame_annotations.py
```diff
@@ -0,0 +1,184 @@
...
+
+        Can be produced in different ways, based on which flag is used.
```
boehmseb:
```suggestion
        Either line-based or AST-based depending on whether `-fvara-ast-GB` is used or not.
```

varats/varats/data/reports/blame_annotations.py
```diff
@@ -0,0 +1,184 @@
...
+        self.__blame_annotations[name] = inst
+
+    @property
+    def blame_annotations(self) -> tp.Dict[str, BlameInstruction]:
+        """Iterate over all blame annotations."""
+        return self.__blame_annotations
+
+
+class BlameAnnotations(BaseReport, shorthand="BA", file_type="yaml"):
+    """Report containing debug blame and blame annotations."""
```
boehmseb:
```suggestion
    """Report containing debug-based and VaRA-based blame annotations."""
```

varats/varats/data/reports/blame_annotations.py
```diff
@@ -0,0 +1,184 @@
...
+    @property
+    def eq_line_ast(self) -> int:
+        """Count of equal instructions between line and ast blame."""
+        return self.__eq_line_ast
+
+
+def compare_blame_annotations(
+    line_ba: BlameAnnotations, ast_ba: BlameAnnotations, path: Path
+) -> ASTBlameReport:
+    """Compares the debug based to the AST based annotations as well as the line
```
boehmseb:
```suggestion
    """Compares the debug-based to the AST-based blame annotations as well as the line
```

varats/varats/data/reports/blame_annotations.py
```diff
@@ -0,0 +1,184 @@
...
+from pathlib import Path
+
+import yaml
+
+from varats.base.version_header import VersionHeader
+from varats.report.report import BaseReport
+
+
+class BlameInstruction():
+    """Collection of debug blame and VaRA blame."""
```
boehmseb:
```suggestion
    """Debug-based and VaRA-based blame annotations for a single instruction."""
```

varats/varats/data/reports/blame_annotations.py
```diff
@@ -0,0 +1,184 @@
...
+                    new_func.add_annotation(new_entry.inst, new_entry)
+                self.__functions[func] = new_func
+
+    @property
+    def functions(self) -> tp.ValuesView[BlameFunctionAnnotations]:
+        """Iterate over all blame annotations."""
+        return self.__functions.values()
+
+
+class ASTBlameReport(BaseReport, shorthand="BAST", file_type="yaml"):
```
boehmseb: The report explicitly deals with the comparison of different blame reports or a quality-check of the AST-based blame report. Maybe we can find a more fitting name.

varats/varats/experiments/vara/blame_ast_experiment.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def analyze(self) -> actions.StepResult:
+        """
+        This step performs the actual analysis with the correct command line
+        flags.
+
+        Flags used:
+            * -vara-BA: to run a commit flow report
+            * -yaml-report-outfile=<path>: specify the path to store the results
+        """
```
boehmseb: This comment seems very copy-pasted. Pleas reformulate to better fit this step.

varats/varats/experiments/vara/blame_ast_experiment.py
```diff
@@ -0,0 +1,280 @@
...
+
+        for binary in self.project.binaries:
+            # Add to the user-defined path for saving the results of the
+            # analysis also the name and the unique id of the project of every
+            # run.
```
boehmseb:
```suggestion
```

varats/varats/experiments/vara/blame_ast_experiment.py
```diff
@@ -0,0 +1,280 @@
...
+        Returns the specified steps to run the project(s) specified in the call
+        in a fixed order.
+
+        Args:
+            project: to analyze
+        """
+        # Try, to build the project without optimizations to get more precise
+        # blame annotations. Note: this does not guarantee that a project is
+        # build without optimizations because the used build tool/script can
+        # still add optimizations flags after the experiment specified cflags.
```
boehmseb:
```suggestion
```

varats/varats/experiments/vara/blame_ast_experiment.py
```diff
@@ -0,0 +1,280 @@
...
```
boehmseb:
```suggestion
```

varats/varats/experiments/vara/blame_ast_experiment.py
```diff
@@ -0,0 +1,280 @@
...
+        # Compile with AST based blame annotations
+        project.cflags += ["-fvara-ast-GB"]
+        analysis_actions = _create_default_bc_file_creation_actions(
+            project,
+            bc_file_extensions if bc_file_extensions else [],
+            extraction_error_handler=create_default_compiler_error_handler(
+                self.get_handle(), project, self.REPORT_SPEC.main_report
+            )
+        )
+        # Generate blame annotation report
```
boehmseb:
```suggestion
```

varats/varats/experiments/vara/blame_ast_experiment.py
```diff
@@ -0,0 +1,280 @@
...
+        BE.setup_basic_blame_experiment(self, project, BA)
+        # Compile with line based blame annotations
+        analysis_actions = _create_default_bc_file_creation_actions(
+            project,
+            bc_file_extensions if bc_file_extensions else [],
+            extraction_error_handler=create_default_compiler_error_handler(
+                self.get_handle(), project, self.REPORT_SPEC.main_report
+            )
+        )
+        # Generate blame annotation report
```
boehmseb:
```suggestion
```

varats/varats/experiments/vara/blame_ast_experiment.py
```diff
@@ -0,0 +1,280 @@
...
+    def actions_for_project(
+        self, project: VProject
+    ) -> tp.MutableSequence[actions.Step]:
+        """
+        Returns the specified steps to run the project(s) specified in the call
+        in a fixed order.
+
+        Args:
+            project: to analyze
+        """
```
boehmseb: This comment seems very copy-pasted. Pleas reformulate to better fit this experiment.

boehmseb (changes requested):

boehmseb (approved):

PR #880, opened 2024-03-27
Add the possibility to view the status of case-studies in the gui

Sinerum:
Add a casestudie filter aswell as the ability to view the status of a case study when browsing revisions in the cast study generation gui

### DIFF
--- a/varats/varats/gui/cs_gen/case_study_generation.py
+++ b/varats/varats/gui/cs_gen/case_study_generation.py
@@ -7,6 +7,8 @@ from pathlib import Path
 
 import benchbuild as bb
 import pygit2
+from benchbuild import Experiment
+from benchbuild.experiment import ExperimentRegistry
 from PyQt5.QtCore import (
     QModelIndex,
     QDateTime,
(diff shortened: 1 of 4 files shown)


varats/varats/gui/cs_gen/case_study_generation.py
```diff
@@ -246,22 +277,39 @@ def show_revision_data(self, index: QModelIndex) -> None:
...
+        self.cs_filter = cs_filter
+        self.invalidate()
+
     def filterAcceptsRow(
         self, source_row: int, source_parent: QModelIndex
     ) -> bool:
         commit_index = self.sourceModel().index(source_row, 0, source_parent)
         author_index = self.sourceModel().index(source_row, 1, source_parent)
-        return self.sourceModel().data(commit_index,
+        return ((not self.cs_filter) or FullCommitHash(self.sourceModel().data(commit_index,Qt.WhatsThisRole).hex) in self._case_study.revisions) and (self.sourceModel().data(commit_index,
```
boehmseb: Please refactor out some of the expressions in this condition. It is way to long and therefore hard to read.

boehmseb (changes requested):
Newer versions of PyGit2 don't support Commit.hex anymore.
This would be a good chance to fix this for the cs-gui.

boehmseb (approved):

PR #882, opened 2024-03-28
Adds lock to FM provider

LuAbelt:
Adds a lock to the feature model provider.

When running experiments on slurm, it can happen that multiple nodes fetch the feature model repository at the same time. As nodes may use a shared feature model repository (E.g. on `/scratch/`) this can lead to issues.

This is no issue if the feature model repository is already up-to-date, however if there are any upstream changes this leads to errors in the fetch on some nodes and subsequent failure of the slurm jobs.

### DIFF
--- a/varats-core/varats/provider/feature/feature_model_provider.py
+++ b/varats-core/varats/provider/feature/feature_model_provider.py
@@ -5,9 +5,9 @@ from pathlib import Path
 import benchbuild as bb
 from benchbuild.project import Project
 from benchbuild.source.base import target_prefix
-from utils.filesystem_util import lock_file
 
 from varats.provider.provider import Provider
+from varats.utils.filesystem_util import lock_file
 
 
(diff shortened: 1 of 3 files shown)


varats-core/varats/provider/feature/feature_model_provider.py
```diff
@@ -90,6 +91,9 @@ def _get_feature_model_repository_path() -> Path:
             refspec="origin/HEAD",
             limit=1,
         )
-        fm_source.fetch()
+        lock_path = Path(target_prefix()) / "fm_provider.lock"
+
```
vulder:
```suggestion

        lock_path = Path(target_prefix()) / "fm_provider.lock"
```

vulder (changes requested):

boehmseb (approved):

vulder (approved):

PR #884, opened 2024-04-27
Implements experiment to detect the hot functions of a project

vulder:


varats/varats/projects/c_projects/xz.py
```diff
@@ -91,7 +91,7 @@ class Xz(VProject):
                 # Use output_param to ensure input file
                 # gets appended after all arguments.
                 output_param=["{output}"],
-                output=SourceRoot("geo-maps/countries-land-250m.geo.json"),
+                output=SourceRoot("geo-maps/countries-land-1km.geo.json"),
```
boehmseb: Should this change be in this PR?

varats-core/varats/report/hot_functions_report.py
```diff
@@ -0,0 +1,98 @@
...
+        return [
+            XRayFunctionWrapper(
+                name=row["function"], count=row['count'], sum_time=row["sum"]
+            ) for _, row in self.__function_data.head(limit).iterrows()
+        ]
+
+    def hot_functions(self, threshold=2) -> tp.List[XRayFunctionWrapper]:
+        """
+        Args:
+            threshold: min percentage a function needs as self time to count as hot
```
boehmseb: This currently does not use self time, but time including children.
This should be reflected in the function's documentation.

varats-core/varats/report/hot_functions_report.py
```diff
@@ -0,0 +1,98 @@
...
+                "Threshold value needs to be in the range [0,...,100] "
+                f"but was {threshold}"
+            )
+
+        self.__function_data.sort_values(
+            by='sum', ascending=False, inplace=True
+        )
+        # The total time tracked only includes time spend in the top n
+        # (MAX_TRACK_FUNCTIONS) functions
+        total_time_tracked = self.__function_data["sum"].sum()
```
boehmseb: This currently calculates the sum of all measured functions' execution times including their children and, thus, gives a number much larger than the total execution time.
To get the correct total time, you should use `max()` instead.

```suggestion
        total_time_tracked = self.__function_data["sum"].max()
```

varats-core/varats/report/hot_functions_report.py
```diff
@@ -0,0 +1,98 @@
...
+        for wl_name in self.workload_names():
+            for report in self.reports(wl_name):
+                report.print_full_dump()
+
+    def hot_functions_per_workload(
+        self, threshold=2
+    ) -> tp.Dict[str, tp.List[XRayFunctionWrapper]]:
+        """
+        Args:
+            threshold: min percentage a function needs as self time to count as hot
```
boehmseb: Same as above.

boehmseb (changes requested):
The `HotFunctionsReport` still needs a little bit of cleanup and some typing stuff needs fixing.

varats/varats/projects/c_projects/xz.py
vulder: It's a quick fix on the side, If you want I can create a separate PR for this but the change itself is correct and fixes a bug.

boehmseb: It would be the cleaner solution.

vulder: If you insist -.-

boehmseb (approved):

PR #890, opened 2024-08-28
Improve handling of git operations

boehmseb:
Our handling of git operations in the tool-suite is a mess; some functions use pygit2, others use command-line git and all have different interfaces and conventions and I am very annoyed that I always have the incorrect parameters at hand when I need to use some git functionality.

This PR aims to improve this situation by introducing a lightweight `RepositoryHandle` class that allows easy and uniform access to both, pygit2 and command-line git`. Since git is ubiquitous in the tool-suite, there are a lot of changes, but most of them are very minor.

As a result of the refactoring, `git_util` no longer depends on `project_util`, but the other way round.
In my opinion this makes more sense since projects make use of git, while git can happily be used without projects.

One open point that is remains is to sort out what belongs in `git_util` vs. `git_commands` since that is currently not very strict and at least one function in `git_commands` uses a function in `git_util`, which imo it should not do.

Also, this PR currently breaks pre-commit because the parser library yapf uses does not fully comply with python 3.
Specifically, I use a starred expression in a location where it is not happy about and just crashes.
Not sure how to fix this, though.


varats-core/varats/utils/git_util.py
```diff
@@ -154,64 +208,50 @@ def __get_git_path_arg(repo_folder: tp.Optional[Path] = None) -> tp.List[str]:
     return ["-C", f"{repo_folder}"]
```
LuAbelt: Is the `__get_git_path_arg` still required after this change? 
It seems to me that wherever it was used, we now use `RepositoryHandle`

LuAbelt (commented):
Looks good; Thanks for the effort!

Regarding `git_util` and `git_commands`:
I don't really have a strong opinion on this - but if we want to establish/enforce a distinction then in my opinion, everything that effectively boils down to a simple call to `git` without any complex further processing/filtering would probably be better placed in `git_commands`.
Which effectively moves most of what is currently in `util` over to `commands`. Also for some functions it is probably debatable, whether the code around the `git` call is now considered complex or not

But maybe someone else can come up with a better distinction

vulder:
From what I see, the * not only breaks yapf but also doc building.

git_util vs git_commands
If I remember correctly, `git_commands` should only contain wrappers for the actual commands and nothing more, everything else should be done in `git_util`. @Sinerum did the initial refactor/split, he should know best what the distinction was.

Overall, the idea sounds good, especially as passing the context seems to make a lot of code easier. However, I still need to do a full review look. Will try to do this quickly in the comming days.

varats-core/varats/utils/git_util.py
boehmseb: That's right, I just forgot to remove that function.

boehmseb:
Fortunately, plumbum correctly unpacks the list before executing the command, so I could work around the * issue by removing it.

boehmseb:
Regarding the `git_commands` vs. `git_utils`, my intent was to suggest doing this in a separate PR at some point.
This one is big enough as it is.

LuAbelt (approved):

Sinerum:
Regarding the issue with Yapf: The yapf mirror we use, has been deprecated for a while now and the[ main yapf repo](https://github.com/google/yapf)  now supports pre-commit, and already has this parsing issue fixed.

Sinerum (approved):

boehmseb:
> Regarding the issue with Yapf: The yapf mirror we use, has been deprecated for a while now and the[ main yapf repo](https://github.com/google/yapf) now supports pre-commit, and already has this parsing issue fixed.

Good to know, we should update that at some point.

varats/varats/plots/paper_config_overview.py
```diff
@@ -48,7 +48,7 @@ def _load_projects_ordered_by_year(
             case_study, experiment_type
         )
 
-        repo = get_local_project_git(case_study.project_name)
+        repo = get_local_project_repo(case_study.project_name).pygit_repo
```
vulder: maybe call `repo` here `pygit_repo` to make a difference and keep things consistent, as everywhere else a repo is a RepositoryHandle

vulder (approved):
Very nice clean up :100:

PR #903, opened 2025-02-18
Add arguments to patches

LuAbelt:
This pull request adds the possibility to use jinja placeholders in patches and provide arguments in the tool suite.

This enables more flexibility when using patches as one can just dynamically change e.g. regression sizes. This should also make the patches repository less cluttered in the long run.


varats-core/varats/provider/patch/patch_provider.py
```diff
@@ -157,6 +180,46 @@ def __str__(self) -> str:
...
+
+        Returns:
+            Path to the rendered patch
+        """
+        if not self.arguments:
+            return self.path
+
+        render_args = self.arguments.copy()
+        for key, value in kwargs.items():
+            render_args[key] = value
```
boehmseb, line 224: Should we fail if `kwargs` contains arguments that are not specified by the patch?

varats-core/varats/provider/patch/patch_provider.py
```diff
@@ -157,6 +180,46 @@ def __str__(self) -> str:
...
+
+        # Verify that all arguments are set
+        missing_args = [
+            arg for arg, value in render_args.items() if value is None
+        ]
+        if missing_args:
+            # TODO: Discuss whether we want to fail here.
+            # Theoretically, someone could have set a default value in the template file via the jinja syntax.
```
boehmseb: I think it's Ok to only fail when the jinja rendering fails.
This gives us more flexibility when creating patches (I can imagine that we would want this at some point) and we can still report an error if the template rendering fails, which happens later in this function.

varats-core/varats/provider/patch/patch_provider.py
```diff
@@ -157,6 +180,46 @@ def __str__(self) -> str:
...
+
+        # Create a temporary patch file with the rendered arguments
+        tmp_file = NamedTemporaryFile(delete=False)
+
+        # Render the patch with the arguments
+        loader = jinja2.FileSystemLoader(searchpath=Path(self.path).parent)
+        env = jinja2.Environment(loader=loader, keep_trailing_newline=True)
+
+        template = env.get_template(Path(self.path).name)
+        tmp_file.write(template.render(render_args).encode())
```
boehmseb: Some error handling here would be nice.
There are multiple potential error scenarios here, most importantly:
- loading the template fails
- rendering the template fails

varats-core/varats/experiment/steps/patch.py
```diff
@@ -51,20 +60,24 @@ class RevertPatch(actions.ProjectStep):
...
+        self.__arguments = kwargs
 
     def __call__(self) -> StepResult:
         self.status = StepResult.OK
+        print(
+            f"Reverting {self.__patch.shortname} on "
+            f"{self.project.source_of_primary}"
+        )
+
+        patch_path = self.__patch.render(**self.__arguments)
```
boehmseb: We also need to report a `StepResult.ERROR` here if template rendering fails.

varats-core/varats/provider/patch/patch_provider.py
```diff
@@ -157,6 +180,46 @@ def __str__(self) -> str:
...
+        ]
+        if missing_args:
+            # TODO: Discuss whether we want to fail here.
+            # Theoretically, someone could have set a default value in the template file via the jinja syntax.
+            raise ValueError(
+                f"Missing arguments for patch rendering: {', '.join(missing_args)}"
+            )
+
+        # Create a temporary patch file with the rendered arguments
+        tmp_file = NamedTemporaryFile(delete=False)
```
boehmseb: It might be a good idea to create these files in the experiment directory (`varats_root/benchbuild/results/...`) so that it is handled like all the other experiment-related files.
This probably needs information from an experiment step though.

boehmseb (changes requested):
I think we should put some more thought into some of the details.
I outlined these in my comments, but we can also discuss this further in person.

tests/provider/test_patch_provider.py
```diff
@@ -18,6 +19,157 @@
 )
 
 
+class TestPatchArguments(unittest.TestCase):
+
+    def __check_render(
+        self, rendered_patch: Path, expected_content: tp.List[str]
+    ):
+        with open(rendered_patch, "r") as patch_file:
```
vulder:
```suggestion
        with open(rendered_patch, "r", encoding="utf-8") as patch_file:
```

varats-core/varats/experiment/steps/patch.py
```diff
@@ -11,26 +11,34 @@
 from varats.utils.git_util import RepositoryHandle
 
 
+def _build_args_string(args: dict) -> str:
```
vulder: Did you mean kwargs or *args here? Passing a dict as args seems a bit confusing^^ If not, maybe give it a different name than `args` to not be confused with the python args/kwargs pattern

varats-core/varats/provider/patch/patch_provider.py
vulder: Hmm, hard to say. We could fail or just emit a warning.

Could it be that one collects a larger set of arguments and puts it into different patches and hence sometimes puts args in that are not "needed"?

varats-core/varats/provider/patch/patch_provider.py
vulder: If there is a default, I would not fail, however, we should detect the case where there is no default and a missing argument. Otherwise, finding typos or missing arguments will be a nightmare.

varats-core/varats/provider/patch/patch_provider.py
LuAbelt: I can see the following cases why there might be arguments not specified by the patch:

1. The case Flo described (If you want to apply/render multiple patches that share some arguments and don't overlap in the others it might be easier to just aggregate all arguments into one dict as opposed to building them per-patch)
2. Someone misspelled an argument name
3. An argument in the template was not specified in the patch info.

For 1. I don´ t see a big issue allowing these extra arguments. There might  be other, more general pitfalls when someone does this (e.g. accidentally overriding arguments for other patches) but that cannot really be solved in the rendering step.

For 2. the rendering would fail later regardless, so having a warning and a subsequent exception raised in the rendering is sufficient.

Case 3. is where I see potential issues in the future, when people just forget/ignore to properly add arguments to the patch info and work around it in their local setup. 

I am unsure whether the benefit of allowing it for 1. outweighs the potential problems caused by 3. But for 3. we could also setup other measures to avoid this (e.g. proper CI in the patches repository)

varats-core/varats/provider/patch/patch_provider.py
LuAbelt: I have changed the behavior to fail on rendering instead.

However, the error reported by jinja will only report the first missing variable it encounters. Should we add a note in the error message listing the potentially undefined variables based on this check?

varats-core/varats/provider/patch/patch_provider.py
LuAbelt: What would be the preferred behavior of the render function if it encounters any error? Currently I just raise a the general `TemplateError` from jinja.

Any other suggestions?

varats-core/varats/provider/patch/patch_provider.py
LuAbelt: I added the experiment step as an optional argument to the render function. If it is provided it will use the experiment directory for the rendered patch

PR #904, opened 2025-03-04
Update Path env in bbcfg after successfully running buildsetup build

Sinerum:
resolves se-sic/VaRA#1065


varats/varats/tools/driver_build_setup.py
```diff
@@ -204,6 +205,7 @@ def build(
             return
 
         if tool.verify_install(__get_install_prefix(tool, install_prefix)):
+            update_env(bb_cfg())
```
LuAbelt, line 208: Generally this looks like a reasonable change to me.

However, I see one issue:
The current implementation completely rebuilds the whole `env` entry in the config. In my setup for example, I have some other entries set in there besides `PATH` (e.g. `LD_LIBRARY_PATH`) which I need for some experiments. The current implementation would always remove those. I would prefer it if the only the `PATH` was updated/rebuild instead of the whole `env` entry.

Even then one could run into issues if people manually add things to the `PATH` for their setup. However I am not sure how frequent that is and if we should also consider this (and if so, how). Maybe @vulder and @boehmseb can provide more insights.

LuAbelt (changes requested):

varats/varats/tools/driver_build_setup.py
boehmseb: Lukas is right, we need to make sure that existing entries in `env` are preserved.
I suggest merging the new env dict onto the current one.
Btw., you could also add a call to `update_env` to `vara-gen-bbconfig --update`.

Sinerum: Keeping other `env` entries is a simple fix. If we want to also retain the original entries in `PATH` I don't see a solution that would not have the potential to accumulate entries over time if the install location of a tool changes.

boehmseb: That's true in theory, but the install location doesn't typically change.

LuAbelt (approved):

boehmseb (approved):

varats/varats/tools/bb_config.py
```diff
@@ -131,6 +131,21 @@ def update_experiments(bb_cfg: s.Configuration) -> None:
     ]
 
 
+def update_env(bb_cfg: s.Configuration) -> None:
+    """Update the given benchbuild config to contain our environment."""
+    old_env = bb_cfg["env"].value
+    old_path = old_env["PATH"]
```
boehmseb: This fails if the config env doesn't contain an entry for "PATH".

boehmseb (changes requested):

boehmseb (approved):

varats/varats/tools/driver_build_setup.py
```diff
@@ -27,6 +27,7 @@
 from varats.ts_utils.cli_util import initialize_cli_tool, cli_yn_choice
 from varats.ts_utils.click_param_types import EnumChoice
 from varats.utils.settings import save_config, vara_cfg, bb_cfg
+from varats.varats.tools.bb_config import update_env
```
boehmseb:
```suggestion
from varats.tools.bb_config import update_env
```

PR #910, opened 2025-03-05
Removes __init__.py from namespace packages.

boehmseb:
This is required for native (PEP 420) namespace packages, e.g., PyCharm fails to detect imports properly otherwise.


boehmseb:
Please test-install locally before approving.

LuAbelt (approved):
Install works and test suite passes for me.

PR #912, opened 2025-03-09
Implements out-of-tree loading support

vulder:
Patches initializations methods to consider out-of-tree repositories when loading projects/experiments/reports/plots or tables.

Through the new config option 'external_source_repositories', users can provide a list of external sources for loading projects/etc.

### DIFF
--- a/varats-core/varats/utils/external_source_handling.py
+++ b/varats-core/varats/utils/external_source_handling.py
@@ -1,0 +1,57 @@
+"""
+Module for handling external varats sources.
+
+This module provides loading facilities for loading custom user
+projects/experiments/etc. into the varats ecosystem.
+"""
+
+import importlib
+import logging
+import pkgutil
+import sys
+import typing as tp
+from pathlib import Path
+
+LOG = logging.getLogger(__name__)
+
+
+def determine_project_source_root(project_folder: Path) -> Path:
+    return project_folder / project_folder.name
+
+
+def relative_module_folder(project_folder: Path, module_folder: Path) -> Path:
+    return module_folder.absolute().relative_to(project_folder.absolute())
+
+
+def convert_path_to_module_path(path: Path) -> str:
+    return str(path).replace('/', '.')
+
+
+def load_python_modules_from_external_project(
+    project_folder: Path,
+    module_folder: Path,
+    export_context: tp.Optional[tp.List[tp.Any]] = None
+) -> None:
+    """
+    Load additional python modules from a external project.
+
+    Args:
+        project_folder: of the project we want to load from
+        module_folder: that we want to load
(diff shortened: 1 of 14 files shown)


varats-core/varats/utils/external_source_handling.py
```diff
@@ -0,0 +1,57 @@
...
+    return str(path).replace('/', '.')
+
+
+def load_python_modules_from_external_project(
+    project_folder: Path,
+    module_folder: Path,
+    export_context: tp.Optional[tp.List[tp.Any]] = None
+) -> None:
+    """
+    Load additional python modules from a external project.
```
boehmseb, line 36:
```suggestion
    Load additional python modules from an external project.
```

varats-core/varats/utils/settings.py
```diff
@@ -43,6 +43,12 @@ def create_new_varats_config() -> s.Configuration:
                 "desc": "Result folder for collected results",
                 "default": os.getcwd() + "/results",
             },
+            "external_source_repositories": {
+                "desc":
+                    "List of external repositories from where we local extra "
```
boehmseb, line 48:
```suggestion
                    "List of external repositories from where we load extra "
```

varats/varats/tools/driver_run.py
```diff
@@ -172,6 +172,7 @@ def main(
     )
 
     env = {k: str(to_yaml(v)) for k, v in bb_cfg().to_env_dict().items()}
+    env["PYTHONPATH"] = ":".join(sys.path)
```
boehmseb, line 175: I assume that is required so benchbuild knows about the loaded external modules?
Does this work then? I thought, benchbuild only discovers experiments/projects from the entries in its config file.
Also, could there be some interference with virtual environments or our container setup if we add the complete PATH as PYTHONPATH?

varats-core/varats/utils/external_source_handling.py
```diff
@@ -0,0 +1,57 @@
...
+
+    Args:
+        project_folder: of the project we want to load from
+        module_folder: that we want to load
+        export_context: into which we want to export our modules (e.g., __all__)
+    """
+    if str(project_folder) not in sys.path:
+        sys.path.insert(1, str(project_folder))
```
boehmseb, line 44: Is there another reason why we need to store the project folder in PATH besides the usage in `driver_run.py` below? If not, wouldn't it make more sense to pass this information in some other way separately from PATH?

boehmseb (commented):
Looks quite straight forward.
The only thing I'm unsure about is the handling of the PAHT/PYTHONPATH (see individual comments).

boehmseb:
Superseeded by #945

PR #913, opened 2025-03-13
Add a Flag to override Filestatus black and whitelists for vara-run

Sinerum:
resolves se-sic/VaRA#1026

### DIFF
--- a/varats/varats/tools/driver_run.py
+++ b/varats/varats/tools/driver_run.py
@@ -10,6 +10,7 @@ import logging
 import re
 import sys
 import typing as tp
+from operator import attrgetter
 from pathlib import Path
 from subprocess import PIPE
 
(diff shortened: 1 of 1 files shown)


Sinerum:
With the current implementation of version sampling for experiments `MISSING` revisions will alway be run regardless what  is specified by `FILE_STATUS_BLACKLIST` and `FILE_STATUS_WHITELIST`. 
Not sure if we want to change that as it likely does not come up often and the goal of the issue was to
 a) easily rerun successfull revisions after changing some part of the setup 
and
 b) specifically rerun compilation errors after fixing the project setup.

varats/varats/tools/driver_run.py
```diff
@@ -101,6 +103,22 @@ def __validate_project_parameters(
...
+    help="Override the file status whitelist."
+)
+@click.option(
+    "-bl",
+    "--black-list",
+    type=click.Choice([x.name for x in FileStatusExtension],
+                      case_sensitive=False),
+    multiple=True,
+    help="Override the file status blacklist."
+)
```
boehmseb: You can use `EnumChoice` from our helper module `click_param_types.py` here.

varats/varats/tools/driver_run.py
```diff
@@ -172,6 +197,7 @@ def main(
     )
 
     env = {k: str(to_yaml(v)) for k, v in bb_cfg().to_env_dict().items()}
+    env |= {k: str(to_yaml(v)) for k, v in vara_cfg().to_env_dict().items()}
```
boehmseb: I'm not sure if we should do this.
At the very least, this needs extensive testing as this could easily break our current slurm and/or container setup.
It's probably easier and safer for now to only add the modified config entries to `env`.

boehmseb (changes requested):

boehmseb (approved):

PR #914, opened 2025-04-02
Perf stat measurement

AKrissi:


.varats.yaml
LuAbelt, line 1: This seems to be left over from executing any of the `vara-*` commands from the wrong working dir. Better remove that

.vscode/launch.json
LuAbelt, line 1: Maybe try to configure your local VS code setup in a way that changes to this file are ignored and revert it to the origina;

benchbuild/.benchbuild.yml
LuAbelt, line 1: See above, this should be deleted

paper_configs/test_pc/PrimeNumbers_0.case_study
LuAbelt, line 1: Should be deleted

paper_configs/tutorial/xz_0.case_study
LuAbelt, line 1: should be deleted

plots/fperf_stat.svg
LuAbelt, line 1: delete

varats/varats/data/reports/perf_stat_report.py
```diff
@@ -0,0 +1,58 @@
...
+from pathlib import Path
+
+
+class PerfStatReport(BaseReport, shorthand="PERFSTAT", file_type="json"):
+    """
+    Converts perf stat output to a dictionary of data
+    """
+    def __init__(self, path: Path):
+        df = pd.DataFrame()
+        with open(path, "r") as file:
```
LuAbelt: If I remember correctly, pandas already has a functionality to directly load a dataframe from a json file. Is there a reason we parse this manually?

varats/varats/experiments/base/perf_sampling.py
LuAbelt, line 1: I think we pulled this in as an example from another branch. Do we still need/use this experiment for our current data collection?

varats/varats/plots/perf_stat_plots.py
```diff
@@ -0,0 +1,83 @@
...
+                    plt.title("Plot")
+                    plt.xlabel("time")
+                    plt.ylabel(self.plot_kwargs["value"])
+
+                    plt.xticks(rotation=90)
+
+class PerfStatPlotGenerator(
+    PlotGenerator, generator_name="fperf-stat", options=[make_cli_option(
+        "--value",
+        type=click.Choice(["task-clock", "context-switches", "cpu-migrations", "page-faults", "cycles",
```
LuAbelt: Not entirely sure whether we want this as a `Choice ` or rather just any string. Later we will probably collect other metrics than the default ones with perf stat

varats/varats/projects/c_projects/bzip2.py
LuAbelt, line 1: I need to double check if we still need to actually change these

LuAbelt (changes requested):
I have marked some of the more obvious changes. In some of the other files I still need to have a closer look and/or discuss what we actually want there.

What you can already do despite the changes I marked:
- Merge the current state of `vara-dev` into this branch (Will make ti easier to merge this branch back once it is finished)
- Add type annotations
- Fix formatting/linting issues
  - I recommend setting up [pre-commit](https://pre-commit.com/), which will automatically check this in the future before you commit. You can also run it manually on all the files in this PR, this should fix most of the issues the CI currently reports for this

PR #915, opened 2025-04-07
Add ZeroMQ benchmark binaries

LuAbelt:
Adds the internal benchmark suite of zeroMQ to the binary list and add support for recompilation


LuAbelt:
I am not sure why the pre-commit is failing in the pipeline. In my local pre-commit it always moves the benchbuild import around when I try to put it where the pipeline suggests to put it.

boehmseb:
> I am not sure why the pre-commit is failing in the pipeline. In my local pre-commit it always moves the benchbuild import around when I try to put it where the pipeline suggests to put it.

The CI is right, though.

boehmseb (approved):
The CI is right about the import, so something on your local system must be off :wink: 
Also, since you are working on this project anyways, you should update the container declaration to use Debian 12. Older versions no longer work on our clusters.

PR #916, opened 2025-04-07
If-Then Project step

LuAbelt:
Adds a project step that conditionally executes a follow up step based on the result of a step.


varats-core/varats/experiment/steps/ifthenelse.py
```diff
@@ -0,0 +1,72 @@
...
+    ) -> None:
+        super().__init__(project)
+        self.__condition = condition
+        self.__if_step = if_step
+        self.__else_step = else_step
+
+    def call_with_output_folder(self, tmp_dir: Path) -> StepResult:
+        result = _run_with_output_if_necessary(tmp_dir, self.__condition)
+
+        run_step = self.__if_step if result == StepResult.OK or result == StepResult.CAN_CONTINUE else self.__else_step
```
boehmseb: This condition can be simplified by checking if the result is in a list of allowed results.
With that: do we want to make this list a parameter of this step? The default would be the same as now, but one could easily customize the condition that way.

```suggestion
        run_step = self.__if_step if result in [StepResult.OK, StepResult.CAN_CONTINUE] else self.__else_step
```

varats-core/varats/experiment/steps/ifthenelse.py
```diff
@@ -0,0 +1,72 @@
...
+        ) + "\n"
+        outstr += textwrap.indent(f"* Then: ", indent * " ") + "\n"
+        outstr += textwrap.indent(
+            str(self.__if_step), (indent + 2) * " "
+        ) + "\n"
+        outstr += textwrap.indent(f"* Else: ", indent * " ") + "\n"
+        outstr += textwrap.indent(
+            str(self.__else_step), (indent + 2) * " "
+        ) + "\n"
+        return outstr
```
boehmseb: The `then` and `else` parts of the string should only be present if the corresponding steps are also present.

boehmseb (changes requested):
Still needs some cleanup (see pylint annotations).

Another idea: should we maybe name the file something like `combinators.py` in case we add more such combinator steps in the future?

boehmseb (approved):

PR #917, opened 2025-04-07
Integrate test suites for projects

LuAbelt:
This PR aims to integrate test suites for projects into our workflow. While benchbuild provides a `run_tests` function for their `Project` class, this lacks any interface to e.g. only run parts of a test suite.

This PR adds a new `SupportsTestSuites` protocol that projects can implement as well as some helper functions and implementations of the protocol for a couple of projects.


LuAbelt:
This has made some progress through the last week following some input from Sebi.

I have added a separate preparation and build step for tests, as for some functionality (collecting test names) one only needs to setup the build environment (e.g. run CMake) but not actually build the tests.

A thing that came to my mind while working on this:
Currently the `run_test_suite` only returns a `bool`, indicating whether the tests ran successfully or not (Current implementations just check the return code of the test executable).
However, maybe we want to make this more verbose by returning e.g. a `dict` that maps from test names to some kind of `TestStatus` enum (e.g. `SUCCESS`, `FAILURE`, `TIMEOUT`, `SKIPPED`, ...). 
In my opinion this would allow for more flexibility on the client side.
On the other hand, we would then most likely need some new dependencies to easily parse common test outputs such as JUnit.

tests/utils/test_testsuite_utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_ctest_get_test_names() -> None:
+    test_report_path = TEST_INPUTS_DIR / "ctest_get_names"
+    build_dir = test_report_path / "build"
+    test_names = ctest_get_test_names(build_dir)
+
+    expected_test_names = ["time", "fail", "timeout"]
+
+    assert test_names == expected_test_names
```
LuAbelt: @blutub This currently breaks the test suite, can you fix it please?
The `ctest_get_test_names` assumes that the passed path already exists and it can run ctest inside of it.
Without running cmake before it also won't generate the necessary files required for ctest to identify the test names.

For this to work properly, you can either (1) explicitly create the `build` directory and run `cmake` inside it before passing the path to `ctest_get_test_names` or (2) run `cmake` directly in your test directory and pass the directory (without `build`) to `ctest_get_test_names`

LuAbelt: @blutub Also, for the next time please make sure that the tests pass on your local machine before pushing ;-)

requirements.txt
```diff
@@ -5,6 +5,7 @@ distro>=1.5.0
 graphviz>=0.14.2
 ijson>=3.1.4
 Jinja2>=3.1.2
+junitparser
```
boehmseb: Please specify the desired version.

varats/varats/experiments/base/just_test.py
```diff
@@ -0,0 +1,112 @@
...
+from varats.report.report import ReportSpecification
+from varats.utils.config import get_current_config_id
+
+
+class JustTest(VersionExperiment, shorthand="JT"):
+    """Generates empty report file."""
+
+    NAME = "JustTest"
+
+    REPORT_SPEC = ReportSpecification(PlainTextReport)
```
boehmseb: This should be `TestsuiteReport` right?
```suggestion
    REPORT_SPEC = ReportSpecification(TestsuiteReport)
```

varats/varats/experiments/base/just_test.py
```diff
@@ -0,0 +1,112 @@
...
+            PrepareTestSuite(project),
+            BuildTestSuite(project),
+            RunTestSuite(project, Path(result_file.full_path())),
+            actions.Clean(project)
+        ]
+
+        return analysis_actions
+
+
+class CollectTestNames(VersionExperiment, shorthand="CTN"):
```
boehmseb, line 74: Is this only for debugging, or do you intend to keep this experiment?
If you want to keep it, I suggest renaming the report to something more specific.
Otherwise, there is no good way to interpret what this report is when only looking at its name.

varats/varats/projects/c_projects/libvpx.py
```diff
@@ -63,13 +102,81 @@ def compile(self) -> None:
         self.cflags += ["-fPIC"]
 
         clang = bb.compiler.cc(self)
+        cxx = bb.compiler.cxx(self)
         with local.cwd(libvpx_source):
-            with local.env(CC=str(clang)):
+            with local.env(CC=str(clang), CXX=str(cxx)):
```
boehmseb: I assume you want to use clang++ to compile C++ programs :)

```suggestion
            with local.env(CC=str(clang), CXX=str(clang++)):
```

tests/TEST_INPUTS/result.xml
```diff
@@ -0,0 +1,59 @@
...
+	disabled="0"
+	skipped="0"
+	hostname=""
+	time="0"
+	timestamp="2025-09-18T12:35:01"
+	>
+	<testcase name="compress_sample1.ref" classname="compress_sample1.ref" time="0.0424357" status="run">
+		<properties/>
+		<system-out>
+Running:  /home/blubub/Uni/hiwi/testing/bzip2/build/bzip2 -1
```
boehmseb: You might want to anonymize/pseudonymize the paths before merging.

varats-core/varats/experiment/steps/testsuite.py
```diff
@@ -0,0 +1,188 @@
...
+        }
+        return all(result_filter.get(status) for status in result.values())
+
+    def __call__(self) -> StepResult:
+        if not isinstance(self.project, SupportsTestSuites):
+            raise TypeError(
+                f"Project {self.project.name} does not support testing."
+            )
+        try:
+            self.project.prepare_test_environment()
```
boehmseb: Isn't this done already in the `PrepareTestSuite` step?
Why do you need to call this here again?

boehmseb (changes requested):

varats-core/varats/experiment/steps/testsuite.py
LuAbelt: That's true. Probably an oversight when we separated this into individual experiment steps

LuAbelt:
```suggestion
```

varats/varats/experiments/base/just_test.py
LuAbelt: This was mainly created for debugging very early when creating this PR. I would still keep it, as it was helpful to me sometimes when adding the protocol to new projects.

But I agree that then we should create a more specific report class.

varats/varats/projects/c_projects/libvpx.py
LuAbelt: Maybe I am missing something, but doesn't `benchbuild.compiler.cxx(...)` (L105) creates a (wrapped) c++ compiler that I pass here?

varats/varats/projects/cpp_projects/hyteg.py
```diff
@@ -143,3 +154,55 @@ def recompile(self) -> None:
...
+        """Prepare the testsuite."""
+        hyteg_source = local.path(self.source_of(self.primary_source))
+
+        mkdir("-p", hyteg_source / "build")
+
+        update_all_submodules(
+            RepositoryHandle(hyteg_source), recursive=True, init=True
+        )
+        # with local.cwd(hyteg_source):
+        #     local["git"]["submodule"]("update", "--init", "--recursive")
```
LuAbelt:
```suggestion
```

varats/varats/projects/c_projects/libvpx.py
boehmseb: Oh, yes. My bad.
I misread this part.

varats-core/varats/project/varats_project.py
```diff
@@ -33,15 +77,70 @@ def binaries(self) -> tp.List['ProjectBinaryWrapper']:
     @staticmethod
     @abstractmethod
     def binaries_for_revision(
-        revision: ShortCommitHash  # pylint: disable=W0613
-    ) -> tp.List['ProjectBinaryWrapper']:
+        revision: ShortCommitHash,  # pylint: disable=W0613
```
boehmseb: Please replace `pylint: disable` with appropriate `noqa` annotation.

varats/varats/tools/bb_config.py
```diff
@@ -167,7 +172,10 @@ def create_new_bb_config(
     Returns:
         a new default bb config object
     """
-    from benchbuild.settings import CFG as BB_CFG  # pylint: disable=C0415
+    from benchbuild.settings import (  # noqa: PLC0415
+        CFG as BB_CFG,  # pylint: disable=C0415
```
boehmseb: Replace with `noqa`.

varats/varats/tools/bb_config.py
```diff
@@ -80,11 +84,12 @@ def update_projects(
         'varats.projects.cpp_projects.ect',
         'varats.projects.cpp_projects.lepton',
         'varats.projects.cpp_projects.hyteg',
-        'varats.projects.cpp_projects.dune'
-    ]
+        'varats.projects.cpp_projects.dune',
+    ]  # yapf: disable
```
boehmseb: Use `# fmt: skip` instead.

boehmseb (approved):
Could use some more tests for testsuite.py and testsuite_utils.py.
Rest looks good.

PR #919, opened 2025-04-08
Manually specify module locations in pyproject.toml

boehmseb:
Setuptools cannot autodetect our current project layout.
This leads to some modules missing from the built project.

There are, in principal, two ways how we can fix this:
1. Manually specify all module locations
2. Change our project layout

This PR implements option 1, as it was easier to implement, but please give your opinions if this is the way we want to fix this issue.


boehmseb:
@LuAbelt @vulder Any comments on this PR?
If no, I would like to merge this soon.

LuAbelt (approved):
Tried an install in a fresh `virtualenv` and encountered no issues when running some basic `vara-*` commands

vulder (approved):

PR #924, opened 2025-04-15
refactor: return a set in sampling_method_names to fater lookup

allrob23:
This PR updates the `sampling_method_names` method in `SamplingMethodBase` to return a `set` instead of a `list` for the registered sampling method names. This refactor aligns the return type with the conceptual uniqueness of method names in the registry, improving code clarity and preparing for potential future scalability. Using a `set` ensures `O(1)` lookup time for membership checks (e.g., using the in operator) compared to `O(n)` for a `list`, making the code more robust for future use cases.

The behavior of the method and its associated tests remains unchanged.

### DIFF
--- a/varats-core/varats/base/sampling_method.py
+++ b/varats-core/varats/base/sampling_method.py
@@ -32,13 +32,13 @@ class SamplingMethodBase(tp.Generic[SamplingMethodSubType], abc.ABC):
         cls._methods[cls.name()] = cls
 
     @classmethod
-    def sampling_method_names(cls) -> tp.List[str]:
+    def sampling_method_names(cls) -> tp.Set[str]:
         """
-        Returns a list of all registered sampling method names.
+        Returns a set of all registered sampling method names.
 
-        Returns: list of sampling method names
+        Returns: set of sampling method names
         """
-        return list(cls._methods.keys())
+        return set(cls._methods.keys())
 
     @classmethod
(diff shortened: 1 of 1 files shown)


vulder:
I currently do not see any need/motivation for this change, even with regard to the given motivation as this function is seldomly invoked. Furthermore, in contrast to what the description promises, there is actually a semantically important change by switching from a list to a set here, which is related to the ordering guarantees of both container types. 

PS: This change, without actual motivation, seems to be a randomly auto generated drive by commit.

allrob23:
Hi @vulder, thanks for your feedback! Iget your point about the change seeming unmotivated. Let me provide some context: this PR stems from a research project we’re working on, where we noticed in the `test_registering_new_sampling_methods` test that we’re doing membership checks (`in`) on the returned `list`. Switching to a set makes these checks more efficient (`O(1)` vs. `O(n)`), which we thought could be a small but meaningful improvement, especially for future scalability.

image


Regarding the ordering concern, I see your point about sets not preserving order like lists do. In this case, since the method’s purpose is to return unique method names and the tests don’t rely on ordering, I figured a set aligns better with that intent. That said, I’m happy to revisit this if you think preserving order is important for other reasons I might’ve missed.

My goal here is to contribute something helpful to the codebase, so I’m open to discussing this further or dropping the change if it doesn’t add enough value. Let me know what you think!

PR #928, opened 2025-06-12
Fix how workload_commands filter binaries/workloads

LuAbelt:
In the current implementation `workload_commands` compares the name of the binary with the stem of the command.

I noticed that this causes issues for (at least) two cases:
- When using `RevisionBinaryMap.specify_binary` the binary name is inferred from the `stem` of the binary path (Fails if your binary has any file ending, like when registering python driver scripts as a binary)
- The binary name can be manually set when using `RevisionBinaryMap.specify_binary` by providing `override_binary_name` which can then be completely different from the command path altogether

### DIFF
--- a/tests/experiment/test_workload_util.py
+++ b/tests/experiment/test_workload_util.py
@@ -1,4 +1,5 @@
 """Test VaRA workload utilities."""
+import typing as tp
 import unittest
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


LuAbelt:
Requires #929 to fix test names for the pipelines

boehmseb (approved):

PR #934, opened 2025-10-17
Add a fetch interval to the Patch Provider

LuAbelt:
This adds a fetch interval to the patch provider to reduce the number of fetches made to GitHub repositories.

Previously each instantiation of a patch provider enforced a fetch on the projects repo to correctly identify the valid revisions of a patch. 
This leads to issues with experiment steps that use the `workload_commands` function from `workload_util`, as each call to  `can_be_executed` (from `VProjectCommand`) creates a new PatchProvider. 

This can lead to severe rate limiting by GitHub and ultimately failing experiments.

### DIFF
--- a/varats-core/varats/provider/patch/patch_provider.py
+++ b/varats-core/varats/provider/patch/patch_provider.py
@@ -282,10 +282,11 @@ class PatchProvider(Provider):
         shallow=False
     )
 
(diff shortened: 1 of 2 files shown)


LuAbelt:
After reading through the description again - Would it also make sense to just cache already created PatchProviders and return the cached provider from `create_provider_for_project`?

PR #935, opened 2025-11-16
Support Python 3.14 and drop support for 3.9/10

ReebaAslam:
This pull request updates the project's Python version support and adjusts dependency constraints accordingly. The main changes are focused on dropping support for older Python versions and ensuring compatibility with newer ones.

**Python version support updates:**

* Updated the CI workflow in `.github/workflows/ci.yml` to test against Python 3.11, 3.12, 3.13, and 3.14, removing 3.9 and 3.10 from the test matrix.
* Changed the `requires-python` field in `pyproject.toml` to require Python 3.11 or newer.

**Testing**
* Ran following commands to run an experiment on both versions:
```
mkdir $VARATS_ROOT
cd $VARATS_ROOT

vara-gen-bbconfig
vara-pc create tutorial
vara-cs gen -p xz select_specific e7da44d515
vara-run -E JustCompile
```

* Experiments ran successfully on both versions
* Ran test using the following command:
```
mkdir -p benchbuild
coverage run -p -m pytest varats varats-core tests
rm -rf benchbuild
```
* On both versions, all test cases passed except one locally which required a container. 
```FAILED tests/test_containers.py::TestContainerSupport::test_create_stage_10_from_source - AttributeError: benchbuild.utils.error_cmd```
* Github check passed all test cases on 3.14
* Warnings count increased up to 11345 from 8 upgrading the version. 

**Dependency updates:**

* Relaxed the `pygit2` dependency constraint in `pyproject.toml` to allow versions 1.16.0 and higher, since support for Python 3.9 is no longer needed.

* [pip_freeze_diff.pdf](https://github.com/user-attachments/files/23569051/pip_freeze_diff.pdf) shows the difference in versions of dependencies between Python3.9.25 and Python3.14.0 

**TODOs**
- [x] Fix warnings raised during tests under Python 3.14
<details>
<summary><strong>Warning Analysis Table (Python 3.14 Compatibility, created by chatgpt so will have to self-review it as well)</strong></summary>

<br>

| **Category** | **Files Affected** | **Cause** | **Impact** | **Recommended Fix** |
|--------------|--------------------|-----------|------------|----------------------|
| **SyntaxWarnings (invalid escape sequences)** | - `varats-core/varats/report/tef_report.py`<br>- `varats/varats/tables/feature_perf_precision.py` | Regex patterns use invalid escape sequences like `"\d"` and `"\D"` | Will become errors in future Python versions | Use raw strings `r"\d"` or escape with `"\\d"` |
| **DeprecationWarnings (Click MultiCommand)** | - `driver_casestudy.py`<br>- `driver_plot.py`<br>- `driver_table.py` | `MultiCommand` deprecated in Click; replaced by `Group` | Will break when Click 9.0 is released | Replace `MultiCommand` with `Group` (low priority) |
| **Pandas FutureWarning** | - `cache_helper.py` | Assigning incompatible dtype (object → int64 column) | Will become an error in future pandas versions | Explicitly cast to compatible dtype before assignment |
| **ijson DeprecationWarning** | - `ijson/compat.py` (third-party) | File opened in text mode instead of binary mode | Will error in future ijson releases | Ensure file-like objects are opened in binary mode (`"rb"`) |
| **datetime.utcfromtimestamp DeprecationWarning** | - `varats/varats/data/reports/blame_report.py` | `utcfromtimestamp()` deprecated in Python 3.14 | Breaks in future Python versions | Use `datetime.fromtimestamp(timestamp, datetime.UTC)` |
| **GitHub API DeprecationWarning** | - `github_util.py` | `login_or_token=` deprecated | Will break with future PyGithub updates | Use `auth=github.Auth.Token(...)` instead |

</details>

- [x] Verified all updated dependencies (from requirements.txt and ) support Python 3.14.

- [x] Run repository formatting tools via pre-commit under python3.14. All ran successfully except docformatter (docformatter hook was been commented out in .pre-commit-config.yaml to allow the rest of the pre-commit toolchain to run, added following lines at the top)
```
default_language_version:
  python: python3.14
```
- [ ] Resolve docformatter Incompatibility with Python 3.14. The pre-commit hook docformatter fails to install under Python 3.14 due to an upstream dependency issue. 
- [x] Static Type Checking (already run as part of CI pipeline but need to confirm if it is reliable)
- [ ] Review test coverage to find any untested or weakly tested code paths, especially in places that showed Python 3.14 warnings or depend on updated libraries, and figure out if they need extra tests or small compatibility fixes.
- [ ] Update python version reference in any documentation

### CODE REVIEW

File: varats/varats/data/cache_helper.py [hunk 1]
```diff
@@ -173,6 +173,10 @@ def is_newer_file(report_file: InDataTy) -> bool:
             create_cache_entry_data, data_entry
         )
         updated_entry.set_index(CACHE_ID_COL, inplace=True)
+        # TODO[temporary]: updated_entry.cache_timestamp has str values, which is in line with the __create_cache_entry function,
```
ReebaAslam: image

Tested by casting that the warning gets resolved

pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
   "pandas>=1.5.3",
   "plotly>=5.13.1",
   "plumbum>=1.6",
-  "pygit2>=1.15.0,<1.16.0", # 1.16 drops support for python 3.9
+  "pygit2>=1.15.0", # 1.16 drops support for python 3.9
```
ReebaAslam: should I change this to 1.16.0? or is that unnecessary?

ReebaAslam:
Docs mentioning python version (not exhaustive), should be updated?
- Readme.md -> python3.6 needed at least for installation. is it still true?
- python_scripting_tips.rst -> MIN_PYTHON
- getting_started.rst -> python3.7 min required

@boehmseb can you confirm?

varats/varats/data/cache_helper.py
ReebaAslam: @boehmseb do you have a quick solution? I was thinking of updating the dftypes sent to the function getting the cached_df.

varats/varats/tables/feature_perf_precision.py
```diff
@@ -388,9 +388,10 @@ def tabulate(self, table_format: TableFormat, wrap_table: bool) -> str:
 
         # Rename columns
         # pylint: disable=anomalous-backslash-in-string
-        overhead_time_c_name = "$\Delta$ Time $(\%)$"
-        overhead_memory_c_name = "$\Delta$ Mem $(\%)$"
-        overhead_memory_val_c_name = "$\Delta$ Mem $(Kbyte)$"
+        # TODO[Reeba]: use raw strings
```
boehmseb:
```suggestion
```

pyproject.toml
boehmseb: I think we should require the newest major revision if that doesn't break anything.
```suggestion
  "pygit2>=1.19.0,<1.20.0",
```

requirements.txt
```diff
@@ -16,7 +16,7 @@ plotly>=5.13.1
 plumbum>=1.6
 pre-commit>=3.2.0
 PyDriller>=2.4.1
-pygit2>=1.15.0,<1.16.0
+pygit2>=1.15.0
```
boehmseb:
```suggestion
pygit2>=1.19.0,<1.20.0
```

varats/varats/tools/driver_plot.py
```diff
@@ -29,7 +29,7 @@
 LOG = logging.getLogger(__name__)
 
 
-class PlotCLI(click.MultiCommand):
+class PlotCLI(click.Group):
```
boehmseb, line 32: Does this work just changing `MultiCommand` to `Group`?
Or is there still some migration for our automatic command injection required?

pyproject.toml
boehmseb: We should also increase the minor version with this change.
```suggestion
version = "14.1.0"
```

tests/report/test_tef_report.py
```diff
@@ -214,10 +215,43 @@ class TestTEFReportParser(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         """Load and prepare TEF report."""
-        with mock.patch(
-            'builtins.open',
-            new=mock.mock_open(read_data=TRACE_EVENT_FORMAT_OUTPUT)
-        ):
+        file_data = {'content': TRACE_EVENT_FORMAT_OUTPUT}
+
+        def mock_open_side_effect(path, mode='r', *args, **kwargs):
```
boehmseb, line 219: Why is this change necessary?

varats/varats/data/cache_helper.py
boehmseb: What is the original warning here? It is hard to say what a good solution is without.

varats/varats/plots/blame_interaction_graph_plots.py
```diff
@@ -119,8 +119,13 @@ def filter_nodes(node: CommitRepoPair) -> bool:
         if not commit:
             return False
         # make filter configurable
-        return datetime.utcfromtimestamp(commit.commit_time
-                                        ) >= datetime(2015, 1, 1)
+
+        #TODO[temporary]: Not covered by tests, should be added
```
boehmseb: Our plots currently aren't covered by tests anyways, so it is fine if this change is uncovered.
How we could effectively test our plots is its own whole topic.

tests/report/test_tef_report.py
ReebaAslam: We were receiving this warning:

```
"tests/report/test_tef_report.py::TestTEFReportParser::test_parse_stack_frames
  /opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/ijson/compat.py:32: DeprecationWarning:
  
  
  ijson works by reading bytes, but a string reader has been given instead. This
  probably, but not necessarily, means a file-like object has been opened in text
  mode ('t') rather than binary mode ('b').
  
  An automatic conversion is being performed on the fly to continue, but on the
  other hand this creates unnecessary encoding/decoding operations that decrease
  the efficiency of the system. In the future this automatic conversion will be
  removed, and users will receive errors instead of this warning. To avoid this
  problem make sure file-like objects are opened in binary mode instead of text
  mode."
```

`TEFReport._parse_json` opens a file in binary form and ijson parses it. This is where the error comes in. 
I tried converting the `TRACE_EVENT_FORMAT_OUTPUT` string to a binary string but then another function within the `TEFReport` gave an error because it opened the same file in a different format. 

We needed a mock that can handle both types.

varats/varats/data/cache_helper.py
ReebaAslam:
```
tests/data/test_cache_helper.py::TestCacheHelper::test_build_cached_report_table
  /home/runner/work/VaRA-Tool-Suite/VaRA-Tool-Suite/varats/varats/data/cache_helper.py:176: FutureWarning:
  
  Setting an item of incompatible dtype is deprecated and will raise in a future error of pandas. Value '[1 1 '2']' has dtype incompatible with int64, please explicitly cast to a compatible dtype first.
```

varats/varats/plots/blame_interaction_graph_plots.py
ReebaAslam: Ok i will remove these ToDos then

tests/report/test_tef_report.py
ReebaAslam: Realized the function was unnecessarily complicated, so have simplified it.

varats/varats/tools/driver_plot.py
ReebaAslam: So i ran the tests that were using the dynamic commands like `pc-overview-plot` and also checked diff in output of `vara-plot --help`.  There was no change in the commands that were listed down. Let me know if this does not seem sufficient. 

Image

boehmseb:
> Docs mentioning python version (not exhaustive), should be updated?
> 
>     * Readme.md -> python3.6 needed at least for installation. is it still true?
> 
>     * python_scripting_tips.rst -> MIN_PYTHON
> 
>     * getting_started.rst -> python3.7 min required
> 
> 
> @boehmseb can you confirm?

These should all be updated.

tests/data/test_cache_helper.py
```diff
@@ -37,7 +42,9 @@ def test_build_cached_report_table(self):
         project_name = "project"
 
         def create_empty_df():
-            return pd.DataFrame(columns=["entry"])
+            df = pd.DataFrame(columns=list(self.column_types.keys()))
+            df = df.astype(self.column_types)
+            return df
```
ReebaAslam, line 47: @boehmseb the old line assumes all columns defined will be strings, which is fine, but I wanted us to be explicit, so updated the original function

varats/varats/data/databases/evaluationdatabase.py
```diff
@@ -27,11 +27,7 @@ class EvaluationDatabase(abc.ABC):
 
     CACHE_ID: str
     COLUMN_TYPES = {"revision": 'str', "time_id": 'int32'}
-    # TODO: Might need to add "COLUMN_TYPES.update(CACHE_COL_TYPES)"
-    # as Pandas Future Warning for incompatible dtypes can occur
-    # if we are reading cached df that has timestamp as int
-    # but currently adding it is failing some tests
-
+    COLUMN_TYPES.update(CACHE_COL_TYPES)
```
ReebaAslam: @boehmseb I tried adding the line here, and some test cases failed because they were reading from an old cache that did not have the cached revision and timestamp columns, so I had to add backward compatibility as well.

There is one scenario of `build_cached_report_table` which would require this change. 

Scenarios:

A - cache exists + cache has timestamp col + data_to_load or data_to_drop
B - cache exists  + cache does not have timestamp col + data_to_load or data_to_drop
C - cache exists + no data_to_load and no data_to_drop
D - cache does not exists  + data_to_load 
E - cache does not exists + no data_to_load and no data_to_drop

E is a useless func call so will ignore it. 
Scenario C and D will work perfectly. 

If Scenario A happens, we will face the Pandas Warning, hence the change.

If Scenario B happens, function will throw error, as it assumes cache should have the timestamp column when loading or dropping data, which means that old cache should have those columns when we are loading new data. **I assume this scenario never happens as this does not have anything to do with the PR**?

**Question**

Does this make sense? Should I move these changes to a separate PR? We do not get any warnings or errors in our unit tests, or Just Compile experiment without this change but we should if scenario A is to happen.

varats/varats/data/cache_helper.py
```diff
@@ -68,6 +69,8 @@ def cache_dataframe(
         dataframe: pandas dataframe to store
     """
     file_path = get_data_file_path(data_id, project_name)
+    # TODO: we should add quoting = csv.QUOTE_NONNUMERIC here, 
+    # so that int/float casted to str are written as strings in csv
```
ReebaAslam, line 73: @boehmseb this could also be moved to another PR. I could not go over each col every database adds to cache which might be casting int/float to str. Those cols will again cause the FutureWarning, but for the sake of this PR, we can ignore it, and resolve separately.

boehmseb: I would ignore this for now and assume pandas uses sane defaults for conversion here.

boehmseb (approved):

PR #938, opened 2026-02-24
Replacing some linting and formatting tools with Ruff

ReebaAslam:
Following tools are replaced in this PR:
1. Isort
2. Docformatter
3. Yapf
4. Pylint

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -28,8 +28,8 @@ jobs:
         pip install .
         pip install -r requirements.txt
         pip install mypy
-        pip install pylint
-        pylint --version
+        pip install ruff==0.15.2
+
     - name: Setup reviewdog
       run: |
         mkdir -p $HOME/bin && curl -sfL https://raw.githubusercontent.com/reviewdog/reviewdog/master/install.sh| sh -s -- -b $HOME/bin
@@ -40,4 +40,4 @@ jobs:
         REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         reviewdog -runners=mypy -reporter=github-pr-check -conf=.reviewdog.yml
-        reviewdog -runners=pylint -reporter=github-pr-check -conf=.reviewdog.yml
+        reviewdog -runners=ruff -reporter=github-pr-check -conf=.reviewdog.yml
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -40,4 +40,4 @@ jobs:
         REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         reviewdog -runners=mypy -reporter=github-pr-check -conf=.reviewdog.yml
-        reviewdog -runners=ruff -reporter=github-pr-check -conf=.reviewdog.yml -filter-mode=file
+        reviewdog -runners=ruff -reporter=github-pr-check -conf=.reviewdog.yml -filter-mode=diff_context
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -40,4 +40,4 @@ jobs:
         REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         reviewdog -runners=mypy -reporter=github-pr-check -conf=.reviewdog.yml
-        reviewdog -runners=ruff -reporter=github-pr-check -conf=.reviewdog.yml
+        reviewdog -runners=ruff -reporter=github-pr-check -conf=.reviewdog.yml -filter-mode=file
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,7 +1,0 @@
-[settings]
-known_first_party = varats
-known_third_party = benchbuild
-multi_line_output=3
-use_parentheses = True
-include_trailing_comma: True
-line_length=80
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,6 +116,17 @@ markers = [
     "slow: this test is slow."
 ]
 
+[tool.ruff]
+line-length = 80
+
+[tool.ruff.lint]
+select = ["I"]  # I = import-sorting rules only
+
+[tool.ruff.lint.isort]
+known-first-party = ["varats"]
+known-third-party = ["benchbuild"]
+
+
 [tool.mypy]
 warn_return_any = true
 warn_unused_configs = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -128,6 +128,9 @@ select = [
   "E",      # pycodestyle errors
   "W",      # pycodestyle warnings
 
+  # pydocstyle
+  "D",
+
   # Ruff-specific rules
   "RUF",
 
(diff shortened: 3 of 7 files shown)


pyproject.toml
```diff
@@ -116,6 +116,68 @@ markers = [
...
+  "PLC",    # pylint convention rules
+]
+
+# Ignoring specific rules that conflict with pre-ruff setup or are not needed
+ignore = [
+  "PLW1514", # unspecified-encoding
+  "G004", # logging-f-string
+  "ISC002", # multi-line-implicit-string-concatenation
+  "N815",  # mixed-case-variable-in-class-scope
+  "PLW0177", # comparison-to-nan
```
ReebaAslam:
```suggestion
```

pyproject.toml
```diff
@@ -116,6 +116,68 @@ markers = [
...
+  "PLW2901", # redefined-loop-variable
```
ReebaAslam:
```suggestion
```

pyproject.toml
```diff
@@ -116,6 +116,68 @@ markers = [
...
+  "PLR2004", # magic-value-comparison
+  "PLR5501", # collapsible-else-if
+  "PLW1641", # eq-without-hash
```
ReebaAslam:
```suggestion
```

pyproject.toml
```diff
@@ -116,6 +116,68 @@ markers = [
...
+  # Full Pylint coverage available in Ruff
+  "PLE",    # pylint errors
+  "PLW",    # pylint warnings
+  "PLR",    # pylint refactor
+  "PLC",    # pylint convention rules
+]
+
+# Ignoring specific rules that conflict with pre-ruff setup or are not needed
+ignore = [
+  "PLW1514", # unspecified-encoding
```
ReebaAslam:
```suggestion
```

pyproject.toml
```diff
@@ -116,6 +116,68 @@ markers = [
...
+  # Naming conventions (replaces BASIC naming rules)
+  "N",
+
+  # McCabe complexity (replaces DESIGN complexity only)
+  "C901",
+
+  # Full Pylint coverage available in Ruff
+  "PLE",    # pylint errors
+  "PLW",    # pylint warnings
+  "PLR",    # pylint refactor
```
ReebaAslam, line 171: PLR0402 needs to be assessed again, are both kinds of imports the same? better to exclude

pyproject.toml
```diff
@@ -116,6 +116,68 @@ markers = [
...
+  "PLC",    # pylint convention rules
+]
+
+# Ignoring specific rules that conflict with pre-ruff setup or are not needed
+ignore = [
+  "PLW1514", # unspecified-encoding
+  "G004", # logging-f-string
+  "ISC002", # multi-line-implicit-string-concatenation
```
ReebaAslam, line 193:
```suggestion
```

pyproject.toml
```diff
@@ -116,6 +116,68 @@ markers = [
...
+  "N815",  # mixed-case-variable-in-class-scope
```
ReebaAslam:
```suggestion
```

pyproject.toml
```diff
@@ -116,6 +116,68 @@ markers = [
...
+  "W",      # pycodestyle warnings
+
+  # Import sorting (replaces isort)
+  "I",
+
+  # Naming conventions (replaces BASIC naming rules)
+  "N",
+
+  # McCabe complexity (replaces DESIGN complexity only)
+  "C901",
```
ReebaAslam: remove this

pyproject.toml
```diff
@@ -116,6 +116,68 @@ markers = [
     "slow: this test is slow."
 ]
 
+[tool.ruff]
+line-length = 80
+exclude = ["CVS", "gui", "filtertree_data.py", ".eggs", "docs/source/conf.py",]
+indent-width = 4
+
+[tool.ruff.lint]
+select = [
```
ReebaAslam, line 125: Add COM, A, C4, RET, ARG, PTH, 
try SIM, PD, NMPY, RUF, FURB

when doing type annotation thing do TC, FA, UP

pyproject.toml
```diff
@@ -116,6 +116,64 @@ markers = [
...
+
+  # Naming conventions (replaces BASIC naming rules)
+  "N",
+
+  # flake8 rules
+  "COM",    # flake8-commas
+  "A",      # flake8-builtins
+  "C4",     # flake8-comprehensions
+  "RET",    # flake8-return
+  "ARG",    # flake8-unused-arguments
```
ReebaAslam, line 160: @boehmseb we might want to re-assess ARG rules because some of this is being prompted on child classes that seem to be defining the arg because they are overriding parent class method. 

Image

pyproject.toml
ReebaAslam: @boehmseb correct me if I am wrong but [PLR0402](https://docs.astral.sh/ruff/rules/manual-from-import/#builtins) is problematic if __init__ of the module defines an attribute of the same name as a sub-module. For us, the rule shows up for standard libraries e.g. `unittest.mock` or `benchbuild.utils`, so they should not have this issue and it should be fine to keep it.

ReebaAslam:
@boehmseb pending checks still show pylint, do you know why?

.github/workflows/reviewdog.yml
```diff
@@ -40,4 +40,4 @@ jobs:
         REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         reviewdog -runners=mypy -reporter=github-pr-check -conf=.reviewdog.yml
-        reviewdog -runners=pylint -reporter=github-pr-check -conf=.reviewdog.yml
+        reviewdog -runners=ruff -reporter=github-pr-check -conf=.reviewdog.yml -filter-mode=file
```
ReebaAslam: @boehmseb default filter mode is added lines only. I can remove the filter mode but wanted to discuss which one we want to have?

.isort.cfg
ReebaAslam, line 1: @boehmseb i deleted this because we migrated completely, should I delete other configs as well or restore this one as well?

pyproject.toml
```diff
@@ -116,6 +116,100 @@ markers = [
...
+  "PLR",    # pylint refactor
+  "PLC",    # pylint convention rules
+]
+
+# Ignoring specific rules that conflict with pre-ruff setup or are not needed
+ignore = [
+  "G004",     # logging-f-string
+  "PLR2004",  # magic-value-comparison
+  "PLR5501",  # collapsible-else-if
+  "FIX004",   # line-contains-hack
```
ReebaAslam, line 180: @boehmseb was not specified in pylint so added here

pyproject.toml
```diff
@@ -116,6 +116,117 @@ markers = [
...
+  "D300",   # triple-single-quotes
+  "COM812", # missing-trailing-comma
+  "COM819", # prohibited-trailing-comma
+  "ISC002", # multi-line-implicit-string-concatenation
+]
+
+
+[tool.ruff.lint.pydocstyle]
+# Use Google-style docstrings.
+convention = "google"
```
ReebaAslam, line 199: @boehmseb docformatter was not following this, it mostly followed pep257. (one difference is adding an empty line before summary in docstring)

https://docformatter.readthedocs.io/en/latest/requirements.html

boehmseb (approved):

PR #943, opened 2026-04-30
Patch Provider improvement

blutub:
1. subdirectories issue
2. specifiying the patch provider repositories in .varats

### DIFF
--- a/varats-core/varats/provider/patch/patch_provider.py
+++ b/varats-core/varats/provider/patch/patch_provider.py
@@ -27,6 +27,7 @@ from varats.utils.git_util import (
     get_initial_commit,
     RepositoryHandle,
 )
+from varats.utils.settings import vara_cfg
 
 
 class Patch:
@@ -285,7 +286,12 @@ class PatchProvider(Provider):
     def __init__(self, project: tp.Type[Project]):
         super().__init__(project)
 
-        self._update_local_patches_repo()
+        # Accessing the configuration (.yml) file
+        cfg = vara_cfg()
+
+        repos = cfg["patch_provider"]["repositories"].value
+
+        self._update_local_patches_repo(repos)
         repo = self._get_patches_repository()
 
         patches_project_dir = repo.worktree_path / self.project.NAME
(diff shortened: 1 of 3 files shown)


varats-core/varats/utils/benchbuild_util.py
```diff
@@ -0,0 +1,19 @@
...
+from contextlib import contextmanager  # contextmanager
+
+from benchbuild.settings import CFG  #  import CFG for the main change
+
+"""Utility functions mainly to change benchbuild settings.
+"""
+
+
+@contextmanager
+def temporary_tmp_dir():
```
LuAbelt, line 12: This is currently very specific to the patch provider.

I was thinking about a more general context manager for modifying the benchbuild config temporarily.
So essentially that one can put in an arbitrary amount of kwargs to modify the config. Basically similar to how plumbum does it with `local.env` (https://plumbum.readthedocs.io/en/latest/local_machine.html#environment)

Maybe take a look whether you could adapt the `bb_cfg()` function in `varats-core/varats/utils/settings.py` to add the context manager functionality.

The syntax I have in mind to use it would be something like:
```
with bb_cfg(tmp_dir=...):
  # Do some operations, but the tmp_dir is changed

# Similarly, you can still just read any value from the bb_cfg:
foo = bb_cfg()["some_key"]
```

varats-core/varats/utils/settings.py
```diff
@@ -43,35 +44,31 @@ def create_new_varats_config() -> s.Configuration:
                 "desc": "Result folder for collected results",
                 "default": os.getcwd() + "/results",
```
LuAbelt, line 45: ruff is still complaining about some uses of os.cwd(), maybe you could fix those along the way.

varats-core/varats/provider/patch/patch_provider.py
```diff
@@ -272,51 +310,78 @@ def __repr__(self) -> str:
...
+            local_cfg = local.str().split("/").join("_")
+            refspec_cfg = "origin/HEAD"
+            patches_sources.append(
+                bb.source.Git(
+                    remote=remote_cfg,
+                    local=local_cfg,
+                    refspec=refspec_cfg,
+                    limit=None,
+                    shallow=False,
+                )
```
LuAbelt, line 351: I don't think this makes too much sense.

Basically, this assumes that everything that people add as a local source is also a git repository. I would just treat local sources as standard directories, where you iterate over the info files and skip all git specific parts.

varats-core/varats/provider/patch/patch_provider.py
```diff
@@ -272,51 +310,78 @@ def __repr__(self) -> str:
...
+
+            for info_path in patches_project_dir.rglob("*.info"):
                 try:
                     current_patch = Patch.from_yaml(info_path)
                     self.__patches.add(current_patch)
                 except YAMLError:
                     warnings.warn(
-                        f"Unable to parse patch info in: '{filename}'"
+                        f"Unable to parse patch info in: `{info_path}`"
                     )
```
LuAbelt, line 382: Move into a separate function that takes a path as an argument and parses all patches in this structure

varats-core/varats/provider/patch/patch_provider.py
```diff
@@ -272,51 +310,78 @@ def __repr__(self) -> str:
...
+                patches_project_dir = repo.worktree_path / self.project.NAME
 
-        # Update repository to have all upstream changes
-        project_repo = get_local_project_repo(self.project.NAME)
-        fetch_repository(project_repo)
+                if not patches_project_dir.is_dir():
+                    warnings.warn(
+                        "Could not find patches directory for project "
+                        f"'{self.project.NAME}'."
+                    )
```
LuAbelt, line 367: This could be a separate function

PR #945, opened 2026-05-08
Vara external

ReebaAslam:
Tool to set external repository.


varats/varats/tools/driver_external.py
```diff
@@ -0,0 +1,126 @@
...
+
+import click
+
+import varats.utils.external_source_handling as esh
+from varats.ts_utils.cli_util import initialize_cli_tool
+from varats.utils.settings import save_config, vara_cfg
+
+LOG = logging.getLogger(__name__)
+
+# TODO: Fetch template from template repo (https://github.com/se-sic/vara-test-repos)
```
ReebaAslam: @boehmseb we discussed doing this, but fetching the template repo every time we have to set an external repo seems like an overkill. 
Doing this is necessary to make sure the latest file structure is validated against. 
How frequently do you think it will change?

An intermediate solution is to clone this repo when someone is installing or updating Vara-TS. We cache the directory structure against the version of Vara-TS. 
If you update structure, Vara-TS version also updates.

varats/varats/tools/driver_external.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Args:
+        repo_path: Path to the external repository root
+
+    Returns:
+        Tuple of (is_valid, missing_folders)
+        - is_valid: True if all template folders exist
+        - missing_folders: List of missing folders
+    """
+    # TODO: Extend validation to check folder contents (e.g., __init__.py files)
```
ReebaAslam: @boehmseb even in the template [https://github.com/se-sic/varats-oot-template/tree/test-oot/example_varats](url), we had files within the folders with also some classes and functions.

How granular do we want the validation to be?

varats/varats/tools/driver_external.py
```diff
@@ -0,0 +1,126 @@
...
+
+    if not repo_path.exists():
+        return False, ["Repository path does not exist"]
+
+    if not repo_path.is_dir():
+        return False, ["Repository path is not a directory"]
+
+    # Apply nested structure pattern: path/path_name/
+    source_root = esh.determine_project_source_root(repo_path)
```
ReebaAslam: @boehmseb this function just appends folder name to the path again, which means the local directory should look something like:

"~/example_tree/example_tree/experiments/file.py"

and the user will need to pass "~/example_tree"

I am not sure why Florian wanted this structure, but it seems unnecessary. 
We can send the internal directory, i.e "~/example_tree/example_tree"

The only potential benefit I can see is that the outer example_tree would be a git repo and will .git file which we can use to verify that the repo was indeed forked from our template, but do we need that?

varats/varats/tools/driver_external.py
```diff
@@ -0,0 +1,126 @@
...
+        for error in validation_errors:
+            LOG.error(f"    - {error}")
+
+        LOG.error(f"\n  Template example: {TEMPLATE_REPO}")
+        LOG.error("  Expected structure:")
+        source_root = esh.determine_project_source_root(repo_path)
+        for folder in TEMPLATE_FOLDERS:
+            LOG.error(f"    {source_root / folder}")
+
+        # TODO: should we unregister the repo?
```
ReebaAslam: @boehmseb

docs/source/vara-ts/vara-external.rst
```diff
@@ -0,0 +1,206 @@
...
+.. code-block:: console
+
+    Repository already registered at: /path/to/your/external-repository
+
+Configuration
+--------------
+
+Registered external repositories are stored in the VaRA configuration file. The configuration key is:
+
+.. code-block:: ini
```
boehmseb: If sphinx supports YAML, that would be the actual language used in the config file.

```suggestion
.. code-block:: yaml
```

docs/source/vara-ts/vara-external.rst
```diff
@@ -0,0 +1,206 @@
...
+
+    external_source_repositories = [
+        "/path/to/repo1",
+        "/path/to/repo2"
+    ]
+
+To manually modify registered repositories, edit your VaRA configuration file directly (usually at ``~/.vara/vara_config.yaml`` or equivalent on your system).
```
boehmseb: Our documentation uses `$VARATS_ROOT` as a system-agnostic shorthand for that directory.
Also, the config file name is incorrect here. 

```suggestion
To manually modify registered repositories, edit your VaRA configuration file directly (``$VARATS_ROOT/.varats.yaml``).
```

docs/source/vara-ts/vara-external.rst
```diff
@@ -0,0 +1,206 @@
...
+Template Repository
+-------------------
+
+For a complete example of a well-structured external repository with all required folders and structure, see the `VaRA Out-of-Tree Template <https://github.com/se-sic/varats-oot-template>`_.
+
+**Recommended approach:**
+
+1. **Clone** the template locally for a quick start
+2. **Fork** the template on GitHub for a portable, version-controlled repository under your own account
+3. Use `vara-external set` to register it with VaRA-TS
```
boehmseb: I think we can streamline this a bit by integrating these two alternatives directly into the **Step 1** part above.

docs/source/vara-ts/vara-external.rst
```diff
@@ -0,0 +1,206 @@
...
+    external_source_repositories = [
+        "/path/to/repo1",
+        "/path/to/repo2"
+    ]
+
+To manually modify registered repositories, edit your VaRA configuration file directly (usually at ``~/.vara/vara_config.yaml`` or equivalent on your system).
+
+Example Workflow
+----------------
+
```
boehmseb, line 114: Maybe, as step 0 or prerequisite, refer to the "Getting Started" guide.

docs/source/vara-ts/vara-external.rst
```diff
@@ -0,0 +1,206 @@
...
+For a complete example of a well-structured external repository with all required folders and structure, see the `VaRA Out-of-Tree Template <https://github.com/se-sic/varats-oot-template>`_.
+
+**Recommended approach:**
+
+1. **Clone** the template locally for a quick start
+2. **Fork** the template on GitHub for a portable, version-controlled repository under your own account
+3. Use `vara-external set` to register it with VaRA-TS
+
+Troubleshooting
```
boehmseb: Not sure how necessary this is. All these commands are already mentioned above and the documentation for this tool is not that long or complicated.

varats/varats/tools/research_tools/discover_research_tools.py
```diff
@@ -0,0 +1,32 @@
...
+
+    _builtin_tools = {
+        "phasar": Phasar,
+        "szzunleashed": SZZUnleashed,
+        "vara": VaRA,
+    }
+
+    for name, tool_class in _builtin_tools.items():
+        if name not in ResearchTool.REGISTRY:
+            ResearchTool.REGISTRY[name] = tool_class
```
boehmseb: This shouldn't be necessary as loading the modules for these research tools registers them automatically via `__init_subclass__()`.
Or am I missing something here?

varats/varats/tools/driver_external.py
```diff
@@ -0,0 +1,152 @@
...
+    # Validate repository structure
+    is_valid, validation_errors = validate_external_repo(repo_path)
+
+    current_repos = vara_cfg()['external_source_repositories'].value
+
+    repo_is_registered = str(repo_path) in current_repos
+
+    if not is_valid:
+        msg_lines = [
+            "Repository Not Complying To Template",
```
boehmseb:
```suggestion
            "Repository not complying to template!",
```

tests/tools/test_driver_external.py
```diff
@@ -0,0 +1,273 @@
...
+        is_valid, errors = driver_external.validate_external_repo(
+            non_existent_path
+        )
+
+        self.assertFalse(is_valid)
+        self.assertIn("Repository path does not exist", errors)
+
+    def test_repo_not_being_directory(self) -> None:
+        """Path pointing to a file returns False with appropriate error message."""
+        tmp_path = Path(tempfile.mkdtemp())
```
boehmseb: Consider using `TemporaryDirectory` as a context manager for automatic cleanup.
Currently, the temp directories get left behind after the test.

This goes for all of the tests, of course.

tests/tools/test_driver_external.py
```diff
@@ -0,0 +1,273 @@
...
+
+        file_path = tmp_path / "not_a_directory.txt"
+        file_path.touch()
+
+        is_valid, errors = driver_external.validate_external_repo(file_path)
+
+        self.assertFalse(is_valid)
+        self.assertIn("Repository path is not a directory", errors)
+
+    @run_in_test_environment()
```
boehmseb: You need this only if the tests modifies persistent state, e.g., the configuration files.
As far as I can see, this is not the case with the `validate_external_repo` function.
So I would remove the annotation and save the overhead.

tests/tools/test_driver_external.py
```diff
@@ -0,0 +1,273 @@
...
+        is_valid, errors = driver_external.validate_external_repo(repo_path)
+
+        self.assertTrue(is_valid)
+        self.assertEqual([], errors)
+
+
+class TestRegisterExternalRepository(unittest.TestCase):
+    """Unit tests for register_external_repository function."""
+
+    @run_in_test_environment()
```
boehmseb: Here, you need to keep the annotation as you write to the config.

tests/tools/test_driver_external.py
```diff
@@ -0,0 +1,273 @@
...
+
+        # Break the structure so validation fails on the next invocation.
+        shutil.rmtree(repo_path / driver_external.TEMPLATE_FOLDERS[0])
+
+        with patch(
+            "varats.tools.driver_external.click.confirm", return_value=True
+        ):
+            result = runner.invoke(
+                driver_external.main, ["set", str(repo_path)]
+            )
```
boehmseb: Use click's built-in input capturing functionality instead of mocking.
The following should work:

```suggestion
        result = runner.invoke(
            driver_external.main, ["set", str(repo_path)],
            input="y\n"
        )
```

tests/utils/test_external_repo_discovery_components.py
```diff
@@ -0,0 +1,321 @@
...
+                    "external_report" in module_name
+                    for module_name in sys.modules.keys()
+                    if "reports" in module_name
+                )
+            )
+        finally:
+            _cleanup_external_tool_modules(repo_path)
+
+
+def _create_external_tool_modules(root: Path) -> None:
```
boehmseb: I think this would be easier to maintain if you create a folder in the `TEST_INPUTS` directory and store these as real files there.
You can then use our `UnitTestFixtures` class in conjunction with `@run_in_test_environment()` to automatically create/delete it.
Take a look at the `tests.helper_utils` module for more information.

boehmseb (changes requested):
Overall, this looks good, but some cleanup needs to be done.
Especially the tests need some more love to run more cleanly and become more maintainable.

PR #951, opened 2026-08-26
Adds a source for git-stored files.

boehmseb:


varats/varats/ts_utils/project_sources.py
```diff
@@ -123,14 +128,93 @@ def version(self, target_dir: str, version: str = 'HEAD') -> pb.LocalPath:
...
+        """The revision from which the files are fetched."""
+        return self.__revision
+
+    def version(self, target_dir: str, version: str = "") -> pb.LocalPath:
+        """
+        Fetches the defined files for a given version to the target directory.
+
+        :param target_dir:
+        :param version:
+        :return:
```
LuAbelt: Seems to be an artefact from my initial implementation, but we should probably add some info here and change it to the Google Docstring style we usually use.

LuAbelt (changes requested):

LuAbelt (approved):

PR #953, opened 2026-09-17
bb_cfg update to context manager

blutub:


tests/utils/test_bb_config.py
```diff
@@ -95,3 +96,23 @@ def test_if_environment_updates_correctly(self):
...
+
+    def test_bb_cfg_function_calls(self):
+        """Test if the bb_cfg function returns the correct config."""
+        cfg = bb_cfg()
+        self.assertEqual(cfg, settings._BB_CFG)
+
+    def test_change_value_kwargs(self):
+        """Test if the change_value function updates the config correctly."""
+        with bb_cfg(tmp_dir="/new/tmp/dir") as cfg:
+            self.assertEqual(cfg["tmp_dir"], "/new/tmp/dir")
```
LuAbelt: Maybe also add another case that does not check the `cfg` object that is returned, but check whether the entry in a call to `bb_cfg()` is as expected, as this is the main use-case for this (i.e. using the context manager to change how `bb_cfg()` behaves when called in some function)

LuAbelt:
@blutub Maybe also already having tests that cover the case of nested configuration values

